Issue #246, opened 2021-06-23
Roadmap for Mava's next release


arnupretorius:
**This issue is meant to act as an interactive roadmap for Mava's stable version release**

We are happy to announce the Beta release of Mava! 🥳  Although we believe the Beta version offers much in the way of building MARL systems, there is still much work to be done. 

Below is a list of what we want to implement before releasing a stable, benchmarked and tested, first official version of Mava. Please note that this checklist refers primarily to core features and components underlying the working philosophy behind the framework. For more in the way of extensions and/or feature requests such as additional system implementations, please refer to our **wish list** -> #247. 
  
### Checklist for first release
* [ ] Create Nightly Release 
* [x] Make quick start notebook for mava
* [ ] Reach >90% code coverage for testing
* [ ] Create general abstract system and builder classes.
* [ ] Benchmark all systems on popular environments, including PettingZoo, SMAC (including system checkpoints and downloadable plotting data).
* [ ] Benchmark system scaling.
* [ ] Add system integration testing.
* [x] Better structuring/fixes for examples, perhaps create helper functions
* [ ] Single process versions (confirm if system.py needed).
* [x] Centralised variable source. Counters and all fixed-length variables can be stored here for distributed access.
* [x] Provide the option for using multiple trainers. This enables faster training for non-weight sharing agents and allows for hyperparameter tuning.
* [x] Enhancing shared weights. Allow for custom specification of which agents share weights and which do not. Each agent has a net_key instead of agent_type. shared_weights now just means all the agent net_keys are the same.
* [ ] Add executor trainer waiting logging info
* [x] Update .keys() and .values() code to fix bugs
* [ ] Batch the executor policies that use the same networks for increased performance.
* [ ] Implement hybrid action space for MA-DDPG, MA-D4PG and MAPPO.
* [x] Recurrency for QMIX and VDN - Inherit from MADRQN. enhancement.
* [ ] Implement a partially observable debugging environment (flicker spread).
* [ ] Handle legal actions in a generic way
* [x] _num_steps in MADDPG should be incremented outside the for loop in _update_target_network().
* [ ] Fix typing for Robocup environment.
* [ ] Investigate MADDPG/MAD4PG training at end of episode
* [x] Fix mypy import issue
* [x] Add/cleanup doc strings and comments
* [x] Add option to update executors every n environment steps instead of every step.
* [x] Support latest Acme and Reverb versions.
[...]

PR #253, opened 2021-06-29
Feature/Multiple Trainers for MA-DDPG

DriesSmit:
## What?
Implements a scaled-up version of MADDPG where multiple trainers can now be used with multiple executors. A centralised variable server is also implemented that absorbs the responsibilities of the counter node, trainer checkpointing and trainer variable source. The trainers and executors now read and write to the centralised variable source directly. A multiple trainer example is included where 3 trainers and 2 executors are used to train 3 non-weight sharing agents on the debugging environment.
## Why?
Multiple trainers allow for the parallelisation of the trainer's tasks. Just like is already done with executors. This also opens the door to hyperparameter tuning directly using Mava in future updates.
## How?
Added a new Scaled MA-DDPG system that allows for the use of multiple trainers.


mava/systems/tf/maddpg/system.py
```diff
@@ -270,6 +282,33 @@ def replay(self) -> Any:
 
         return self._builder.make_replay_tables(self._environment_spec)
 
+    def create_system(
```
arnupretorius, line 285: This is great 😄  We should also do this for the other systems.

mava/systems/tf/maddpg_scaled/execution.py
```diff
@@ -0,0 +1,184 @@
+# python3
```
arnupretorius, line 1: Once the code is approved, it will become the default MADDPG right? Or would there be `maddpg` and `maddpg_scaled`?

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scaled.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # distributed program
+    program = maddpg_scaled.MADDPG(
+        environment_factory=environment_factory,
+        network_factory=network_factory,
+        logger_config=logger_config,
+        num_executors=2,
+        num_trainers=3,
+        shared_weights=False,
+        trainer_net_config={
```
arnupretorius, line 103: Do we have default trainer net configs as well? For example, `{"trainer_0": ["agent_0", "agent_1", "agent_2"]}`? Also, the argument doesn't seem to be in the system code?

mava/systems/tf/variable_sources.py
```diff
@@ -0,0 +1,123 @@
...
+import time
+from typing import Any, Dict, Sequence, Union
+
+import numpy as np
+from acme.tf import utils as tf2_utils
+
+from mava.systems.tf import savers as tf2_savers
+
+
+class VariableSource:
```
arnupretorius, line 11: Is this also adapted from acme?

mava/systems/tf/variable_utils.py
```diff
@@ -0,0 +1,154 @@
...
+from typing import Any, Dict, List, Optional
+
+import numpy as np
+import tensorflow as tf
+from acme.tf import utils as tf2_utils
+
+from mava.systems.tf.variable_sources import VariableSource as MavaVariableSource
+
+
+class VariableClient:
```
arnupretorius, line 28: What changed here compared to the acme client?

arnupretorius (commented):
Massive effort, thanks @DriesSmit!! 🥳 

Just see my few comments.

mava/systems/tf/maddpg/system.py
DriesSmit: Agreed. It looks nice. I added it to #246 :+1:

mava/systems/tf/maddpg_scaled/execution.py
DriesSmit: I changed maddpg_scaled to MADDPG :+1: There is now just one MADDPG algorithm that can scaled or run as normal with one trainer.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scaled.py
DriesSmit: I added a default setup for the trainer_config :+1: So if no trainer_net_config is specified it sets trainer_0 to train on all the networks. We don't inherit from the system as it is still used by other algorithms which do not have the trainer_net_config setup. Does this make sense or can we do something else about this?

mava/systems/tf/variable_sources.py
DriesSmit: No this is a new Mava variable_source that stores all the variables in a centralised place. This allows the executors, evaluators and trainers to set and get the latest variables.

mava/systems/tf/variable_utils.py
DriesSmit: I made this client more general. The Acme client could just pull variables every few steps. This variable client provides functions to set and get variables from the centralised variable source.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
```diff
@@ -0,0 +1,131 @@
...
+        network_factory=network_factory,
+        logger_config=logger_config,
+        num_executors=2,
+        shared_weights=False,
+        trainer_networks={
+            "trainer_0": ["agent_0"],
+            "trainer_1": ["agent_1"],
+            "trainer_2": ["agent_2"],
+        },
+        executor_samples=[["agent_0", "agent_1", "agent_2"]],
```
arnupretorius, line 107: Why is this called `executor_samples`?

mava/adders/reverb/base.py
```diff
@@ -114,6 +140,151 @@ def __init__(
...
+                        else:
+                            is_in_entry = False
+                            break
+
+                    if is_in_entry:
+                        created_item = True
+
+                        # Create new empty transition
+                        if type(trajectory) == Step:
+                            soe = trajectory.start_of_episode  # type: ignore
```
arnupretorius, line 200: Do we need all the `type: ignore` here? There are quite a few in the file. Perhaps better to ignore the entire file for now? But I guess would be better to fix the type issues.

mava/adders/reverb/sequence.py
```diff
@@ -51,6 +54,8 @@ def __init__(
         client: reverb.Client,
         sequence_length: int,
         period: int,
+        int_to_nets: List[str] = None,
```
arnupretorius, line 57: What does `int_to_nets` mean? Also, please update the docstrings wherever appropriate.

mava/adders/reverb/sequence.py
```diff
@@ -131,9 +138,12 @@ def _maybe_create_item(
         table_priorities = utils.calculate_priorities(self._priority_fns, trajectory)
 
         # Create a prioritized item for each table.
-        for table_name, priority in table_priorities.items():
-            self._writer.create_item(table_name, priority, trajectory)
-            self._writer.flush(self._max_in_flight_items)
+        # for table_name, priority in table_priorities.items():
```
arnupretorius, line 141: Remove commented code?

mava/components/tf/networks/convolution.py
```diff
@@ -0,0 +1,29 @@
+from typing import List
+
+import sonnet as snt
+import tensorflow as tf
+
+
+class Conv1DNetwork(snt.Module):
```
arnupretorius, line 7: Should we put this in Mava? Not sure how generally useful it is for various projects? Definitely open to the idea, just not sure.

mava/systems/tf/mad4pg/__init__.py
```diff
@@ -17,7 +17,7 @@
     MAD4PGFeedForwardExecutor,
     MAD4PGRecurrentExecutor,
 )
-from mava.systems.tf.mad4pg.networks import make_default_networks
+from mava.systems.tf.mad4pg.networks import make_1d_conv_networks, make_default_networks
```
arnupretorius, line 20: Do we want this in Mava?

mava/systems/tf/mad4pg/networks.py
```diff
@@ -175,3 +181,154 @@ def make_default_networks(
         "critics": critic_networks,
         "observations": observation_networks,
     }
+
+
+def make_1d_conv_networks(
```
arnupretorius, line 186: Again, not sure if we should have this in Mava?

mava/systems/tf/mad4pg/system.py
```diff
@@ -48,10 +48,10 @@ def __init__(
         ] = training.MAD4PGDecentralisedTrainer,
         executor_fn: Type[core.Executor] = MADDPGFeedForwardExecutor,
         num_executors: int = 1,
-        num_caches: int = 0,
         environment_spec: mava_specs.MAEnvironmentSpec = None,
+        trainer_networks: Dict[str, List] = {},
```
arnupretorius, line 52: Need docstrings for these.

mava/systems/tf/maddpg/execution.py
```diff
@@ -13,30 +13,52 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""MADDPG system executor implementation."""
```
arnupretorius, line 16: Please keep the string here.

mava/systems/tf/maddpg/execution.py
```diff
@@ -13,30 +13,52 @@
...
+from mava.utils.sort_utils import sort_str_num
 
 Array = specs.Array
 BoundedArray = specs.BoundedArray
 DiscreteArray = specs.DiscreteArray
-
 tfd = tfp.distributions
 
 
+def sample_new_agent_keys(
```
arnupretorius, line 43: Is this only applicable to MADDPG? If not, perhaps we can move it to a utils file or something similar?

mava/systems/tf/maddpg/execution.py
```diff
@@ -173,11 +226,18 @@ def observe(
...
+        self._adder.add(policy, next_timestep, next_extras)  # type: ignore
 
-        if self._adder:
-            _, policy = actions
-            # TODO (dries): Sort out this mypy issue.
-            self._adder.add(policy, next_timestep, next_extras)  # type: ignore
+    def update(self, wait: bool = False) -> None:
+        """Update the policy variables."""
+        if self._variable_client:
+            # TODO (dries): Maybe changes this to a async get?
```
arnupretorius, line 239: Sounds good to me :)

mava/systems/tf/maddpg/system.py
```diff
@@ -13,10 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""MADDPG system implementation."""
+"""MADDPG scaled system implementation."""
```
arnupretorius, line 16: Remove "scaled"? Since this will become the default.

mava/systems/tf/maddpg/system.py
```diff
@@ -82,7 +89,7 @@ def __init__(
         bootstrap_n: int = 10,
         sigma: float = 0.3,
         max_gradient_norm: float = None,
-        max_executor_steps: int = None,
+        # max_executor_steps: int = None,
```
arnupretorius, line 92: Remove?

mava/systems/tf/maddpg/system.py
```diff
@@ -25,26 +25,33 @@
...
 from mava.utils.loggers import MavaLogger, logger_utils
+from mava.utils.sort_utils import sort_str_num
 from mava.wrappers import DetailedPerAgentStatistics
 
 
 class MADDPG:
     """MADDPG system."""
 
-    def __init__(
+    def __init__(  # noqa
```
arnupretorius, line 54: Why do you need the `# noqa` here?

mava/systems/tf/maddpg/system.py
```diff
@@ -58,10 +65,10 @@ def __init__(
         ] = training.MADDPGDecentralisedTrainer,
         executor_fn: Type[core.Executor] = MADDPGFeedForwardExecutor,
         num_executors: int = 1,
-        num_caches: int = 0,
```
arnupretorius, line 61: Do we not support trainer caching anymore? Or how does the caching work in this version?

mava/systems/tf/maddpg/system.py
```diff
@@ -269,84 +347,36 @@ def _get_extra_specs(self) -> Any:
         networks = self._network_factory(  # type: ignore
             environment_spec=self._environment_spec,
             agent_net_keys=self._agent_net_keys,
+            net_spec_keys=self._net_spec_keys,
         )
         for agent in agents:
-            agent_type = agent.split("_")[0]
+            # agent_type = agent.split("_")[0]
```
arnupretorius, line 353: Remove?

mava/systems/tf/maddpg/system.py
```diff
@@ -362,70 +392,54 @@ def trainer(
...
-        _ = system.create_system()
-
-        # behaviour policy networks (obs net + policy head)
-        behaviour_policy_networks = system.create_behaviour_policy()
+        # Create the system
+        behaviour_policy_networks, networks = self.create_system()
 
         # Create the executor.
         executor = self._builder.make_executor(
+            # executor_id=executor_id,
```
arnupretorius, line 441: Remove?

mava/wrappers/system_trainer_statistics.py
```diff
@@ -149,6 +149,112 @@ def _compute_statistics(self, data: Dict[str, Dict[str, float]]) -> None:
                 self._network_loggers[network].write(network_running_statistics)
 
 
+class ScaledTrainerStatisticsBase(TrainerWrapperBase):
```
arnupretorius, line 152: I'm guessing these classes will also become the default once we convert all the systems? I.e. we will replace the `TrainerStatisticsBase` and related classes with these.

arnupretorius (commented):
Hey @DriesSmit, I did a quick lighting review round on this. 😄 Will do a more in-depth review soon.

Thanks for the PR, looks great, a massive effort! 🥳  Going to be huge when we get this in! 🔥 

Please see the few comments I made, most of them minor. Once we have the benchmarking results with these new changes, we should be very close to merging. 💪

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
DriesSmit: Executor samples are the network samples that are chosen at random to construct an agent pool at the start of an episode. In the vanilla case, there will be only one network configuration to sample from. But this also allows for more complex sampling. In the case above the executor will sample networks randomly one at a time until all the agent slots are filled. Then at the end of the episode, the experience gets sent to the trainers corresponding to the networks that were used. I can explain it in more detail in a presentation.

mava/adders/reverb/base.py
DriesSmit: Yes, I would ideally like to fix the type: ignores. The problem is that trajectory can either be a Step or Transition which means it has different properties. mypy does not seem to pick up that I do a check for Step and therefore still complains that if it is a Transition line 200 will fail. I am not sure how one can fix this? @KaleabTessera any ideas?

mava/systems/tf/maddpg/system.py
DriesSmit: The caching never worked or did it? I don't mind adding it back in, but just removed it because caching just crashed if you set it to 1.

mava/wrappers/system_trainer_statistics.py
DriesSmit: Yes, exactly. Hopefully, this becomes the default trainer wrapper.

mava/adders/reverb/sequence.py
DriesSmit: I updated the docstrings of the system.py and adders. int_to_nets is used to convert integers to network keys. The reason this is used is the networks_keys need to be sent to the adders. This is to allow the adder to know to which table to write parts of each data. The problem is that the adders don't like to work with strings, because at some point zero padding is used. So now I use ints which maps to the network keys.

mava/adders/reverb/sequence.py
DriesSmit: Removed.

mava/components/tf/networks/convolution.py
DriesSmit: I was also wondering about this. I am open to just use it locally in my project. So maybe I should move it? @KaleabTessera what do you think?

mava/systems/tf/maddpg/execution.py
DriesSmit: Added.

mava/systems/tf/mad4pg/system.py
DriesSmit: Added.

mava/systems/tf/maddpg/execution.py
DriesSmit: Good point. In the future, this will hopefully be applicable to all the executors. I moved it to the sort_utils.

mava/systems/tf/maddpg/system.py
DriesSmit: Fixed it.

mava/systems/tf/maddpg/system.py
DriesSmit: For some reason, mypy is complaining that the function is too complex. I am guessing it has too many arguments.

mava/systems/tf/maddpg/system.py
DriesSmit: Fixed.

mava/systems/tf/maddpg/system.py
DriesSmit: Fixed.

mava/systems/tf/maddpg/system.py
DriesSmit: Removed.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
KaleabTessera: @DriesSmit As discussed, I think we need to simplify the names of these variables so that it is easier to tell what they are doing.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
```diff
@@ -0,0 +1,131 @@
...
+
+    # networks
+    network_factory = lp_utils.partial_kwargs(make_default_networks)
+
+    # loggers
+    # set custom logger config for each process
+    # -- log trainer,executor and evaluator to TF
+    # -- log only evaluator to terminal
+    log_every = 10
+    logger_config = {
```
KaleabTessera, line 68: Can we make like the other examples, where we share a single config, just to make the examples easier to follow :)

mava/adders/reverb/base.py
```diff
@@ -74,6 +89,17 @@ def spec_like_to_tensor_spec(
     return tf.TensorSpec.from_spec(spec, name="/".join(str(p) for p in paths))
 
 
+def get_trans_net_agents(
+    trajectory: Union[Trajectory, mava_types.Transition], entry_net_keys: Dict[str, str]
+) -> Tuple[List, Dict[str, List]]:
+    agents = sort_str_num(trajectory.actions.keys())
+    unique_nets = sort_str_num(set(entry_net_keys.values()))
+    trans_nets_agent: Dict[str, List] = {key: [] for key in unique_nets}
```
KaleabTessera, line 97: `trans_nets_agent` ? I think docstrings for this function would help a bit.

mava/components/tf/networks/convolution.py
KaleabTessera: I think we should have it in Mava, seems like a useful component that could be useful for a few systems. @DriesSmit @arnupretorius

KaleabTessera (commented):
Thanks @DriesSmit , this will be a really great improvement to Mava :fire: :fire:  It is quite a lot of files so I will review it in stages. In general adding docstrings and more descriptive var names might make reviewing this a bit easier.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
DriesSmit: What names do you think I should use? Maybe `network_sample_sets`?

mava/adders/reverb/sequence.py
DriesSmit: Should I maybe call `int_to_nets` > `int_ids_to_net_keys_map` ?

mava/systems/tf/mad4pg/__init__.py
DriesSmit: See above. I am happy to remove it or leave it.

mava/systems/tf/mad4pg/networks.py
DriesSmit: See above. I am happy to remove it or leave it.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
DriesSmit: Good catch :+1: I fixed it.

mava/adders/reverb/base.py
DriesSmit: Thanks. Yay, I really need to update the variable names a bit. I updated this variable name and added a docstring :+1:

mava/adders/reverb/sequence.py
KaleabTessera: Can we do this right before adding. Like having everything like before, just before adding convert to int?

KaleabTessera (commented):
Will do another pass of the rest of this PR.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # distributed program
+    program = maddpg.MADDPG(
+        environment_factory=environment_factory,
+        network_factory=network_factory,
+        logger_factory=logger_factory,
+        num_executors=2,
+        shared_weights=False,
+        trainer_networks={
+            "trainer_0": ["agent_0"],
```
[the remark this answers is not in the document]
DriesSmit: Updated.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
```diff
@@ -0,0 +1,114 @@
...
+        },
+        executor_samples=[["agent_0", "agent_1", "agent_2"]],
+        policy_optimizer=snt.optimizers.Adam(learning_rate=1e-4),
+        critic_optimizer=snt.optimizers.Adam(learning_rate=1e-4),
+        checkpoint_subpath=checkpoint_dir,
+        max_gradient_norm=40.0,
+    ).build()
+
+    # launch
+    gpu_id = -1
```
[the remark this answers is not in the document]
DriesSmit: Updated.

examples/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+    # distributed program
+    program = maddpg.MADDPG(
+        environment_factory=environment_factory,
+        network_factory=network_factory,
+        logger_factory=logger_factory,
+        num_executors=2,
+        shared_weights=False,
+        trainer_networks={
```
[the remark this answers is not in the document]
DriesSmit: Good point:) I added enums for easier trainer_networks configuration.

mava/adders/reverb/base.py
```diff
@@ -74,6 +89,29 @@ def spec_like_to_tensor_spec(
...
+    Args:
+        trajectory (Union[Trajectory, mava_types.Transition]): Trajectory recorded by
+        the adders.
+        entry_net_keys (Dict[str, str]): The network_keys used by each entry.
+    Returns:
+        agents (List): A sorted list of al the agent_keys.
+        trajectory_net_agents (Dict[str, List]): A dictionary that maps network_keys to
+        a list of agents using each network key.
+    """
+    agents = sort_str_num(trajectory.actions.keys())
```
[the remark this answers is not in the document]
DriesSmit: Thanks. I did check. The sort is necessary here because of the actions.keys. This could be in any order but the agents returned by this function need to be in the correct order.

tests/networks/networks_test.py
```diff
@@ -70,10 +70,19 @@ def test_shared_networks(self, system: Any) -> None:
...
-        networks = system.make_default_networks(  # type: ignore
-            environment_spec=environment_spec,  # type: ignore
-            agent_net_keys=agent_net_keys,
-        )
+
+        if system == mad4pg:
+            networks = system.make_default_networks(  # type: ignore
+                environment_spec=environment_spec,  # type: ignore
+                agent_net_keys=agent_net_keys,
+                vmin=-150,
```
arnupretorius, line 78: Should we maybe add `**kwargs` to the `make_default_networks` so that we don't have to have if statements like this?

arnupretorius (commented):
Epic effort @DriesSmit! 🔥  Thanks for all the new updates. Also, for all the detailed comments and docstrings!

Please just see my minor comments. 

@KaleabTessera I think we should start doing some test runs on the experiment runner. 😄 

Almost there! 🥳

mava/adders/reverb/base.py
```diff
@@ -146,60 +158,101 @@ def write_experience_to_tables(  # noqa
         trajectory: Union[Trajectory, mava_types.Transition],
         table_priorities: Dict[str, Any],
     ) -> None:
-        # Get a dictionary of the transition nets and agents.
+        """Write an episode eperience (trajectory) to the reverb tables. Each
+        table represents experience used by each of the trainers. Therefore
+        this function dynamically determines to which table(s) to write
+        parts of the trajectory based on what networks where used by
```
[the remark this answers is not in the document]
DriesSmit: Yes. The tables can point to different parts of the underlying trajectory. So e.g. one table can take experience from agent_1 with network_1 and another table can take experience from agent_2 with network_2 from the same trajectory. A table can even batch more than one entry from the same trajectory if e.g. network_1 is used by more than one trainer.

mava/adders/reverb/base.py
```diff
@@ -146,60 +158,101 @@ def write_experience_to_tables(  # noqa
...
+            created_item = False
+
+            # In this loop we go through every table one by one.
+            # We check for each table if the experience contains
+            # the correct networks for that trainer. If it does
+            # we add to the table a subset of the trajectory's agents
+            # which only contains the networks combination that the
+            # trainer is interested in. Note that a table might
+            # find multiple copies of the correct network combination
+            # and therefore might write more than once to a table
```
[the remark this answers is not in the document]
DriesSmit: To the best of my knowledge (and research) the underlying data is only stored once. The tables only link to the data so there should not be any data duplication.

mava/adders/reverb/base.py
```diff
@@ -146,60 +158,101 @@ def write_experience_to_tables(  # noqa
...
+                # a fresh copy of all the agents and removes agents as it
+                # pushes the experience to its table.
+                trajectory_dict_copy = copy.deepcopy(trajectory_nets_agent)
 
                 # While the networks are in the data keep creating tables
                 # Each training example can therefore create multiple items
                 is_in_entry = True
                 while is_in_entry:
+                    # Go through all the networks in the table specification
+                    # and see if all the networks are still in the remaining
```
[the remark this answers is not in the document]
DriesSmit: Apologises, I could have worded that better. I updated the comments. Basically the table searches through the remaining agents to see if the set networks it is looking for are still in the trajectory. If it is they agents, with the correct combination of networks get extracted from the trajectory and sent to the table as an entry.

mava/adders/reverb/base.py
```diff
@@ -208,58 +261,65 @@ def write_experience_to_tables(  # noqa
                                 extras={},
                             )
                         else:
-                            new_trans = mava_types.Transition(  # type: ignore
+                            # Create a new transition trajectory
+                            new_trajectory = mava_types.Transition(  # type: ignore
```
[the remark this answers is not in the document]
DriesSmit: The trajectory can be either a `Step` or a `Transition` depending whether it is sequence based or transition based. So it might make sense to keep it as is? Not sure. What do you think?

mava/adders/reverb/transition.py
```diff
@@ -102,6 +102,10 @@ def __init__(
             (s_t, a_t, r_t, d_t, s_t+1, e_t).
           discount: Discount factor to apply. This corresponds to the
             agent's discount in the class docstring.
+          int_to_nets: A list of network names to convert from integers to
```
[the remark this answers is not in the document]
DriesSmit: Agreed. We should call it `net_ids_to_keys`. Great catch :)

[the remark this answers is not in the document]
DriesSmit: Good point. I simplified the explanation of `net_ids_to_keys`. It is a list of network keys. By indexing the list with the network id the corresponding network key is returned.

mava/systems/tf/maddpg/system.py
```diff
@@ -198,52 +186,86 @@ def __init__(  # noqa
...
-            self._agent_net_keys = {
-                agent: agent.split("_")[0] if shared_weights else agent
-                for agent in agents
-            }
-            self._executor_samples = [
-                [
-                    self._agent_net_keys[key]
-                    for key in sort_str_num(self._agent_net_keys.keys())
+        self._network_sampling_setup = network_sampling_setup
+
```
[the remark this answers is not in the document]
DriesSmit: I think this space is fine. Or am I missing something? The red lines are removed and therefore that last forloop is not in the code anymore.

mava/systems/tf/maddpg/system.py
```diff
@@ -198,52 +186,86 @@ def __init__(  # noqa
...
+                    agent: agent.split("_")[0] if shared_weights else agent
+                    for agent in agents
+                }
+                self._network_sampling_setup = [
+                    [
+                        self._agent_net_keys[key]
+                        for key in sort_str_num(self._agent_net_keys.keys())
+                    ]
+                ]
+            elif network_sampling_setup == enums.NetworkSampler.random_policy_per_agent:
```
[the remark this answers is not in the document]
DriesSmit: Good idea:) I updated the code.

tests/networks/networks_test.py
DriesSmit: Yes so **kwargs is what we had. But I removed the kwargs for vmin and vmax. The reason for this being that vmin and vmax is environment specific and assigning any default value would not really make sense. I am just afraid that the user might just accept the default value and this would significantly hamper their training if the default is values are not correct. So I changed it to force the user to specify it. Let me know what you think.

mava/adders/reverb/base.py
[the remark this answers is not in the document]
arnupretorius: Ok sure, no that is fine for now.

mava/systems/tf/maddpg/system.py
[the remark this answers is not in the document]
arnupretorius: It might just be github issue on my side, but below the space is a hanging list in a list `[["network_0", "network_1", "network_2"]]`?

mava/adders/reverb/base.py
```diff
@@ -146,60 +158,101 @@ def write_experience_to_tables(  # noqa
...
+                    # Transition adder case.
+                    traj_net_keys[agent] = self._int_to_nets[arr]
 
-            # Get the unique agents and mapping from net_keys to agents.
-            agents, trans_nets_agent = get_trans_net_agents(
-                trajectory=trajectory, entry_net_keys=entry_net_keys
+            # Get a list of the agents and mapping from net_keys to all
+            # agents using that network.
+            agents, trajectory_nets_agent = get_trajectory_net_agents(
+                trajectory=trajectory, traj_net_keys=traj_net_keys
```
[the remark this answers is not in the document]
DriesSmit: Agreed :) I updated the code.

mava/systems/tf/maddpg/system.py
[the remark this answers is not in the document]
DriesSmit: O I see. Weird. I did remove it. If you look [here](https://github.com/instadeepai/Mava/blob/feature/mava-scaling/mava/systems/tf/maddpg/system.py#L189) it does not seem to be in the code anymore.

tests/networks/networks_test.py
arnupretorius: Would it not work to have `**kwargs` and then simply not have defaults for vmin and vmax in the mad4pg default networks function?

mava/systems/tf/maddpg/system.py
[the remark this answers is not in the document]
arnupretorius: Ah I see, ok thanks :)

tests/networks/networks_test.py
DriesSmit: Thanks. This was resolved in a call.

KaleabTessera: Just checking if this will change or stay like this?

mava/adders/reverb/base.py
```diff
@@ -115,6 +154,214 @@ def __init__(
...
+            if not created_item:
+                raise EOFError(
+                    "This experience was not used by any trainer: ",
+                    trajectory.actions.keys(),
+                )
+        else:
+            # Default setting (deprecate this) with only one table. In this setting
+            # we write the entire trajectory to that table.
+            table_items = list(table_priorities.items())
+            assert len(table_items) == 1
```
KaleabTessera, line 356: Why are we assuming there is one table here? Shouldn't we just ` for table_name, priority in table_priorities.items():` ...  & write, e.g. for systems that haven't implemented multiple trainers yet?

mava/adders/reverb/base.py
```diff
@@ -115,6 +154,214 @@ def __init__(
         )
         self._use_next_extras = use_next_extras
 
+    def write_experience_to_tables(  # noqa
+        self,
+        trajectory: Union[Trajectory, mava_types.Transition],
+        table_priorities: Dict[str, Any],
+    ) -> None:
+        """Write an episode eperience (trajectory) to the reverb tables. Each
```
KaleabTessera, line 162: Very minor, `eperience` -> `experience` :smile:

mava/adders/reverb/base.py
```diff
@@ -115,6 +154,214 @@ def __init__(
...
+            # we write the entire trajectory to that table.
+            table_items = list(table_priorities.items())
+            assert len(table_items) == 1
+            table, priority = table_items[0]
+            self._writer.create_item(
+                table=table, priority=priority, trajectory=trajectory
+            )
+
+        # Flush the writer.
+        self._writer.flush(self._max_in_flight_items)
```
KaleabTessera, line 363: So for certain adders, we previously flushed after every write , e.g. sequence adder  (https://github.com/instadeepai/Mava/blob/cbdbf6a90a3750428f29cc5e893ad67b54d6cee8/mava/adders/reverb/sequence.py#L136). Not we flush the writer once at the end (like we did for the transition adder). Not sure if this has any side-effects?

mava/systems/tf/maddpg/system.py
```diff
@@ -94,89 +105,71 @@ def __init__(
...
-                Defaults to 5.
-            sequence_length (int, optional): recurrent sequence rollout length. Defaults
-                to 20.
-            period (int, optional): Consecutive starting points for overlapping
-                rollouts across a sequence. Defaults to 20.
-            bootstrap_n (int, optional): Used to determine the spacing between
+            trainer_networks: networks each
+                trainer trains on.
+            network_sampling_setup: List of networks that are randomly
+                sampled from by the executors at the start of an environment run.
```
KaleabTessera, line 127: For clarity, can we please add another sentence on this. Something like defaults to `fixed_agent_networks`, which means ....  . Essentially so that people who don't need to work about this var, can ignore it, while others can tweak it. I know you do mention this below, but I think it would be useful here.

mava/systems/tf/maddpg/system.py
```diff
@@ -531,63 +536,86 @@ def evaluator(
...
 
         eval_loop = DetailedPerAgentStatistics(eval_loop)
         return eval_loop
 
+    def trainer(
+        self,
+        trainer_id: str,
+        replay: reverb.Client,
+        variable_source: MavaVariableSource,
+        # counter: counting.Counter,
```
KaleabTessera, line 551: Please remove comment.

mava/systems/tf/maddpg/system.py
```diff
@@ -531,63 +536,86 @@ def evaluator(
...
+
+        # Create the system
+        _, networks = self.create_system()
+
+        dataset = self._builder.make_dataset_iterator(
+            replay, f"{self._builder._config.replay_table_name}_{trainer_id}"
+        )
+
+        return self._builder.make_trainer(
+            # trainer_id=trainer_id,
```
KaleabTessera, line 579: Please remove comment.

mava/systems/tf/variable_utils.py
```diff
@@ -0,0 +1,175 @@
...
+    def add_and_wait(self, names: List[str], vars: Dict[str, Any]) -> None:
+        """Adds the specified variables to the corresponding variables in source
+        and waits for the process to complete before continuing."""
+        self._client.add_to_variables(names, vars)
+
+    def get_and_wait(self) -> None:
+        """Updates the get variables with the latest copy from source
+        and waits for the process to complete before continuing."""
+        self._copy(self._request())  # type: ignore
+        return
```
KaleabTessera, line 141: I asked this in another comment, does `return` have any effect or is this the same as acme's `update_and_wait` (https://github.com/deepmind/acme/blob/e4b6046f7ab70b2cf1769847605f06059800f599/acme/tf/variable_utils.py#L97)?

KaleabTessera (commented):
Thanks so much for this massive effort @DriesSmit  :raised_hands:  :trophy: This is close to ready from my standpoint, I just had some comments and then we can do the benchmarks :rocket:  :fire:

mava/systems/tf/mad4pg/system.py
```diff
@@ -48,10 +49,14 @@ def __init__(
         ] = training.MAD4PGDecentralisedTrainer,
         executor_fn: Type[core.Executor] = MADDPGFeedForwardExecutor,
         num_executors: int = 1,
-        num_caches: int = 0,
```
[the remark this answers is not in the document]
DriesSmit: `num_caches` does not work. Do you think it should be kept in the code? I think we should either try and fix it or remove it.

mava/systems/tf/mad4pg/system.py
```diff
@@ -72,7 +77,6 @@ def __init__(
         bootstrap_n: int = 10,
         sigma: float = 0.3,
         max_gradient_norm: float = None,
-        max_executor_steps: int = None,
```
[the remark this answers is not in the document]
DriesSmit: Good point. I think we do need it. However I will code in the logic manually and therefore we have more freedom of choice. So let me know what you want. The system can exit after a certain amount of executor steps, trainer steps or time? Your call.

mava/utils/adder_utils.py
```diff
@@ -0,0 +1,18 @@
...
+from acme.adders.reverb import base
+from acme.tf import utils as tf2_utils
+
+
+def calculate_priorities(
+    priority_fns: base.PriorityFnMapping, steps: Union[base.Step, Sequence[base.Step]]
+) -> Dict[str, float]:
+    """Helper used to calculate the priority of a sequence of steps."""
+
+    if isinstance(steps, list):
```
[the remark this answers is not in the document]
DriesSmit: Sure. However do we even need our own `calculate_priorities` or can't we just use Acme's one and delete this function?

mava/utils/environments/render_utils.py
```diff
@@ -22,322 +22,337 @@
...
+try:  # noqa
     os.environ["PYGAME_HIDE_SUPPORT_PROMPT"] = "hide"
     import pygame
     from pysc2.lib import colors, features, point, transform
     from pysc2.lib.renderer_human import _Surface
     from smac.env import StarCraft2Env
-except ModuleNotFoundError:
-    pass
 
+    class Renderer:
```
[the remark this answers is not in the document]
DriesSmit: I am not actually sure what these changes are. Maybe it is undoing one of your changes?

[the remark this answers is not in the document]
DriesSmit: O I think I reordered some of the logic because I wanted to make it not crash Mava if you do not use render and don't have the correct imports. But I am not sure if it is undoing some of your commits. Should I delete it completely for now?

mava/utils/sort_utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+def sort_str_num(str_num: Any) -> List[Any]:
+    return sorted(str_num, key=natural_keys)
+
+
+def sample_new_agent_keys(
+    agents: List,
+    network_sampling_setup: List,
+    net_keys_to_ids: Dict[str, int] = None,
+) -> Tuple[Dict[str, np.array], Dict[str, np.array]]:
```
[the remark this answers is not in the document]
DriesSmit: Done.

mava/systems/tf/mad4pg/training.py
```diff
@@ -132,13 +128,12 @@ def __init__(
...
+            variable_client=variable_client,
+            counts=counts,
+            num_steps=num_steps,
             agent_net_keys=agent_net_keys,
-            checkpoint_minute_interval=checkpoint_minute_interval,
             max_gradient_norm=max_gradient_norm,
-            counter=counter,
             logger=logger,
-            checkpoint=checkpoint,
-            checkpoint_subpath=checkpoint_subpath,
```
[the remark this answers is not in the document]
DriesSmit: Yes, we no longer need this. Checkpointing was previously done in the trainer but is now done in the variable_source.

mava/environment_loop.py
```diff
@@ -423,7 +423,26 @@ def run_episode(self) -> loggers.LoggingData:
             return self._get_running_stats()
         else:
             # Record counts.
-            counts = self._counter.increment(episodes=1, steps=episode_steps)
+            if hasattr(self._executor, "_counts"):
+                loop_type = "executor"
+                if "_" not in self._loop_label:
```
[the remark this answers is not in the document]
DriesSmit: Yes, good idea. That would simplify the code quite a bit.

mava/systems/tf/maddpg/builder.py
```diff
@@ -267,83 +322,158 @@ def make_adder(
...
-                update_period=self._config.executor_variable_update_period,
+                variables=variables,
+                get_keys=get_keys,
+                get_period=self._config.executor_variable_update_period,
             )
 
             # Make sure not to use a random policy after checkpoint restoration by
             # assigning variables before running the environment loop.
-            variable_client.update_and_wait()
+            variable_client.get_and_wait()
```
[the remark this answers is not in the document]
DriesSmit: There is no practical difference between Mava's `get_and_wait` and Acme's  `update_and_wait.` However in Mava we have a `set_and_wait` as well which we do not have in Acme. So to make it easier to differentiate the two calls I made one `get_and_wait` that pulls all the latest variables (inside the `get_keys`) from the variable source. And `set_and_wait` that pushes all the variables in the `set_keys` to the variable source.

mava/wrappers/smac.py
```diff
@@ -27,269 +27,283 @@
...
-    pass
-
-
 from mava import types
+from mava.utils.environments.render_utils import Renderer
+from mava.utils.sort_utils import sort_str_num
 from mava.utils.wrapper_utils import convert_np_type, parameterized_restart
 from mava.wrappers.env_wrappers import ParallelEnvWrapper
 
+try:  # noqa
```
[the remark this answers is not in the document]
DriesSmit: So the problem for me was that if I don't have SMAC installed this import code complained. I had to uninstall SMAC to get PyGame working for my environment.

mava/adders/reverb/base.py
DriesSmit: I am replying to the above comment here. For the moment it has to stay like that yes. The vmin and vmax cannot be kwargs. But if you can think about a cleaner way to do this let me know? Maybe having a dictionary with kwargs in instead of a if statement?

For the second comment. Maybe I am mistaken, but I don't think any system uses more than one table except for multiple trainers. But I am happy to change it back if you think it allows for more flexibility. Let my know.

mava/adders/reverb/base.py
DriesSmit: Fixed :)

mava/adders/reverb/base.py
DriesSmit: Great observation! I have not considered this. Flush seems to wait for all the data to be written. Please see [here](https://github.com/deepmind/reverb/blob/master/reverb/trajectory_writer.py#L356). So maybe its fine to do this only at the end of a trajectory step. Not sure. I am happy either way. I just don't want to make the writing even slower if we flush to often. But also if we need to flush after each write then we should change it.

mava/systems/tf/maddpg/system.py
DriesSmit: Agreed. I updated the docstring.

mava/systems/tf/maddpg/system.py
DriesSmit: Removed.

mava/systems/tf/maddpg/system.py
DriesSmit: Removed.

mava/systems/tf/variable_utils.py
DriesSmit: Please see the above comment. The return is just to make mypy happy. It has no effect.

tests/networks/networks_test.py
arnupretorius: We decided to keep it as is.

mava/systems/tf/mad4pg/system.py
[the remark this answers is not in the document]
KaleabTessera: Thanks @DriesSmit ! When you say manually, do you mean without using `lp_utils.StepsLimiter` ?

 I think we can make this flexible, maybe `termination_condition={"executor_steps": 10000}`, that way we can do `termination_condition={"trainer_steps": 10000}` as well.  `StepsLimiter` accepts a `steps_key` key, so that can be the dict key.

mava/systems/tf/mad4pg/system.py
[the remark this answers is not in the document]
KaleabTessera: Let's take it out then. We can always re-add it :)

mava/utils/adder_utils.py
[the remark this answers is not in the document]
KaleabTessera: So we use acme's calc props in various places (https://github.com/instadeepai/Mava/blob/cbdbf6a90a3750428f29cc5e893ad67b54d6cee8/mava/adders/reverb/transition.py#L194). I think if you don't anything specific, let's just use acme's one.

mava/utils/environments/render_utils.py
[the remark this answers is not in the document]
KaleabTessera: No worries.  I will merge and fix this in my PR.

mava/environment_loop.py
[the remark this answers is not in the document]
KaleabTessera: If it is too much effort, let's just add a todo :)

mava/adders/reverb/base.py
KaleabTessera: Sorry, I phrased my previous message incorrectly. The flush is after every `create_item` , not write. 

Let's do a flush after every `create_item`. Acme did that recently across all writers (https://github.com/deepmind/acme/commit/94711b1a59877ebe2d8ce8e79187f7af24ff3472). Please check we do it for the case where it is not multiple trainers as well.

mava/wrappers/smac.py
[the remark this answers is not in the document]
KaleabTessera: Let's chat about this tomorrow. The try catch on `from smac.env import StarCraft2Env` should mean you don't need smac when using mava, without these changes, but maybe I missed something.

mava/utils/adder_utils.py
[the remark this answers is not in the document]
DriesSmit: Sounds good. I deleted the Mava function.

mava/adders/reverb/base.py
DriesSmit: Done.

mava/systems/tf/mad4pg/system.py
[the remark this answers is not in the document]
DriesSmit: Done:) You can now use any of the following arguments as termination conditions: trainer_steps, trainer_walltime, evaluator_steps, evaluator_episodes, executor_episodes or executor_steps.

mava/adders/reverb/base.py
DriesSmit: I changed this back to how it was previously. I think doing it after every `create_item` broke my training.

mava/environment_loop.py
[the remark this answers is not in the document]
DriesSmit: Maybe we can add it to a todo :) Could you provide the wording for the todo?

mava/adders/reverb/base.py
DriesSmit: 1.) I like where the solution is going!  made a slight adjustment to the code, but the idea is there. Let me know if you are happy.
2.) We won't ignore tables due to the assert. So if there is more than one table the user will know about it. But sure I don't mind adding it back. I fixed it.

DriesSmit:
Trainer logging does not seem to be working for `mava/examples/petting_zoo/sisl/multiwalker/recurrent/decentralised/run_maddpg.py`. Fix this. Also compare training MAD4G mutliwalker executor and training speeds with develop.

PR #258, opened 2021-07-01
Feature/Improve Mava agent networks

DriesSmit:
## What?
Update the way one specifies agent networks in Mava. The shared_weights option is still available if needed. Now a custom network key can be specified per agent. This allows for a wider range of training settings while also simplifying the Mava code. If the agent_net_config is not used the system defaults back to the shared_weights method of doing things.
## Why?
To simplify how networks are assigned to agents in Mava, while also allowing for more customisation.
## How?
Updated all the existing systems in Mava with the new agent_net_config specification.
## Extra
Fixed the example: debugging/simple_spread/recurrent/state_based/run_maddpg.py.


mava/systems/tf/dial/system.py
```diff
@@ -132,6 +133,8 @@ def __init__(
                 Defaults to None.
             shared_weights (bool, optional): whether agents should share weights or not.
```
arnupretorius, line 134: We should perhaps just add here that when a `agent_net_config` is supplied the value of `shared_weights` is ignored, if I understand correctly?

arnupretorius (commented):
Thanks @DriesSmit! 🔥  This looks nice and clean. Definitely better than before! 

I think we just need some tests. Then, one other thing, do you think we should call it `agent_net_config`? Because that sounds like a config specifying the network architectures/hyperparameters for each agent? Perhaps we can call it `agent_net_assignment`? Or is that too long? Let me know what you think.

mava/systems/tf/dial/system.py
DriesSmit: Thanks for the suggestions @arnupretorius :+1: I updated the code to reflect your comments.  `agent_net_config` is now changed to `agent_net_keys`.

mava/components/tf/architectures/networked.py
```diff
@@ -38,20 +38,21 @@ def __init__(
...
             observation_networks=observation_networks,
             policy_networks=policy_networks,
-            shared_weights=shared_weights,
+            agent_net_keys=agent_net_keys,
         )
 
         self._network_spec = network_spec
 
-        if self._shared_weights:
+        if len(set(self._agent_net_keys.values())) < len(self._agent_net_keys.values()):
```
KaleabTessera, line 52: @DriesSmit Can you give me an example of when this would be true?

KaleabTessera (commented):
Thanks @DriesSmit , this looks really good! Much nicer way of handling agent networks :fire: :open_hands:  

I know you made this backward compatible, but just checking that you checked this doesn't cause issues with the examples? Also, could you please also update `make_custom_network` in `quickstart.ipynb`.

mava/components/tf/architectures/networked.py
DriesSmit: > Thanks @DriesSmit , this looks really good! Much nicer way of handling agent networks  
> 
> I know you made this backward compatible, but just checking that you checked this doesn't cause issues with the examples? Also, could you please also update `make_custom_network` and this line `network_factory = lp_utils.partial_kwargs(maddpg.make_default_networks,shared_weights=False)` in `quickstart.ipynb`.

I checked a few different examples throughout Mava and everything seems to still be training :+1:  I updated the notebook.

DriesSmit: > @DriesSmit Can you give me an example of when this would be true?

When agents share networks. Then the set of possible networks are less than the number of agents.

tests/networks/networks_test.py
```diff
@@ -0,0 +1,77 @@
...
+                "agent_1": "agent",
+                "agent_2": "agent",
+            }
+            networks = system.make_default_networks(  # type: ignore
+                environment_spec=environment_spec,  # type: ignore
+                agent_net_keys=agent_net_keys,
+            )
+
+            for key in networks.keys():
+                assert ["agent"] == list(networks[key].keys())
```
arnupretorius, line 61: I think we should perhaps make the two asserts (one here and below) two separate tests? Also, this test is nice for checking that the key strings much up, but should we also add a test to check the actual difference in the weights in the networks? Or do you think it is sufficient to just check the keys?

arnupretorius (commented):
Thanks for the tests @DriesSmit! 😄

Please just see my comments.

tests/networks/networks_test.py
```diff
@@ -0,0 +1,69 @@
...
+                "agent_0": "agent_0",
+                "agent_1": "agent_1",
+                "agent_2": "agent_2",
+            }
+            networks = system.make_default_networks(environment_spec=environment_spec, # type: ignore
+                        agent_net_keys=agent_net_keys)
+            for key in networks.keys():
+                assert ["agent_0", "agent_1", "agent_2"] == list(networks[key].keys())
+
+TestNetworks().test_network_keys()
```
KaleabTessera, line 77: I don't think we need this. Did it not run without it?

tests/networks/networks_test.py
KaleabTessera: I think two tests might also be better. Incase it fails, easier to know which is one failed and no dependency between the two. 

@DriesSmit  Are we still adding the tests that check the networks at a system level? I think that answers @arnupretorius 's questions.

tests/networks/networks_test.py
DriesSmit: I removed it.

tests/networks/networks_test.py
DriesSmit: @arnupretorius I changed it to be two tests. I don't think we can test if the variables are shared here? Variables only get initialised inside the builder. Let me know if that answers your question?

DriesSmit: > I think two tests might also be better. Incase it fails, easier to know which is one failed and no dependency between the two.
> 
> @DriesSmit Are we still adding the tests that check the networks at a system level? I think that answers @arnupretorius 's questions.

@KaleabTessera We can still add system-level tests yes. Let me know if you still think it is needed :+1:

arnupretorius: @DriesSmit , I think we should be able to test it in a similar way to how we do the smoke tests for the systems, i.e. write the test once the variables are already initialised. I think this is what @KaleabTessera is referring to.

arnupretorius: For example, here: https://github.com/instadeepai/Mava/blob/develop/tests/systems/maddpg_system_test.py

KaleabTessera (approved):

arnupretorius (approved):
Thanks @DriesSmit 🔥

PR #271, opened 2021-07-12
Feature/Enforce docstring code coverage.

KaleabTessera:
## What?
Added `interogate` package to check and enforce docstring code coverage. 
## Why?
Better quality code documentation. 
## How?
-
## Extra
Currently, the pre-commits will fail if docstring code coverage is less than 54.4% (our current docstring coverage is 54.5%). This will prevent new code without docstrings from being committed and we can increase this percentage as we go along. @arnupretorius Sounds good?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,3 +59,6 @@ jobs:
       - name: Test with pytest and get code coverage
         run: |
           pytest --cov-report term-missing --cov=mava  -n "$(grep -c ^processor /proc/cpuinfo)" tests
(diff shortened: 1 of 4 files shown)


DriesSmit (approved):
Thanks, @KaleabTessera :fire: This will definitely help us save time in the long run.

arnupretorius (approved):
Thanks @KaleabTessera! 👍

arnupretorius:
Sound good to me 👍  Will it check the structure though? E.g. a description of the function, plus each arg, and the return? Or is it only checking that each function has """some text"""?

PR #275, opened 2021-07-19
feature: working version of importance sampling on feedforward madqn.

jcformanek:
## What?
Implemented importance sampling (IS) / prioritized experience replay for feedforward madqn.
## Why?
Hopefully importance sampling will assist performance.
## How?
After each learning step we compute new priorities for the samples we drew from the replay buffer. This is done using  Q-value errors. The `mutate_priorities()` function is used to update the priorities in the reverb table.
## Extra
For now IS only works in feedforward madqn. I have also tested to make sure all of the other systems that inherit from feedforward MADQN still work. Recurrent MADQN, MADQN with comms, Dial, VDN and Qmix all still work after I made these changes. 

Because so many systems inherit from feedforward MADQN, it is quite hard to make changes without breaking the other systems. So my strategy mocing forward is going to be to make very incremental upgrades to MADQN and ensure at each step that nothing breaks the other systems.


DriesSmit (approved):
This looks great thanks @jcformanek :fire: So this PR should go in after the new adder PR?

KaleabTessera (commented):
Looks really great @jcformanek ! Thanks so much for this. The only thing left from my side is to benchmark this. After that, I am happy to approve.

jcformanek:
I am going to benchmark this on Flatland now.

arnupretorius (approved):
Nice @jcformanek! 🔥

PR #276, opened 2021-07-20
Feature/New acme adders and tests.

KaleabTessera:
## What?
- Upgraded acme adders and wrote tests.
- Enforced docstring styles.
- Wrote tests for all systems.
## Why?
- To use new version of acme.
## How?
- Updated adders.
## Extra
- V2 of this PR :laughing:


mava/adders/reverb/episode.py
```diff
@@ -29,25 +29,24 @@
 import tree
 from acme import specs, types
 from acme.adders.reverb import utils
-from acme.utils import tree_utils
```
DriesSmit, line 32: Did you upgrade the episode adder and just use tests to determine if it is probably working? We don't have any system implementation using the episode adder correct?

mava/systems/tf/maddpg/system.py
```diff
@@ -240,6 +244,7 @@ def __init__(
                 n_step=n_step,
                 sequence_length=sequence_length,
                 period=period,
+                bootstrap_n=bootstrap_n,
```
DriesSmit, line 247: Nice catch @KaleabTessera! So bootstrap_n could not be set before from the system interface?

mava/systems/tf/madqn/networks.py
```diff
@@ -61,13 +60,20 @@ def make_default_networks(
         Mapping[str, types.TensorTransformation]: returned agent networks.
     """
 
-    # Set Policy function and layer size
+    # Set Policy function and layer size.
+    # Default size per arch type.
     if archecture_type == ArchitectureType.feedforward:
+        if not policy_networks_layer_sizes:
```
DriesSmit, line 66: Nice I like that we can now have different defaults for feedforward and recurrent.

DriesSmit (approved):
Great work @KaleabTessera :smile: Thanks for the massive effort :fire:  Did you spot-check that the different properties like recurrent, state-based, centralised still work? I think I checked most of them though. So they should probably still work.

mava/adders/reverb/episode.py
KaleabTessera: Yeah, I just wrote the tests to check the adder works as expected.

KaleabTessera:
Thanks @DriesSmit ! 

I wrote tests for feedforward and recurrent systems, so they should work/run. I will quickly add smoke tests for state based, centralized and networked archs.

mava/adders/base.py
```diff
@@ -51,14 +51,14 @@ def add(
         self,
         actions: Dict[str, types.NestedArray],
         next_timestep: dm_env.TimeStep,
-        extras: Dict[str, types.NestedArray] = {"": ()},
+        next_extras: Dict[str, types.NestedArray] = {"": ()},
```
arnupretorius, line 54: Do we need `{"": ()}`? Can't we just have an empty dict?

mava/adders/reverb/base.py
```diff
@@ -196,73 +147,43 @@ def add(
...
+            next_step,
+            partial_step=True,
+        )
+        self._write()
+
         if next_timestep.last():
-            self._start_of_episode = False
-            self._write()
+            # Complete the row by appending zeros to remaining open fields.
+            # TODO(b/183945808): remove this when fields are no longer expected to be
```
arnupretorius, line 184: Do we want to keep this TODO?

mava/adders/reverb/base.py
```diff
@@ -54,6 +40,9 @@ class Step(NamedTuple):
     extras: Dict[str, types.NestedArray]
 
 
+Trajectory = Step
```
arnupretorius, line 43: Is there a reason we do this instead of calling the above class `Trajectory`?

mava/adders/reverb/episode.py
```diff
@@ -29,25 +29,24 @@
...
         self,
         client: reverb.Client,
         max_sequence_length: int,
         delta_encoded: bool = False,
-        chunk_length: Optional[int] = None,
         priority_fns: Optional[base.PriorityFnMapping] = None,
-        pad_end_of_episode: bool = True,
-        break_end_of_episode: bool = True,
-        max_in_flight_items: Optional[int] = 25,
+        max_in_flight_items: int = 1,
```
arnupretorius, line 48: I see the previous default for this was 25, do we want to set it to 1 now?

mava/adders/reverb/sequence.py
```diff
@@ -20,43 +20,45 @@
...
-from acme.utils import tree_utils
+from acme.adders.reverb.sequence import SequenceAdder
+from acme.types import NestedSpec
 
 from mava.adders.reverb import base
-from mava.adders.reverb import utils as mava_utils
+from mava.adders.reverb.base import ReverbParallelAdder
+from mava.adders.reverb.utils import trajectory_signature
 
+# TODO Clean this up, when using newer versions of acme.
```
arnupretorius, line 38: Are you referring to the version `0.2.1` here or newer than that, i.e. future versions?

mava/adders/reverb/sequence.py
```diff
@@ -20,43 +20,45 @@
...
-class ParallelSequenceAdder(base.ReverbParallelAdder):
+class ParallelSequenceAdder(SequenceAdder, ReverbParallelAdder):
     """An adder which adds sequences of fixed length."""
 
     def __init__(
         self,
         client: reverb.Client,
         sequence_length: int,
         period: int,
+        *,
```
arnupretorius, line 53: What is the purpose of this argument?

mava/adders/reverb/sequence.py
```diff
@@ -65,183 +67,83 @@ def __init__(
...
-            buffer_size=sequence_length,
-            max_sequence_length=sequence_length,
+            # We need an additional space in the buffer for the partial step the
+            # base.ReverbAdder will add with the next observation.
+            max_sequence_length=sequence_length + 1,
             delta_encoded=delta_encoded,
-            chunk_length=chunk_length,
             priority_fns=priority_fns,
-            max_in_flight_items=max_in_flight_items,
+            max_in_flight_items=max_in_flight_items,  # type: ignore
```
arnupretorius, line 95: Why is there a type mismatch here? Would it not always be an `int`?

mava/adders/reverb/sequence.py
```diff
@@ -65,183 +67,83 @@ def __init__(
...
-            # ... or every `period`th time after hitting max length.
-            (
-                self._step > self._max_sequence_length
-                and (self._step - self._max_sequence_length) % self._period == 0
-            )
-        ):
+        if not first_write and not period_reached and not force:
             return
 
+        # TODO(b/183945808): will need to change to adhere to the new protocol.
```
arnupretorius, line 121: Do we want to keep this?

mava/adders/reverb/sequence.py
```diff
@@ -65,183 +67,83 @@ def __init__(
...
 
     @classmethod
     def signature(
         cls,
         environment_spec: specs.EnvironmentSpec,
-        extras_spec: tf.TypeSpec = {},
+        sequence_length: Optional[int] = None,
+        extras_spec: NestedSpec = (),
     ) -> tf.TypeSpec:
-        """This is a helper method for generating signatures for Reverb tables.
```
arnupretorius, line 207: Need to update the docstring here and a few other places. But I guess the docstring checker will catch this in future PRs? ;)

mava/adders/reverb/transition.py
```diff
@@ -83,7 +83,9 @@ def __init__(
         client: reverb.Client,
         n_step: int,
         discount: float,
+        *,
```
arnupretorius, line 86: So I saw online: "The * indicates the end of the positional arguments. Every argument after that can only be specified by keyword." Just wondering why we are being explicit about it?

mava/adders/reverb/transition.py
```diff
@@ -100,153 +102,151 @@ def __init__(
...
 
-        # Give the same tree structure to the n-step return accumulator,
-        # n-step discount accumulator, and self.discount, so that they can be
-        # iterated in parallel using tree.map_structure.
+        # Note: this getter is meant to be used on a TrajectoryWriter.history to
+        # obtain its numpy values.
+        def get_all_np(x: np.array) -> np.array:
+            return x[self._first_idx : self._last_idx].numpy()
+
+        # get_all_np = lambda x: x[self._first_idx : self._last_idx].numpy()
```
arnupretorius, line 132: Remove?

mava/systems/tf/mad4pg/training.py
```diff
@@ -469,15 +482,15 @@ def __init__(
...
     # Forward pass that calculates loss.
-    def _forward(self, inputs: Any) -> None:
+    def _forward(self, inputs: reverb.ReplaySample) -> None:
         """Trainer forward pass
 
         Args:
             inputs (Any): input data from the data table (transitions)
         """
 
         # TODO: Update this forward function to work like MAD4PG
```
arnupretorius, line 492: Do we want to keep this TODO?

mava/wrappers/open_spiel.py
```diff
@@ -17,11 +17,15 @@
 
 import dm_env
 import numpy as np
-import pyspiel  # type: ignore
+
+try:
+    import pyspiel  # type: ignore
+    from open_spiel.python import rl_environment  # type: ignore
+except ModuleNotFoundError:
```
arnupretorius, line 24: Nice, not sure if this is the reason, but we should perhaps consider this for other environments as well for when they haven't been installed as dependencies.

arnupretorius (commented):
Great work! Thanks @KaleabTessera!! 🔥 Huge effort! 💪 

Please just see my few comments. Most of them are quite minor.

mava/adders/base.py
KaleabTessera: Sure, I will change this.

mava/adders/reverb/base.py
KaleabTessera: Will remove.

mava/adders/reverb/base.py
KaleabTessera: I also found this a bit weird. I think acme is planning on possibly having the trajectory class change in the future, but for now it is the same as the `Step` class. I think we should leave it for now incase the Trajectory class in the near future. Thoughts?

mava/adders/reverb/episode.py
KaleabTessera: For consistency with acme's episode adder (https://github.com/deepmind/acme/blob/master/acme/adders/reverb/episode.py).

mava/adders/reverb/base.py
arnupretorius: Sounds good :)

mava/adders/reverb/base.py
arnupretorius: Ok makes sense, not sure if this was from our side or theirs. But happy to keep it as is for now.

mava/adders/reverb/sequence.py
KaleabTessera: Future versions. 

In the latest codebase (not yet a package), `EndOfEpisodeBehavior` was renamed to `EndBehavior`. If we simply import `EndOfEpisodeBehavior` , it will break in `0.2.2`. Importing `EndBehavior` currently breaks since it was added after `0.2.1`.

mava/systems/tf/mad4pg/training.py
KaleabTessera: Not sure, @DriesSmit do we still need this?

mava/wrappers/open_spiel.py
KaleabTessera: I did a quick pass and did this for the other environments.

mava/adders/reverb/sequence.py
KaleabTessera: Removed.

mava/adders/reverb/sequence.py
KaleabTessera: It is a mismatch between `Optional[int]` and `int`. Sorted it out in new commit though.

mava/adders/reverb/transition.py
KaleabTessera: My understanding of `*,` is that it is useful when they are multiple arguments and you want to split "common" arguments from "less common" arguments. 

The arguments before `*,` can be provided positionally and that usually makes sense for parms that are common to a function call. The params after `*,` , are less common and would be difficult to keep track of their positional spot since there are a lot of them and so are keyword args.

mava/adders/reverb/sequence.py
KaleabTessera: Addressed in other comment.

arnupretorius (approved):
Thanks @KaleabTessera! 🥇

PR #279, opened 2021-07-26
Feature/double q learning recurrent madqn

jcformanek:
## What?
Implemented Double Q-learning for Recurrent MADQN. Also implemented prioritized experience replay for recurrent MADQN. 

## Why?
Double Q-learning helps reduce Q-value overestimation in DQN. Prioritized experience replay has been shown to speed up learning.

## How?
PER and Double Q-learning were implemented in a similar way to the feedforward case.


mava/systems/tf/madqn/training.py
```diff
@@ -593,65 +626,131 @@ def _forward(self, inputs: Any) -> None:
...
-            q_network_losses: Dict[str, NestedArray] = {}
 
+            # Unroll over time dimension.
+            T = list(observations.values())[0].observation.shape[0]  # time dimension
+            q_stacks: Dict[str, list] = {agent: [] for agent in self._agents}
+            q_target_stacks: Dict[str, list] = {agent: [] for agent in self._agents}
+            for t in range(T):
+
+                # Step each agent.
+                for agent in self._agents:
```
DriesSmit, line 663: Why is two loops over the agents needed for recurrent DQN instead of one loop (e.g. MADDPG)?

DriesSmit (commented):
This looks great @jcformanek! :fireworks: You are really pushing MA-DQN forward :fire: :fire:  Does recurrent MADQN with its enhancements train on Flatland? Also just see my one other comment.

mava/systems/tf/madqn/training.py
jcformanek: It is because I want to build communication directly into here instead of having a separate trainer for it and when doing communication, at every time step each agent sends a message.

jcformanek: With regard to experiments on Flatland: I am setting things up on GCP now. Will hopefully have some results in a couple hours.

KaleabTessera (commented):
@jcformanek Thanks for pushing madqn to new heights! :rocket: 

We need to benchmark the performance (walltime) of these changes before we get them into Mava.

PR #283, opened 2021-07-28
fix/ Smac Load

KaleabTessera:
## What?
- Fix SC2 load errors. 
- Make system tests use one executor for more consistency. 
## Why?
-
## How?
-
## Extra
-


arnupretorius (approved):
Thanks @KaleabTessera! 👍

Issue #282, opened 2021-07-26
How to render SMAC on Linux?


mnguyen0226:
Thank you for providing the examples script for SMAC. There has been a lot of problem in rendering SMAC via Linux and it seems like the way that the author on SMAC suggested to do is to trained agent on PyMARL then save and render on Window or Wine for linux. 

I installed both SMAC and pysc2 on my linux and was able to run PySC2 example "python -m pysc2.bin.agent --map Simple64" with rendering screen but **not** for SMAC example "python -m smac.examples.random_agents".

I was able to run your SMAC's 2 examples scripts and saw their improvement via Tensorboard. However, there is no rendering screen. 

I acknowledge that Mava use Launchpad for distribution process which is available only for Linux for now. W**ill you provide steps on how to provide this .gif/.HTML or render on Linux?**

Here is the .gif/.HTML I mentioned: https://github.com/instadeepai/Mava/blob/develop/docs/images/sc2_animation.gif

Thank you

KaleabTessera:
Hi @mnguyen0226 , seems like a similiar issue to https://github.com/instadeepai/Mava/issues/280 . Please follow step 4 from the [Optional Install doc](https://github.com/instadeepai/Mava/blob/develop/OPTIONAL_INSTALL.md), in Ubuntu (or Linux based systems) this should allow you to record agents acting in the environment. 

Please let us know if you still have issues after the install.

KaleabTessera:
Closing and continuing the discussion in issue https://github.com/instadeepai/Mava/issues/280 .

PR #284, opened 2021-07-29
Update Dockerfile by uninstall enum34

mnguyen0226:
## What?
This modification related to the issue when install Docker for SMAC: https://github.com/instadeepai/Mava/issues/282

I tried to install Docker to both Python 3.6 and Python 3.7 and figure out that pysc2 needed as well as uninstall enum34 is needed before pip install oxwhiri/smac github repo to prevent enum-Python 3.6 error. Note that there are suggestion for update enum instead of uninstall but this will not work

## Why?
[Describe why the change was made.]
To be able to install Docker on Python 3.6, 3.7 correctly

## How?
[Describe how the change was made i.e. describe technically what the change does.]
Shown below

## Extra
[Any extra information.]
Also, I suggested to update the main Docker repo to up-to-date with the develop one since it does not have uninstall enum and it cause problem in install Docker.

I know that develop is a default git clone but many user might still try to use main branch instead


DriesSmit:
Thanks for the PR @mnguyen0226 ! @KaleabTessera does these changes seem fine?

KaleabTessera:
Thanks for your first PR @mnguyen0226 ! I am happy to accept the dockerfile changes, but we can't add the SMAC binaries. 

Can you please either create a new PR with just the dockerfile changes or remove the binary file changes & example changes from this PR ?

mnguyen0226:
@KaleabTessera  Awesome! I have made a new pull request accordingly.

KaleabTessera:
Duplicate of https://github.com/instadeepai/Mava/pull/286 .

Issue #287, opened 2021-08-05
Typo on simple_spread debugging env


mlanas:
Hi,
I was playing around with the simple_spread debugging environment and found a typo in the observation function, specifically in the landmark distance calculation. Instead of subtracting the agent position the same landmark position is being subtracted resulting on the last 4 values of the observation vector always being zero.

https://github.com/instadeepai/Mava/blob/e6d0ea8d0b1e7e02d19f2cc89446948d8de74888/mava/utils/debugging/scenarios/simple_spread.py#L113

arnupretorius:
Hi @mlanas, thanks for spotting this! 🤗  

We will look into it.

DriesSmit:
Good catch @mlanas ! :fire: I will create a PR correcting this. Just make sure that you agree with the changes :+1:

mlanas:
Thanks @DriesSmit ! The fix looks good and the code is much more clear now. :+1:

PR #288, opened 2021-08-05
Bugfix: Simple_spread observation code.

DriesSmit:
## What?
Fix a bug in the observation calculation for the debugging environment, as pointed out by @mlanas in #287. 
## Why?
Bugfix.
## How?
Fixed a landmark calculation typo error in `utils/scenarios/simple_spread.py`.


arnupretorius (approved):
Thanks @DriesSmit! 😄

PR #300, opened 2021-08-18
fix: Autorom manual download.

KaleabTessera:
## What?
Autorom temp fix. 
## Why?
-
## How?
Manual install atari roms. 
## Extra
-


DriesSmit (commented):
This looks great! Thanks @KaleabTessera :+1: I really like the `lp_utils.cpu_only(program)` code. Can you maybe do the same for the examples? Something like `lp_utils.gpu_trainer_only(program)` should probably work.

mava/utils/lp_utils.py
```diff
@@ -18,17 +18,34 @@
...
 
 import launchpad as lp
 from absl import flags, logging
 from acme.utils import counting
+from launchpad.nodes.python.local_multi_processing import PythonProcess
 
 FLAGS = flags.FLAGS
 
 
+def cpu_only(program: Any) -> Dict:
```
arnupretorius, line 31: Nice 🔥

arnupretorius (approved):
Thanks @KaleabTessera! 👍

arnupretorius:
> This looks great! Thanks @KaleabTessera 👍 I really like the `lp_utils.cpu_only(program)` code. Can you maybe do the same for the examples? Something like `lp_utils.gpu_trainer_only(program)` should probably work.

We could also perhaps think about making a `to_cpu` and `to_gpu` where you send in only the list of nodes you want to have run on those devices (instead of the entire program).

DriesSmit (commented):
Can we do the examples conversions here as well? It should be fast right? Or do you want to leave it for future work?

KaleabTessera:
@DriesSmit I am still adding the example conversions :)

DriesSmit:
> @DriesSmit I am still adding the example conversions :)

Thanks @KaleabTessera :+1:

examples/debugging/simple_spread/feedforward/centralised/run_maddpg.py
```diff
@@ -92,13 +91,9 @@ def main(_: Any) -> None:
...
-    gpu_id = -1
-    env_vars = {"CUDA_VISIBLE_DEVICES": str(gpu_id)}
-    local_resources = {
-        "trainer": [],
-        "evaluator": PythonProcess(env=env_vars),
-        "executor": PythonProcess(env=env_vars),
-    }
+    local_resources = lp_utils.to_gpu(
+        program_nodes=program.groups.keys(), nodes_on_gpu=["trainer"]
+    )
```
DriesSmit, line 96: This looks great :raised_hands: :fire: Thanks @KaleabTessera!

DriesSmit (approved):

mava/utils/lp_utils.py
```diff
@@ -28,18 +28,33 @@
...
+
+def to_gpu(program_nodes: List, nodes_on_gpu: List = ["trainer"]) -> Dict:
+    """Specifies which nodes should run on gpu.
+
+    Args:
+        program_nodes (List): nodes in lp program.
+        nodes_on_gpu (List, optional): nodes to run on gpu. Defaults to ["trainer"].
 
     Returns:
         Dict: dict with cpu only lp config.
```
arnupretorius, line 51: Edit :)

tests/systems/dial_system_test.py
```diff
@@ -65,7 +65,7 @@ def test_recurrent_dial_on_debugging_env(self) -> None:
         trainer_node.disable_run()
 
         # Launch gpu config - don't use gpu
-        local_resources = lp_utils.cpu_only(program)
+        local_resources = lp_utils.cpu_only(program.groups.keys())
```
arnupretorius, line 68: I know I suggested feeding in the list, but with the `nodes_on_gpu` list as an additional argument (specifying the subset of the full list), I think it is cleaner to feed in only the program and keep the `for node in program.groups.keys()` inside the `to_gpu`. 😄  That way, the `cpu_only` looks nice and clean as well. Also, one suggestion is perhaps to have a single utile called `to_device` which defaults to all cpu if the `nodes_on_gpu` is empty. What do you think @KaleabTessera? Happy either way, since the former is more clear when ready the code but the later is more compact.

arnupretorius (commented):
Nice @KaleabTessera!! 😄  Much cleaner than before! Just see my few comments.

tests/systems/dial_system_test.py
KaleabTessera: I think having `cpu_only` and `to_gpu` taking in a list might be the better option, mainly because it is easier to test. We don't have to initialize a program to test the functions works. From my perspective, passing `program` or `program.groups.keys()` seems quite similar in terms of cleaness.

 For the `to_device` function, it sounds like a good idea. I will update the code :)

PR #313, opened 2021-09-08
Fix per agent loggers

DriesSmit:
## What?
Fix step loggers to only log at the end of the episode.
## Why?
The per agent loggers are doing an unnecessary amount of computation and are more noise than they have to be for the reward logging. The logger charts also did not seem to align in time but this seems to now be fixed.
## How?
Update the per agent loggers to only write once at the end of each episode.
## Extra
.

### DIFF
--- a/mava/utils/wrapper_utils.py
+++ b/mava/utils/wrapper_utils.py
@@ -172,6 +172,7 @@ class RunningStatistics:
     a specific quantity.
     """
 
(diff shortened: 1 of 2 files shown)


arnupretorius (approved):
Thanks @DriesSmit 👍

KaleabTessera (approved):
Thanks @DriesSmit , this looks good :+1:

PR #310, opened 2021-09-01
Feature/ Smac wrapper Update, MADQN/QMIX/VDN upgrades and Dockerfile improvements

KaleabTessera:
## What?
- Smac & Wrapper:
   - Switched to PZ SMAC wrapper (https://github.com/oxwhirl/smac/tree/master/smac/examples/pettingzoo).
   - Minor bug fixes to PZ core wrapper.
- MADQN/QMIX/VDN:
   - Updates and minor fixes to action selectors.
   - Moved epsilon decay out of trainers to executors and evaluators. 
   - Added support for timestep based epsilon decay e.g. decay for 50000 time steps. 
- All Systems:
   - Added optional LR decay and examples of usage.   
- Other:
  -  Redesigned dockerfiles (i.e. dockerfile per env) & makefiles. @DriesSmit @arnupretorius @mmorris44 Please check the redesigned dockerfiles :) 
  -  Tested and updated versions of most packages. 
  - Changed how github actions are done. 
## Why?
 - Fixes and more features. 
## How?
-
## Extra
-

PR #323, opened 2021-10-18
Feature/eval intervals

KaleabTessera:
## What?
- Added eval interval so that evaluation loops can run at certain intervals.
- Minor changes for new version of mypy. 
## Why?
- To be able to schedule evaluator runs. 
## How?
- Added extra condition to environment loop - to check if interval exists. 
## Extra
- @DriesSmit @arnupretorius Most of the file changes were mypy changes (for new version of mypy) so you can focus on looking the environment loop code. 
- Also note this is a PR into `mava-scaling` and not into `develop`. 
- I ran all test and checks locally and they pass :+1:


mava/environment_loop.py
```diff
@@ -338,6 +348,36 @@ def _compute_episode_statistics(
     ) -> None:
         pass
 
+    def get_counts(self) -> Any:
+        if hasattr(self._executor, "_counts"):
```
arnupretorius, line 352: I didn't see this being used anywhere? I.e. executor examples that have this attribute. What is the use case for this?

arnupretorius (dismissed):
Thanks @KaleabTessera! Going to be super useful for comparisons. 🔥 

Just left a few minor suggestions.

mava/systems/tf/maddpg/system.py
```diff
@@ -533,7 +543,7 @@ def evaluator(
 
         # Create the agent.
         executor = self._builder.make_executor(
-            # executor_id="evaluator",
+            evaluator=True,
```
DriesSmit, line 546: Nice simplification.

mava/environment_loop.py
```diff
@@ -481,10 +494,46 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
                 num_steps is not None and step_count >= num_steps
             )
 
+        def should_run_loop(eval_condtion: Tuple) -> bool:
```
DriesSmit, line 497: Why do we define this function inside the run function and not inside the ParallelEnvironmentLoop class?

DriesSmit: And also the should_terminate function. Maybe there is a reason for these functions being here?

mava/environment_loop.py
```diff
@@ -481,10 +494,46 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
...
+            if (not environment_loop_schedule) or should_run_loop(eval_condtion):
+                result = self.run_episode()
+                episode_count += 1
+                step_count += result["episode_length"]
+                # Log the given results.
+                self._logger.write(result)
+
+            # We need to get the latest counts if we are using eval intervals.
+            if environment_loop_schedule:
+                self._executor.update()
```
DriesSmit, line 539: Is this update not already performed in the run_episode call?

DriesSmit (commented):
Thanks @KaleabTessera the changes looks great :rocket: Just see my few comments.

mava/environment_loop.py
```diff
@@ -338,6 +348,36 @@ def _compute_episode_statistics(
...
         pass
 
+    def get_counts(self) -> Any:
+        if hasattr(self._executor, "_counts"):
+            counts = self._executor._counts
+        else:
+            counts = self._counter.get_counts()
+        return counts
+
+    def record_counts(self, episode_steps: int) -> Any:
```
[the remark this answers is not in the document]
KaleabTessera: So I updated this. It returns a `counting.Counter` object.

mava/environment_loop.py
```diff
@@ -338,6 +348,36 @@ def _compute_episode_statistics(
     ) -> None:
         pass
 
+    def get_counts(self) -> Any:
```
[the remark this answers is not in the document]
KaleabTessera: Updated to `counting.Counter`.

mava/environment_loop.py
KaleabTessera: So this is for systems that haven't been moved to mava scaling yet. These systems use `_counts` .

mava/environment_loop.py
KaleabTessera: So this is a side-effect of our environment loop being based on [acme's one ](https://github.com/deepmind/acme/blob/881f1b01c01a73e0a0df9bd0ac5c205028ca1b31/acme/environment_loop.py#L152). 

From a function level, the only real difference I see is you don't need to call `self.[function]`. Also might be slightly faster being called this way, although I am not sure about that.

mava/environment_loop.py
KaleabTessera: So this update is important to get the latest counts.

Since these loops run on different processes, while the evaluator is waiting, it doesn't run the loop and so it vars counts don't get updated (this part doesn't [run](https://github.com/instadeepai/Mava/blob/52cd450c08e695380b81023f45228fe7a4e43749/mava/environment_loop.py#L439)). 

So while the evaluator is waiting, it still need to call `self._executor.update()` to get latest counts from the executor runs. 

Not sure if that makes sense?

KaleabTessera:
From my side this can be merged in. The issues with the github actions/ package versions are resolved in this PR - https://github.com/instadeepai/Mava/pull/310.

mava/environment_loop.py
DriesSmit: I see thanks. We can keep it as is :)

mava/environment_loop.py
DriesSmit: I see. That makes sense yes. Should there not maybe be a time.sleep somewhere in here? This is so that the evaluator does not constantly speak to the variable_server while waiting. And also `self._executor.update()` only updates the executor every 1000 steps right? So we maybe need to change that setting to 1 in the case of using an evaluator that has waiting intervals?

mava/utils/training_utils.py
```diff
@@ -1,11 +1,24 @@
 import os
+import time
 from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple, Union
 
 import sonnet as snt
 import tensorflow as tf
 import trfl
 
 
+def non_blocking_sleep(time_in_seconds: int) -> None:
```
DriesSmit, line 10: Nice. I like this.

mava/environment_loop.py
```diff
@@ -481,10 +494,50 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
...
-            episode_count += 1
-            step_count += result["episode_length"]
-            # Log the given results.
-            self._logger.write(result)
+            if (not environment_loop_schedule) or should_run_loop(eval_condtion):
+                result = self.run_episode()
+                episode_count += 1
+                step_count += result["episode_length"]
+                # Log the given results.
+                self._logger.write(result)
```
DriesSmit, line 539:
```suggestion
     if (not environment_loop_schedule) or should_run_loop(eval_condtion):
                result = self.run_episode()
                episode_count += 1
                step_count += result["episode_length"]
                # Log the given results.
                self._logger.write(result)
     else:
               time.sleep(1)
```

mava/environment_loop.py
```diff
@@ -481,10 +494,50 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
...
+
```
DriesSmit, line 540:
```suggestion
else:
               time.sleep(1)
```

mava/environment_loop.py
```diff
@@ -481,10 +494,50 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
...
+        if environment_loop_schedule:
+            eval_condtion = check_count_condition(self._executor._interval)
+
         while not should_terminate(episode_count, step_count):
-            result = self.run_episode()
-            episode_count += 1
-            step_count += result["episode_length"]
-            # Log the given results.
-            self._logger.write(result)
+            if (not environment_loop_schedule) or should_run_loop(eval_condtion):
```
DriesSmit, line 534:
```suggestion
            if (not environment_loop_schedule) or should_run_loop(eval_condition):
```

mava/environment_loop.py
```diff
@@ -481,10 +494,50 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
...
+            return should_run_loop
+
         episode_count, step_count = 0, 0
+
+        # Currently, we only use intervals for eval loops.
+        environment_loop_schedule = (
+            self._executor._evaluator and self._executor._interval
+        )
+        if environment_loop_schedule:
+            eval_condtion = check_count_condition(self._executor._interval)
```
DriesSmit, line 531:
```suggestion
            eval_condition = check_count_condition(self._executor._interval)
```

mava/environment_loop.py
```diff
@@ -481,10 +494,50 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
                 num_steps is not None and step_count >= num_steps
             )
 
+        def should_run_loop(eval_condtion: Tuple) -> bool:
+            """Check if the eval loop should run in current step.
+
+            Args:
+                eval_condtion : tuple containing interval key and count.
```
DriesSmit, line 501:
```suggestion
                eval_condition : tuple containing interval key and count.
```

mava/environment_loop.py
```diff
@@ -481,10 +494,50 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
...
+
+            Returns:
+                a bool indicatings if eval should run.
+            """
+            should_run_loop = False
+            eval_interval_key, eval_interval_count = eval_condtion
```
DriesSmit, line 507:
```suggestion
            eval_interval_key, eval_interval_count = eval_condition
```

DriesSmit (approved):
Great work @KaleabTessera :rocket:

PR #328, opened 2021-11-08
Fix Flatland package error in Docker build

DriesSmit:
## What?
Fix an error where the building of Mava crashes when trying to build Flatland.
## Why?
Fix a bug in the building of Mava.
## How?
Manually install the package `pyparsing` before installing Flatland.
## Extra
.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,6 @@ ENV SC2PATH /home/app/mava/3rdparty/StarCraftII
 COPY . /home/app/mava
 RUN python -m pip uninstall -y enum34
 RUN python -m pip install --upgrade pip
-RUN pip install pyparsing==3.0.3
 RUN python -m pip install -e .[flatland]
 RUN python -m pip install -e .[open_spiel]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -40,6 +40,7 @@ ENV SC2PATH /home/app/mava/3rdparty/StarCraftII
 COPY . /home/app/mava
 RUN python -m pip uninstall -y enum34
 RUN python -m pip install --upgrade pip
+RUN pip install pyparsing==3.0.3
```
mmorris44, line 43: Could we rather enforce this in `setup.py`, under `flatland-requirements`, instead of in the Dockerfile?
Seems a little neater to me.

DriesSmit: Fixed.

DriesSmit: Unfixed.

mmorris44 (approved):

arnupretorius (approved):
Thanks @DriesSmit! :)

PR #334, opened 2021-11-24
Fix RoboCup environment wrapper

DriesSmit:
## What?
Update RoboCup to work with the latest Mava version.
## Why?
The RoboCup example is currently broken. This small change fixes the example.
## How?
Update the state name that gets passed to the agents to the correct expected state name.


mmorris44 (approved):

KaleabTessera (approved):
Thanks @DriesSmit !

PR #337, opened 2021-11-30
Feature/apex flatland

paulcaron:
## What?
- Create new multi-agent exploration schedulers using Ape-X.
- Add scripts to run multi-agent Ape-X on flatland.
## Why?
This update allows to use new exploration schedulers.
## How?
Created new functions in exploration_scheduling.py

### DIFF
--- a/examples/flatland/feedforward/decentralised/run_apex.py
+++ b/examples/flatland/feedforward/decentralised/run_apex.py
@@ -30,6 +30,7 @@ from mava.systems.tf import madqn
 from mava.utils import lp_utils
 from mava.utils.environments.flatland_utils import flatland_env_factory
 from mava.utils.loggers import logger_utils
+from mava.wrappers.environment_loop_wrappers import MonitorParallelEnvironmentLoop
 
 FLAGS = flags.FLAGS
 
@@ -74,7 +75,7 @@ def main(_: Any) -> None:
     )
 
     # Checkpointer appends "Checkpoints" to checkpoint_dir
-    checkpoint_dir = f"{FLAGS.base_dir}/{FLAGS.mava_id}"
+    checkpoint_dir = f"{FLAGS.base_dir}/apex-{FLAGS.mava_id}"
 
     # Log every [log_every] seconds.
     log_every = 10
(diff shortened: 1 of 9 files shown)


examples/flatland/feedforward/decentralised/run_apex.py
```diff
@@ -0,0 +1,123 @@
...
+
+def main(_: Any) -> None:
+    """Run program
+
+    Args:
+        _ (Any): Arguments
+    """
+
+    # Environment.
+    environment_factory = functools.partial(
```
vcoyette, line 68: Any reason why you didn't use lp_utils.partial_kwargs instead ? (like below)
Also, `include_agent_info=False` is the default, so unless you want to make it very explicit that we set it to False, you don't have to put it here

examples/flatland/feedforward/decentralised/run_madqn.py
```diff
@@ -86,37 +87,48 @@ def main(_: Any) -> None:
...
+    #             epsilon_start=1.0, epsilon_min=0.08, epsilon_decay=4e-4
+    #         ),
+    #         "train_1": LinearExplorationScheduler(
+    #             epsilon_start=1.0, epsilon_min=0.09, epsilon_decay=5e-4
+    #         ),
+    #         "train_2": LinearExplorationScheduler(
+    #             epsilon_start=1.0, epsilon_min=0.10, epsilon_decay=6e-4
+    #         ),
+    #     },
+    # },
```
vcoyette, line 115: Try to avoid commenting code, if it is not needed anymore you can just delete it. If is needed again some time, you  can still find it in git history

mava/components/tf/modules/exploration/exploration_scheduling.py
```diff
@@ -237,14 +237,89 @@ def decrement_epsilon(self) -> float:
...
+
+    The schedulers are assigned in increasing order to all executor.
+    Within every scheduler, the espilons are randomy assigned to agents.
+
+    Returns:
+        Monotonic Multi-Agent Ape-X exploration scheduler exploration_scheduler_fn
+
+    """
+    num_agents = len(agent_ids)
+    exploration_scheduler_fn: Dict = {}
```
vcoyette, line 278: This might be a good place to use a `defaultdict(dict)`

vcoyette:
One general remark, it seems that there is a lot of duplicated code between the different examples scripts. I may be missing something, but I think almost only the `exploration_scheduler_fn` is changing. Maybe could you use only one file, and specify which exploration scheduler should be used as a Flag ? It would be easier to maintain in the future

vcoyette:
I left some minor comments, but the PR looks good to me ! 
Of course I don't know MAVA code at all, so I let the team correct if I said anything wrong. Other than that congrats Paul ! :tada:

examples/flatland/feedforward/decentralised/run_apex.py
paulcaron: Good question. @KaleabTessera, is there any reason to use `functools.partial` instead of `lp_utils.partial_kwargs` as in robocup?

mava/components/tf/modules/exploration/exploration_scheduling.py
paulcaron: Done.

PR #342, opened 2021-12-13
Feature/auto-push-docker-images and version upgrades

KaleabTessera:
## What?
- Added a github action to auto-push docker images
- Updates to package version to fix `ModuleNotFoundError: No module named 'chex'` .
## Why?
-
## How?
- Updated github actions file.
- Updated setup.py to change auto install chex and updated package versions. 
## Extra
-


.github/workflows/ci.yaml
```diff
@@ -6,6 +6,7 @@ on:
   pull_request:
     branches: [develop]
 
+# TODO(Kale-ab) Specify mava container to run tests on.
```
mmorris44, line 9: TODO still un-done. Is this for a future PR?

.github/workflows/ci_main.yaml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches: [main]
 
+# TODO(Kale-ab) Specify mava container to run tests on.
```
mmorris44, line 7: Similar un-done TODO

Dockerfile
```diff
@@ -1,6 +1,6 @@
 ##########################################################
 # Core Mava image
-FROM nvcr.io/nvidia/cuda:11.4.2-cudnn8-runtime-ubuntu20.04 as mava-core
+FROM registry.kao.instadeep.io/library/nvidia/cuda:11.4.2-cudnn8-runtime-ubuntu20.04 as mava-core
```
mmorris44, line 3: Is this standard practice for us going forward? To use a InstaDeep-hosted image rather than an nividia one?

Makefile
```diff
@@ -41,7 +41,7 @@ else ifneq (,$(findstring meltingpot,$(example)))
 DOCKER_IMAGE_TAG=meltingpot
 endif
 
-IMAGE = $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)
+IMAGE = $(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)-latest
```
mmorris44, line 44: I was quite happy with not having the `latest` tag, but this is just a personal preference

mmorris44 (changes requested):
Just some small notes. Otherwise looks great!

.github/workflows/ci.yaml
KaleabTessera: Yes. Once we have an instadeepct dockerhub account, I will pull a container from there in the workflow. Currently, we are using my personal dockerhub so I haven't added it here.

.github/workflows/ci_main.yaml
KaleabTessera: Commented above.

mmorris44 (approved):

PR #343, opened 2021-12-17
Updated pip installation instructions in README

RuanJohn:
## What?
Update the `pip install` command for installing dependencies from source
## Why?
It did not include `#egg=id-mava[reverb,tf,launchpad]`
## How?
I just added it in.


KaleabTessera (approved):
Thanks @RuanJohn !

PR #344, opened 2022-01-05
Bugfix/dockerfile no module found

KaleabTessera:
## What?
Changed the dockerfile to use virtual envs. 
## Why?
Recent dockerfiles had issues with finding the mava package. 
Error : 
```
  File "examples/petting_zoo/sisl/multiwalker/feedforward/decentralised/run_mad4pg.py", line 26, in <module>
    from mava.systems.tf import mad4pg
ModuleNotFoundError: No module named 'mava'
```
## How?
Updated dockerfile. 
## Extra
- Also switched the dockerhub repo that images get pushed to. 
- Added info on how to use pre-built docker images from dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ##########################################################
 # Core Mava image
-FROM nvidia/cuda:11.4.2-cudnn8-runtime-ubuntu20.04 as mava-core
+FROM registry.kao.instadeep.io/library/nvidia/cuda:11.4.2-cudnn8-runtime-ubuntu20.04 as mava-core
 # Flag to record agents
 ARG record
 # Ensure no installs try launch interactive screen
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,12 @@ ARG record
 # Ensure no installs try launch interactive screen
 ARG DEBIAN_FRONTEND=noninteractive
 # Update packages
-RUN apt-get update -y && apt-get install -y python3-pip && apt-get install -y python3-venv
+RUN apt-get update -y && apt-get install -y python3-pip
 # Update python path
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 10 &&\
     rm -rf /root/.cache && apt-get clean
 # Setup virtual env
+RUN python -m pip install --user virtualenv
 RUN python -m venv mava
 ENV VIRTUAL_ENV /mava
 ENV PATH /mava/bin:$PATH
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ COPY . /home/app/mava
 # For box2d
 RUN apt-get install swig -y
 ## Install core dependencies.
-RUN python -m pip install -e .[reverb,launchpad]
+RUN pip install -e .[reverb,launchpad]
 ## Optional install for screen recording.
 ENV DISPLAY=:0
 RUN if [ "$record" = "true" ]; then \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,17 +1,20 @@
 ##########################################################
 # Core Mava image
-FROM registry.kao.instadeep.io/library/nvidia/cuda:11.4.2-cudnn8-runtime-ubuntu20.04 as mava-core
+FROM nvidia/cuda:11.4.2-cudnn8-runtime-ubuntu20.04 as mava-core
```
mmorris44, line 3: I thought we were switching away from externally hosted images in favour of ones hosted within InstaDeep?

Dockerfile
```diff
@@ -1,17 +1,20 @@
...
 # Ensure no installs try launch interactive screen
 ARG DEBIAN_FRONTEND=noninteractive
 # Update packages
-RUN apt-get update -y && apt-get install -y python3-pip
+RUN apt-get update -y && apt-get install -y python3-pip && apt-get install -y python3-venv
 # Update python path
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 10 &&\
     rm -rf /root/.cache && apt-get clean
-# Upgrade pip
-RUN python -m pip install --upgrade pip
```
mmorris44, line 14: I like that we're moving away from having a global env in the container to a virtual env.

Just double checking that this won't cause problems with how we run anything? Off the top of my head, it should all be fine because the `PATH` includes the venv, but I'm concerned that this will break some fringe way we have of running things.

Dockerfile
```diff
@@ -1,17 +1,20 @@
...
+# Setup virtual env
+RUN python -m venv mava
+ENV VIRTUAL_ENV /mava
+ENV PATH /mava/bin:$PATH
+RUN pip install --upgrade pip setuptools wheel
```
mmorris44, line 17: Does the venv not have to be activated before this?

mmorris44 (commented):
A few minor concerns. Looks good otherwise!

Dockerfile
KaleabTessera: The internally hosted image was down yesterday so I switched back the official nvidia cuda one. Previously, we had issues when making a lot of pulls to the official nvidia one, if those issues pop up again, we can switch back to an internally hosted one.

Dockerfile
KaleabTessera: So in docker, this part 
```
ENV VIRTUAL_ENV /mava
ENV PATH /mava/bin:$PATH
```
activates the virutal env. Similiar to `source mava/bin/activate` , but it activates the env for whole dockerfile and not just a specific command.

Dockerfile
KaleabTessera: I have tested the mava tf core, pz and flatland images (the images impacted by this change). They all seem to run and work with the new changes :+1:

mmorris44 (approved):

arnupretorius (approved):
Thanks @KaleabTessera! :) Looks good. We can look at a export path based solution perhaps sometime in the future. But for now, since this works we keep it.

README.md
```diff
@@ -169,6 +169,17 @@ All modules in Mava aim to work in this way.
...
 
 ### Docker (**Recommended**)
+#### Using pre-built images
+You can pull & run the latest pre-built images from our [DockerHub](https://hub.docker.com/repository/docker/instadeepct/mava) by specifying the docker image and example/file you want to run.
+
+For example, this will pull the latest mava tensorflow core image and run the `examples/debugging/simple_spread/feedforward/decentralised/run_maddpg.py` example:
+```
+docker run --gpus all -it --rm  -v $(pwd):/home/app/mava -w /home/app/mava instadeepct/mava:tf-core-latest python examples/debugging/simple_spread/feedforward/decentralised/run_maddpg.py --base_dir /home/app/mava/logs/
+```
+- For windows, replace `$(pwd)` with `$(curdir)`.
```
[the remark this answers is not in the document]
KaleabTessera: So I have tested the linux version, but not the windows version (as I don't have a windows device). It technically should works since we do something similiar in the makefile (https://github.com/instadeepai/Mava/blob/8c4cee233ac9a8186f356a3c2f01a17f95253a08/Makefile#L11).

PR #345, opened 2022-01-10
Feature/Mava launcher

DriesSmit:
## What?
Add a Mava launcher that can be used to construct and launch distributed (and in the future single process) runs.
## Why?
This launcher simplifies the Launchpad code setup and also paves the way for a future single process implementation of the launcher.
## How?
Added a Mava launcher class [here](https://github.com/instadeepai/Mava/blob/feature/mava2_playground/mava/utils/experimental/launcher.py). This launcher is only used by MADDPG/MAD4PG for now.
## Extra
.


DriesSmit:
Only going to implement this in the redesigned version of Mava.

PR #348, opened 2022-01-11
Mava Redesign/Update Launcher

DriesSmit:
## What?
Create a setup component to handle general building setup code.
## Why?
This is to move some of the general setup code that does not have hooks, in the Launcher, to a component that has hooks.
## How?
Added a setup component [here](https://github.com/instadeepai/Mava/blob/feature/update_launcher/mava/components/building/setup.py) that handles general setup instructions.
## Extra
.

### DIFF
--- a/mava/systems/tf/maddpg_v2/building.py
+++ b/mava/systems/tf/maddpg_v2/building.py
@@ -16,16 +16,20 @@
 """MADDPG system builder implementation."""
 
 import dataclasses
-from typing import Any, Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union
 
 import reverb
 import sonnet as snt
 from acme.utils import counting, loggers
 
 from mava import specs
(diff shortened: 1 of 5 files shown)


mava/systems/tf/maddpg_v2/building.py
```diff
@@ -212,3 +225,12 @@ class MADDPGConfig:
 
 # Builder
 system_builder = Builder(config=config, components=system_components)
+
+# Create the launcher
+program = Launcher(multi_process=True, nodes_on_gpu=["trainer"], name="MADDPG")
```
arnupretorius, line 230: I liked the idea of having the program constructed outside of the builder but happy to chat about this some more. :)

arnupretorius (approved):
Thanks @DriesSmit 😄 

We should perhaps have another chat about how the launcher and builder interact but having the general setup outside the launcher makes sense to me.

PR #349, opened 2022-01-13
Minor bug-fixing to get redesign Launcher running.

DriesSmit:
## What?
Fix minor bugs in Mava init code. The Launcher can now be called and works on both GPU/CPU.
## Why?
To be able to call the Mava Launcher and use it in an external project.
## How?
Minor bug-fixes throughout the codebase.
## Extra
.

### DIFF
--- a/mava/core.py
+++ b/mava/core.py
@@ -17,16 +17,17 @@
 """Core Mava interfaces."""
 
 import abc
-from typing import Dict, Optional, Tuple, TypeVar, Union, Any, Iterator, List
+from typing import Any, Dict, Iterator, List, Optional, Tuple, TypeVar, Union
 
 import dm_env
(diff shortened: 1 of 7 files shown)


mava/core.py
```diff
@@ -189,4 +190,4 @@ def trainer(
 
     @abc.abstractmethod
     def distributor(self) -> Any:
```
arnupretorius, line 192: We can remove this since it is no longer used. But I will do it in another PR. :)

arnupretorius (approved):
Thanks @DriesSmit! :)

mava/core.py
DriesSmit: Sounds good :slightly_smiling_face:

PR #352, opened 2022-01-14
Feature/bugfix redesign callbacks

DriesSmit:
## What?
Fix bugs in callbacks code. The first non-Mava implementation of a PPO implementation is now running using the Mava redesign callbacks method!
## Why?
Small bug-fixing and enhancements to get callbacks code running.
## How?
Small changes throughout the code.
## Extra
@arnupretorius let me know what you think about this setup with the distributor component. I sent you the system.py file and a snippet of the main code.


arnupretorius (approved):
Thanks @DriesSmit! I think this looks good 🔥 

There might be some minor things I will change on my side, but will keep you in the loop.

PR #365, opened 2022-01-31
Small bugfixes. Simplify redesign system setup code.

DriesSmit:
## What?
Implement small bugfixes. Simplify the system setup code to make it slightly more general.
## Why?
The distributor component can now also be updated or replaced before launching a system.
## How?
Removed the build function from the base system code and fixed typing errors.
## Extra
.


arnupretorius:
Thanks @DriesSmit 👍  I'm closing this since we will be changing workflows and will reconsider our component building under this setting.

PR #370, opened 2022-02-01
Fix/broken launchpad link

sash-a:
## What?
There was a broken launchpad link in the readme, it was trying to point to a launchpad folder

## Why?
Broken links are irritating...?

## How?
Seems like the reference should have used square brackets instead of round ones


KaleabTessera (approved):
Thanks @sash-a ! Well done your first PR :fire: :open_hands:

PR #372, opened 2022-02-01
Fix/mad4pg optimizers

EdanToledo:
## There is single optimizer support for MADDPG and MAD4PG
Users can now choose if they want to use a single optimizer for the observation networks, critic networks and policy networks. Currently, if two optimizers are used, they still both gradient step on the observation network. This is more of a design decision as only having one of the two gradient step on the obs network will still cause instability. 
## Why?
Having the policy and critic learn a common representation can be really effective for generalisation and certain problems.
## How?
The backward method checks to see if the user wants to use a single optimizer, if so, then it adds the critic variables to the policy and obs variables and performs a gradient step. The forward method checks to see if the user wants to use a single optimizer and if so adds the critic loss to the policy loss. 
## Extra
[Any extra information.]

### DIFF
--- a/mava/systems/tf/maddpg/training.py
+++ b/mava/systems/tf/maddpg/training.py
@@ -1399,16 +1399,27 @@ class MADDPGBaseRecurrentTrainer(mava.Trainer):
         for agent in self._agents:
             agent_key = self._agent_net_keys[agent]
 
+            if self._use_single_optimizer:
             # Get trainable variables.
-            policy_variables = (
-                self._observation_networks[agent_key].trainable_variables
-                + self._policy_networks[agent_key].trainable_variables
-            )
-            critic_variables = (
-                # In this agent, the critic loss trains the observation network.
-                self._observation_networks[agent_key].trainable_variables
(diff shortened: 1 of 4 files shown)


mava/systems/tf/maddpg/training.py
```diff
@@ -451,36 +457,43 @@ def _backward(self) -> None:
...
-                + self._critic_networks[agent_key].trainable_variables
-            )
+                policy_variables = (
+                    self._observation_networks[agent_key].trainable_variables
+                    + self._policy_networks[agent_key].trainable_variables
+                    + self._critic_networks[agent_key].trainable_variables
+                )
+            else:
+                policy_variables = (
+                    self._observation_networks[agent_key].trainable_variables
```
KaleabTessera, line 469: I think this was broken before, but like acme https://github.com/deepmind/acme/blob/b9681d9d9ad9ccaec4271bf2abeb25281185def2/acme/agents/tf/ddpg/learning.py#L215 , we should only include obs vars as part of the critic vars, not policy ones.

PR #375, opened 2022-02-02
feat: filter docker image push based on label.

KaleabTessera:
## What?
Only push docker images using github actions if PR has label `benchmark`. Closes https://github.com/instadeepai/Mava/issues/376 . 
## Why?
Pushing docker images can be a blocker.
## How?
-
## Extra
-


arnupretorius (approved):
Thanks @KaleabTessera 🙂

PR #401, opened 2022-02-09
chore: add optional benchmark questions to feature

arnupretorius:
## What?
Add benchmarking questions to feature template.
## Why?
Remove the need to separate issue. Use benchmark issue template for planning/larger-scale benchmarking.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_internal.md
+++ b/.github/ISSUE_TEMPLATE/feature_internal.md
@@ -15,12 +15,17 @@ A plan or brief overview of how the feature will be implemented.
 
 ### Testing
 A detailed description of how this feature will be tested.
-
 * What are the core aspects of the feature that need testing?
 * Outline the test cases and for each specify:
     * The purpose of the test.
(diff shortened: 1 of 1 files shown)


RuanJohn (commented):
Aside from just how and why, I think we need to also ask that any relevant parameter configurations to be used for benchmarking should be included.

AsadJeewa:
How do we expect to use this, since it is optional? We would often create simple issues without adding benchmark info so is this for "big" changes?

KaleabTessera (approved):

KaleabTessera:
> Aside from just how and why, I think we need to also ask that any relevant parameter configurations to be used for benchmarking should be included.

I guess the real question for this is whether to make the benchmarking results public. We can always link to our internal results - which contains config etc, they just won't be available to the public. 

I think in the beginning it is fine if we keep the benchmarks private (might a bit of admin to share the public results for every PR).

AsadJeewa (approved):

PR #403, opened 2022-02-10
Test/acceptance test for new mava build

### LINKED ISSUE #402: [TEST] Acceptance test for Jax-based MADDPG system
### What do you want to test?
The test is to outline a typical example of a user trying to build a system for MARL using Mava. The purpose of the test is to guide the long-term development of features and components for the new build system for Jax-based systems.

### Outline of test structure
This is an **acceptance test** and looks to test the interface, design and setup of the following:
* Factories: environment, networks and loggers
* System: components, signatures, class functions and distribution.
* Config handling for all of the above.
* Performance
* Speed
* Scalability

### Definition of done
The acceptance test covers all of the above components that need to be tested and fails. All assertions fail and must link to what is being tested to create a sensible rubric and incremental roadmap for guiding code that can make tests pass.

* [x] Test code is documented - docstrings for methods and classes, static types for arguments.

arnupretorius:
## What?
Acceptance test for running Jax maddpg system on the simple spread debug environment.
## Why?
To guide the development of new jax build system and test against TF2 maddpg.
## How?
Created example code for building and launching maddpg using Jax and the new build system.
## Extra
Note that the example is very basic and does not include all the new build system functionality we hope to support (e.g. adding components on the fly). We can increase the complexity of our acceptance tests as we make progress.

The many file changes in the PR is because I had to move all the examples into a tf folder and change the example links. Please focus on this file: `examples/jax/run_maddpg_test.py`.

closes #402


KaleabTessera (commented):
The new design looks really nice :open_hands: :fire: , just some comments. 

- I think it is better to be more explicit about the config. Rather then just accessing the properties like `config.environment_factory=...`, I think it is better to do this `config = maddpg.Config(batch_size=..., ...)`. It is more obvious what the required config is for a system and you are less likely to make mistakes when entering config (more likely to make a mistake accesing a prop like `.` than using named params).

- I also think it is better to separate config that is specific to the algorithm from things like env and log factories. 
E.g. 
  ```
  system.build( config=config, environment_factory=environment_factory, logger_factory=logger_factory, network_factory=network_factory,...)
  ```
  I think this will make the design more flexible and reusable. For example, with single and multiple process systems we can share the alg config, but then define the necessary factories.

arnupretorius:
> The new design looks really nice 👐 🔥 , just some comments.
> 
> * I think it is better to be more explicit about the config. Rather then just accessing the properties like `config.environment_factory=...`, I think it is better to do this `config = maddpg.Config(batch_size=..., ...)`. It is more obvious what the required config is for a system and you are less likely to make mistakes when entering config (more likely to make a mistake accesing a prop like `.` than using named params).
> * I also think it is better to separate config that is specific to the algorithm from things like env and log factories.
>   E.g.
>   ```
>   system.build( config=config, environment_factory=environment_factory, logger_factory=logger_factory, network_factory=network_factory,...)
>   ```  
>     
>   I think this will make the design more flexible and reusable. For example, with single and multiple process systems we can share the alg config, but then define the necessary factories.

Thanks @KaleabTessera 🙂  I hear what you are saying. The config handling I have in mind for the Jax systems is based on a mixture of `SimpleNamespaces` and `dataclasses` which will hopefully make them far more dynamic and do away with our systems having growing signatures as we add components. Components will have their own data classes and when you build a system using multiple components your config will be autogenerated as a Namespace of dataclasses each associated with a specific component. This config is then accessible and modifiable further downstream and can dynamically grow or shrink on the fly. I hope this makes sense. It is still a bit of a work in progress, but I'm happy for us to perhaps find a middle ground that has the best of both worlds.

arnupretorius:
I think I found a way to have the best of both worlds. Will update the test accordingly. 👍

arnupretorius:
Closing and splitting into two different PRs. One for moving examples in TF folder and the other for the acceptance test.

PR #405, opened 2022-02-10
Feature/jax abstract system class

### LINKED ISSUE #404: [FEATURE] Core abstract system class for Jax based systems
### Feature
Abstract call to describe new system design for Jax systems.

### Proposal
The design should include the following functions:
* Configure - to specific systems components
* Update - to update system components
* Add - to add system components
* Launch - to run the system

### Testing
A detailed description of how this feature will be tested.
All of the above class functions should be tested.
#### configure: 
- input is a config file - a simpleNamespace of dataclasses
- return SimpleNamespace of components
#### update:
- input is a component and str name
- return None (just side effect of updating, i.e. "overwriting" component)
#### add:
- input is a component and str name
- return None
#### launch:
- Input is number of executors, boolean for multi-processing, nodes specification on device and system name.
- return None (simple runs system)

### Definition of done
Class is done when all of the above functions have been implemented, typed and decorated.

* [x] Code is documented - docstrings for methods and classes, static types for arguments.
* [x] Code is tested - unit, integration and/or functional tests are added.
* [x] All functional tests are green.

arnupretorius:
## What?
Abstract system class for new build system for jax agents.
## Why?
Create an interface for systems with functions supporting component-wise builds.
## How?
Created a new base abstract class for systems.
## Extra
Regarding tests, please note that so far I'm following a different approach than what we had previously been following, which is to locate test files alongside feature files. Two reasons for this: (1) to separate it from our TF2 tests and (2) to experiment whether this might be a more preferred method. But happy to chat more about this. 

closes #404


AsadJeewa:
Really glad that we are finally adding a base class for systems.

AsadJeewa (commented):
Small comments. Simple class but very useful.

mava/core_jax.py
```diff
@@ -0,0 +1,80 @@
...
+    @abc.abstractmethod
+    def launch(
+        self,
+        name: str = "system",
+    ) -> None:
+        """Run the system.
+
+        Args:
+            name : name of the system
+        """
```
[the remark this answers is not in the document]
arnupretorius: Yes, for now I'm adding the `_jax.py` part to keep the Jax focused code separate from the TF code. Since the fundamental system design and build will also be different I think it is better this way. We can do a renaming at some later stage when the jax part starts maturing.

mava/core_jax_test.py
```diff
@@ -0,0 +1,107 @@
...
+    dummy_system.add(COMPONENT, "add")
+
+    # build system
+    dummy_system.build(CONFIG)
+
+    # distribute system
+    dummy_system.distribute(num_executors=1, nodes_on_gpu=["process"])
+
+    # launch system
+    dummy_system.launch()
```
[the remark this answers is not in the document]
arnupretorius: Yes, please see the extras section of the PR where I explain the reasoning behind this. But happy to discuss further.

mava/core_jax.py
```diff
@@ -0,0 +1,72 @@
...
+
+import abc
+from typing import Any, List
+
+
+class BaseSystem(abc.ABC):
+    """Abstract system object."""
+
+    @abc.abstractmethod
+    def design(self) -> List[Any]:
```
AsadJeewa, line 27: I get that we add the required components and then configure them but why then do we need a design method to select certain ones only?

AsadJeewa (approved):
Looks good. Just had one small question

mava/core_jax.py
arnupretorius: Hey @AsadJeewa, the design will list the default components of the system. For example, for MADQN it will contain the components for a "base" MADQN. The `.add()` and `.update()` is then for changing the existing components or adding new ones to the base system.

mava/core_jax_test.py
```diff
@@ -0,0 +1,88 @@
...
+        assert name == "system"
+
+
+@pytest.fixture
+def dummy_system() -> TestDummySystem:
+    """Create complete system for use in tests"""
+    return TestDummySystem()
+
+
+def test_dummy_system(dummy_system: TestDummySystem) -> None:
```
KaleabTessera, line 73: @arnupretorius I am not too sure on the utility of these tests. If we are testing the abstract base class, then we should tests things like no subclass can implement it without implement those abstract methods or test something more foundational. 

Here it looks like we are mocking out a system and then testing the mocked system, without testing any real functionality. I think other things we could test, could be for example , have `BaseSystem` class create properties or variables according to inputs and test those exist e.g. if we pass components to a subclass of `BaseSystem` , then those components should be returned when we call `design()` or actually implement `update()` or `add()`. 

My main worry here is that it doesn't seem to be testing any real code.

KaleabTessera (commented):
Thanks @arnupretorius , I left a comment on the tests.

mava/core_jax_test.py
arnupretorius: I agree, the tests aren't really doing much here. Initially, all I wanted to test was what you suggested, that no subclass can implement it without implementing the abstract methods. However, I ran into the mypy errors I shared with you a while ago, where even if I use pytest's with raises exception or a try catch to have the test pass when the class has not implemented the required functions, mypy complains. So I simply remove that test. As for the other tests you are suggesting, I think they are more appropriate for when we are actually implementing the general class that will be shared between systems, as was done in this PR: feature/jax-general-system-class

mava/core_jax.py
```diff
@@ -0,0 +1,72 @@
...
+    @abc.abstractmethod
+    def configure(self, **kwargs: Any) -> None:
+        """Configure system hyperparameters."""
+
+    @abc.abstractmethod
+    def launch(
+        self,
+        num_executors: int,
+        nodes_on_gpu: List[str],
+        multi_process: bool = True,
```
RuanJohn, line 60: Everything is clear. I just have a question about `multi_process`. Does thie toggle whether we run distributed or as a single process? Just a bit unclear why it is whether to run locally or distributed?

mava/core_jax_test.py
arnupretorius: If you can find a way around the mypy issue, that would really help. 😉  The `with pytest.raises()` works nicely in some of the other tests, but for some reason doesn't want to play nice with mypy regarding this specific test of inheritance from an abstract class.

mava/core_jax.py
arnupretorius: Good point, probably better to have the wording be single vs multi-process 👍

mava/core_jax_test.py
KaleabTessera: Will look into it :)

arnupretorius: Hey @KaleabTessera, I think I found a workaround. Please see the updated tests. :)

mava/core_jax_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+import pytest
+
+from mava.core_jax import BaseSystem
+
+# Dummy component
+COMPONENT = None
+
+
+def test_exception_for_incomplete_child_class() -> None:
```
AsadJeewa, line 30: If the class is abstract this is expected behavior so what is the purpose of this test? To make sure that over time, we don't remove the abstract annotations?

AsadJeewa (approved):
Approved but have a question on the test

mava/core_jax_test.py
KaleabTessera: Thanks @arnupretorius , I really like the `test_exception_for_incomplete_child_class` test :open_hands: 

Personally, I still recommend not adding `TestDummySystem` and `test_dummy_system` until we are testing some implemented functionality. I think this PR is good with the abstract class + `test_exception_for_incomplete_child_class`. In future, we can implement tests for things like component adds etc.

arnupretorius: Cool, I removed the dummy system test. 👍

mava/core_jax_test.py
arnupretorius: Yes, I think that is the purpose of most tests, to ensure the behaviour of code is maintained over time. Of course this is quite small and it is unlikely that anyone remove some of the decorators, but writing the tests are cheap so I think it is still worth having this extra layer of checks.

RuanJohn (approved):
Thank you for all your effort @arnupretorius! :muscle: :fire:

KaleabTessera (approved):
Thanks @arnupretorius :+1:

Issue #412, opened 2022-02-16
[BUG] Fix Smac Wrapper Relying on Flatland


KaleabTessera:
### Describe the bug
If your image/virtual env doesn't have flatland installed, it will fail to import the smac wrapper. 

### To Reproduce
1. Build SC2 docker image.
2. Try to import Smac Wrapper.

### Expected behavior
You are able to import the smac wrapper and run exps. 

### Context (Environment)
 - Mava (and other package versions): result of `pip freeze > requirements.txt`
 - OS

### Proposal for a fix
Have a separate try catch for flatland and smac. 

### Definition of done
Smac experiments works. 

### Mandatory checklist before making a PR
* [ ] Can import smac wrapper without installing flatland. 
* [ ] Smac experiments run.

PR #413, opened 2022-02-16
Fix/Smac Wrapper Relies on Flatland Installation

KaleabTessera:
## What?
Fixes the bug where the smac wrapper won't be imported if flatland is not installed. 
## Why?
- Run smac experiments without flatland being installed. 
## How?
- Two separate try catch statements. 
## Extra
Closes https://github.com/instadeepai/Mava/issues/412 .


RuanJohn (approved):
Thank you @KaleabTessera :call_me_hand:

mmorris44 (approved):

PR #414, opened 2022-02-17
Feature/jax mava custom config class

### LINKED ISSUE #411: [FEATURE] custom Mava config class
### Feature
A custom class to handle config for Mava systems.

### Proposal
The class should be easy to use for setting system configs and be able to grow/shrink dynamically based on the component configuration of the system. The proposal is to achieve this using `SimpleNamespaces` containing a `dataclass` for each component. The class will also support a property method to easily retrieve the config for the system and a method for specific component configs by name. 

### Testing
All methods should be tested including specific get 
As input, the config will expect a dataclass with side effects of updating the current config.
The "get" method for component-specific config should take in the component name and return that component's config.

### Definition of done
The class is implemented and supports the above functionality.

arnupretorius:
## What?
Config handler class for Jax based Mava systems.
## Why?
More flexible configs that dynamically change with the component setup of a system.
## How?
Implement a config class with methods to handle adding and updating config dataclass for different system components and updating config hyperparameters.

closes #411


mava/systems/jax/config.py
```diff
@@ -0,0 +1,139 @@
...
+from mava.utils.config_utils import flatten_dict
+
+
+class Config:
+    """Config handler for Jax-based Mava systems."""
+
+    def __init__(self) -> None:
+        """Initialise config"""
+        self._config: Dict = {}
+        self._current_params: List = []
```
KaleabTessera, line 31: Do we need `self._current_params`, if we can just get the keys from `self._config.keys()`? Are they serving a different purpose? 

If they are the same, we can do `self._current_params = self._config.keys()`  and then we don't have to update `self._current_params`.

mava/systems/jax/config.py
```diff
@@ -0,0 +1,139 @@
...
+                    raise Exception(
+                        "The given component config is not part of the current \
+                        system. Perhaps try adding the component using .add() \
+                        in the system builder."
+                    )
+            else:
+                raise Exception("Component configs must be a dataclass.")
+
+    def build(self) -> None:
+        """Build the config file, i.e. unwrap dataclass nested dictionaries"""
```
KaleabTessera, line 95: Might be worth adding `if not self._built` check.

mava/systems/jax/config.py
```diff
@@ -0,0 +1,139 @@
...
+        config_unwrapped: Dict = {}
+        for param in self._config.values():
+            config_unwrapped.update(flatten_dict(param.__dict__))
+        self._built_config = config_unwrapped
+        self._built = True
+
+    def set(self, **kwargs: Any) -> None:
```
KaleabTessera, line 102: Can we possibly make this function name a little more descriptive, something like `set_build_config`. This could be confused with the update methods.

mava/systems/jax/config_test.py
```diff
@@ -0,0 +1,252 @@
...
+        dummy_hyperparameter_config : component config dataclass of hyperparameters
+    """
+    config.add(component=dummy_component_config)
+    config.update(component=dummy_hyperparameter_config)
+    config.build()
+    conf = config.get()
+
+    assert conf.param_0 == 2.7
+    assert conf.param_1 == 3.8
+
```
KaleabTessera, line 148: Possibly add `assert not hasattr(config,"setting")` since we shouldn't have this prop anymore.

KaleabTessera (approved):
This is excellent @arnupretorius , super excited to use this. 

I just made some minor comments. My biggest comment is that it should be obvious what we are getting/setting when we call `add`,`update`,`get`,`set`. This are all called like `config.` , but some apply to `self._config` and others to `self._built_config`. This might be a bit confusing when using it.

mava/systems/jax/config.py
arnupretorius: They serve slightly different purposes. The `config` is a dictionary of dataclasses where each dataclass contains a parameter keys and values, whereas the `current_params` maintains a dict of the unwrapped config, i.e. all leaf parameter keys and values from the config. The reason we need to update the `current_params` is for when new components are added/updated we need to check if all leaf parameters still have unique names, i.e. we haven't by accident overwritten a leaf parameter by adding a new dataclass that shares a parameter name with another dataclass. I hope this makes?

mava/systems/jax/config.py
arnupretorius: Good suggestion, will do.

mava/systems/jax/config.py
arnupretorius: I will have a bit of a think around updating the names to try to make them more descriptive.

mava/systems/jax/config_test.py
arnupretorius: Good idea. Will do

mava/utils/config_utils.py
```diff
@@ -0,0 +1,42 @@
...
+        parent_key : optional parent key identifier
+        sep : optional separator between parent and leaf keys
+    Returns:
+        Flattened dictionary
+    """
+    items: List = []
+    for k, v in d.items():
+        new_key = parent_key + sep + k if parent_key else k
+        if isinstance(v, MutableMapping):
+            pass
```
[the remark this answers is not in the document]
arnupretorius: Nope, no idea why it is there. Will remove it.

mava/systems/jax/config.py
arnupretorius: Done

arnupretorius: Also added some additional checks

mava/systems/jax/config_test.py
arnupretorius: Done

mava/systems/jax/config.py
```diff
@@ -0,0 +1,153 @@
...
+                            in the system builder."
+                        )
+                else:
+                    raise Exception("Component configs must be a dataclass.")
+        else:
+            raise Exception(
+                "Component configs cannot be updated if config has already been built."
+            )
+
+    def build(self) -> None:
```
DriesSmit, line 104: Why do we have a build and a launch? Can't we just build it when we launch it? Then we don't have to check if it has already been built?

mava/systems/jax/config.py
DriesSmit: Can't we just redo calculating the list of parameters without needing `_current_params`? It might simplify the code with no real extra computational overhead?

DriesSmit (commented):
Looks good, thanks @arnupretorius :fire: Just see my comments.

mava/systems/jax/config.py
arnupretorius: There is no launch in the config? The building here separates the process of configs being added and updated when the system components are being added and updated and the process of setting the system hyperparameters. Once all the components have been added updated, the system can be configured, this then has the config built internally by the system which allows the user to set the parameters of the system before launching. Perhaps it might help to look at the acceptance test (#418) and general system class (#425).

arnupretorius: There is no launch in the config? The building here separates the process of configs being added and updated when the system components are being added and updated and the process of setting the system hyperparameters. Once all the components have been added updated, the system can be configured, this then has the config built internally by the system which allows the user to set the parameters of the system before launching. Perhaps it might help to look at the acceptance test (#418) and general system class (#425).

mava/systems/jax/config.py
arnupretorius: I don't think this would be more efficient since it would require flattening the config to obtain the leaf parameters every time a component is added/updated. I think keeping a temp list is easiest and more efficient, but there might be better ways still. Will think a bit about it.

mava/utils/config_utils.py
```diff
@@ -0,0 +1,41 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utility functions for Mava config"""
+
+from collections.abc import MutableMapping
+from typing import List
+
+
+def flatten_dict(
```
RuanJohn, line 22: Is there a test for this? If not, I am happy to write one :)

arnupretorius: No test, please feel free to write one @RuanJohn! 🙌

arnupretorius: No test, if you are keen, please write one! 🙌

RuanJohn: On it :)

RuanJohn: Sorted :call_me_hand:

arnupretorius: Thanks 👍

tests/utils/config_utils_test.py
```diff
@@ -0,0 +1,50 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from mava.utils.config_utils import flatten_dict
+
+test_mock_nested_dict = {"a": 1, "c": {"a": 2, "b": {"x": 5, "y": 10}}, "d": [1, 2, 3]}
```
arnupretorius, line 18: @RuanJohn, I think we can make this a pytest fixture, then remove the test class (just have the test functions) and feed the fixture as input to each test function. This isn't necessarily better, but it will be more consistent with the other testing code. :)

mava/systems/jax/config.py
arnupretorius: I changed the name to `set_parameters`. Let me know if you think this is distinctive enough from `update`.

mava/utils/config_utils.py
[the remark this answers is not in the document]
arnupretorius: Done.

mava/systems/jax/config.py
KaleabTessera: Sounds good :)

RuanJohn (approved):
Thank you @arnupretorius!

mava/systems/jax/config.py
DriesSmit: Sounds good :+1:

Issue #420, opened 2022-02-21
[BUG] New Gym version breaks tests


KaleabTessera:
### Describe the bug
New gym version (`0.22.0`) breaks our tests. This is related to the output of `wrapped_env.action_spaces[agent].sample()` for `pettingzoo.mpe.simple_spread_v2`. Old version returns `{'agent_0': 3, 'agent_1': 3, 'agent_2': 3} ` , new version returns `{'agent_0': 0, 'agent_1': 0, 'agent_2': 0}` resulting observations not being updated and thus the test fails.

### To Reproduce
1. Upgrade to gym `0.22.0`
2. Run tests. 

### Expected behavior
Observation to be updated. 

### Context (Environment)
-

### Proposal for a fix
Change random seed. This seed causing no action in this version of gym, meaning no update to obs and so test fails. 

### Definition of done
Tests pass. 

### Mandatory checklist before making a PR
* [ ] The success criteria laid down in “Definition of done” are met.
* [ ] Code is documented - docstrings for methods and classes, static types for arguments.
* [ ] Code is tested - unit, integration and/or functional tests are added.
* [ ] Documentation is updated - README, CONTRIBUTING, or other documentation.
* [ ] All functional tests are green.
* [ ] Link experiment/benchmarking after implementation (optional).

PR #421, opened 2022-02-21
fix: Fix broken tests due to new gym version. 

KaleabTessera:
## What?
Fixes broken tests due to new gym version. 

- New gym version (0.22.0) breaks our tests. This is related to the output of wrapped_env.action_spaces[agent].sample() for pettingzoo.mpe.simple_spread_v2. Old version returns `{'agent_0': 3, 'agent_1': 3, 'agent_2': 3} `, new version returns `{'agent_0': 0, 'agent_1': 0, 'agent_2': 0}` - i.e. `no_action` resulting observations not being updated and thus the test fails.

## Why?
Fix tests. 
## How?
Change seed and specify gym version so this doesn't happen in the future. 
## Extra
Closes https://github.com/instadeepai/Mava/issues/420

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,7 @@ setup(
         "matplotlib",
         "dataclasses",
         "box2d-py",
-        "gym",
+        "gym==0.21.0",
     ],
     extras_require={
         "tf": tf_requirements,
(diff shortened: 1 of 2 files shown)


arnupretorius (approved):
Thanks @KaleabTessera 🙌

RuanJohn (approved):
Thank you @KaleabTessera :fire:

Issue #427, opened 2022-02-23
[BUG] Shared weights across agent types


AsadJeewa:
### Describe the bug
When setting shared_weights to true, a single network is created for all agents. It should not be possible to share weights across agent types

### To Reproduce
1. Run examples for MADQN, MADDPG or MAPPO
2. Log the number of networks/ agent network keys

### Expected behavior
Each agent type should share weights 

### Context (Environment)
 - No specific dependencies or OS, Latest MAVA repo issue

### Proposal for a fix
Tweak the network ID allocation to assign new ids for different agent types

### Definition of done
Examples still run with new agents network IDS

### Mandatory checklist before making a PR
* [ ] The success criteria laid down in “Definition of done” are met.
* [ ] Code is documented - docstrings for methods and classes, static types for arguments.
* [ ] Link experiment/benchmarking after implementation (optional).

PR #428, opened 2022-02-23
fix: shared weights with agent type

AsadJeewa:
## What?
Updated agent network types to only share weights between agent types
https://github.com/instadeepai/Mava/issues/427
## Why?
When setting shared_weights to true, a single network is created for all agents. It should not be possible to share weights across agent types
## How?
changed self._agent_net_keys in systems.py to assign the correct keys (for MADQN, MAPPO, MADDPG)
## Extra
If we are happy with this change, would we then need to benchmark?


mava/systems/tf/maddpg/system.py
```diff
@@ -227,10 +228,12 @@ def __init__(  # noqa
         if type(network_sampling_setup) is not list:
             if network_sampling_setup == enums.NetworkSampler.fixed_agent_networks:
                 # if no network_sampling_setup is fixed, use shared_weights to
-                # determine setup
+                # assign a single network to all agents of the same type
                 self._agent_net_keys = {
-                    agent: "network_0" if shared_weights else f"network_{i}"
-                    for i, agent in enumerate(agents)
+                    agent: f"""network_{agent.split("_")[0]}"""
```
arnupretorius, line 233: Why use triple quotes here? Will `f"network_{agent.split("_")[0]}"` not do?

mava/systems/tf/mappo/system.py
```diff
@@ -190,12 +190,15 @@ def __init__(
         self._network_factory = network_factory
         self._logger_factory = logger_factory
         self._environment_spec = environment_spec
-        # Setup agent networks
+        # Setup agent networks assign a single network to all agents of the same type
```
arnupretorius, line 193: This doesn't read too nicely. Perhaps a line break between "networks" and "assign", or think about rephrasing.

arnupretorius (commented):
Thanks @AsadJeewa! 😄 

Just see my few comments. Also, did we test this? At the very least we should do some runs on environments with and without different agent types.

mava/systems/tf/maddpg/system.py
AsadJeewa: Triple quotes are needed due to the use of double quotes in "_". It is possible to use single quotes but the code formatter automatically converts ' to "
https://stackoverflow.com/questions/68461626/how-to-fix-unterminated-expression-in-f-string-missing-close-brace-in-python

AsadJeewa: Triple quotes are needed due to the use of double quotes in "_". It is possible to use single quotes but the code formatter automatically converts ' to "
https://stackoverflow.com/questions/68461626/how-to-fix-unterminated-expression-in-f-string-missing-close-brace-in-python

AsadJeewa:
I tested that everything works in multiple different environments with different (openspiel Tic Tac Toe, Petting Zoo Pong) or single-agent types (debugging, Petting Zoo Multiwalker, Flatland) to make sure that the networks were being assigned correctly and that execution proceeds. Since shared_weights default value is True, I think that we should run a small set of benchmarks @RuanJohn

examples/tf/petting_zoo/atari/pong/recurrent/decentralised/run_madqn.py
```diff
@@ -87,7 +87,6 @@ def main(_: Any) -> None:
         exploration_scheduler_fn=LinearExplorationScheduler(
             epsilon_start=1.0, epsilon_min=0.05, epsilon_decay=1e-6
         ),
-        shared_weights=False,
```
mmorris44, line 90: If we want to use shared weights here, can we rather just make this explicitly `True` rather than not setting the argument? Seems clearer to me.

mava/systems/tf/maddpg/system.py
```diff
@@ -226,11 +227,14 @@ def __init__(  # noqa
...
+                # if no network_sampling_setup is specified, assign a single network
+                # to all agents of the same type if weights are shared
+                # else assign seperate networks to each agent
                 self._agent_net_keys = {
-                    agent: "network_0" if shared_weights else f"network_{i}"
-                    for i, agent in enumerate(agents)
+                    agent: f"network_{agent.split('_')[0]}"
+                    if shared_weights
+                    else f"network_{agent}"
+                    for agent in agents
```
mmorris44, line 237: This seems to actually change network keys from what they were before, no?

Previously, they would be `network_{i}` for each agent index `i`. But now they're `network_{agent}` for each `agent`.

Unsure if this is intentional, and if it is, will it not cause problems in other places where we assume a specific agent net key name convention?

mava/systems/tf/madqn/system.py
```diff
@@ -214,11 +214,14 @@ def __init__(  # noqa
...
+                # if no network_sampling_setup is specified, assign a single network
+                # to all agents of the same type if weights are shared
+                # else assign seperate networks to each agent
                 self._agent_net_keys = {
-                    agent: "network_0" if shared_weights else f"network_{i}"
-                    for i, agent in enumerate(agents)
+                    agent: f"network_{agent.split('_')[0]}"
+                    if shared_weights
+                    else f"network_{agent}"
+                    for agent in agents
```
mmorris44, line 224: Same potential issue as above

mava/systems/tf/madqn/system.py
```diff
@@ -255,7 +258,7 @@ def __init__(  # noqa
                 agents,
                 self._network_sampling_setup,  # type: ignore
             )
-
+        print(self._agent_net_keys)
```
mmorris44, line 261: I assume this is just for debug?

Needs to be removed before we merge.

mava/systems/tf/mappo/system.py
```diff
@@ -190,12 +190,15 @@ def __init__(
...
+        # if weights are shared else assign separate networks to each agent
         self._agent_net_keys = agent_net_keys
         if not agent_net_keys:
             agents = environment_spec.get_agent_ids()
             self._agent_net_keys = {
-                agent: agent.split("_")[0] if shared_weights else agent
+                agent: f"network_{agent.split('_')[0]}"
+                if shared_weights
+                else f"network_{agent}"
                 for agent in agents
```
mmorris44, line 202: And another same comment as above

mmorris44 (changes requested):
- 1 potential issue in 3 places
- 1 debug print statement
- 1 minor comment

mava/systems/tf/madqn/system.py
AsadJeewa: ah yes I have removed it

examples/tf/petting_zoo/atari/pong/recurrent/decentralised/run_madqn.py
AsadJeewa: I removed this statement as it does not make sense to share weights in multi-agent of Pong since there are two agent types in a competitive game (It is possible to make it a single-agent problem). The new functionality makes sure that each agent has there own network even when shared weights is True since they are different types But I guess you are right in that it is better to be explicit.

mava/systems/tf/maddpg/system.py
arnupretorius: Ah I see, thanks 👍

mava/systems/tf/maddpg/system.py
AsadJeewa: Yes, it does change the naming convention but this was not standardised in the past i.e. PPO was different from MADQN & MADDPG. I have standardised it now. There is only a naming convention assumed for agents_ids and not network_ids (from my investigation into the repo)

examples/tf/petting_zoo/atari/pong/recurrent/decentralised/run_madqn.py
AsadJeewa: @arnupretorius @DriesSmit Would it not make more sense to have shared_weights set to false by default?

AsadJeewa:
This line (https://github.com/instadeepai/Mava/blob/develop/mava/specs.py#L68) assumes that all environments will follow the naming convention of type_identifier but this is not always the case e.g. OpenSpiel TicTacToe: agents are named player_0 and player_1 so Mava assumes they are the same type

examples/tf/petting_zoo/atari/pong/recurrent/decentralised/run_madqn.py
arnupretorius: Sharing weights is more common in MARL than having separate parameters for each agent, or at least in the cooperative, mixed (team) settings. I think it makes sense to have `shared_weights` default to `True`.

mava/systems/tf/mappo/system.py
AsadJeewa: addressed in comment above

mava/systems/tf/madqn/system.py
AsadJeewa: addressed in comment above

mava/systems/tf/mappo/system.py
AsadJeewa: Updated the comments to be more explicit

mmorris44 (approved):
PR looks good to go from my side.
Thanks @AsadJeewa!

mava/systems/tf/mappo/system.py
arnupretorius: Thanks 🙂

arnupretorius (approved):
Thanks @AsadJeewa! 👍

PR #425, opened 2022-02-23
Feature/jax general system class

arnupretorius:
## What?
General system class for Jax-based Mava systems.
## Why?
The new system class supports a more general and flexible build system using callback components.
## How?
Create a system class with functions for specifying a system design using components and adding, updating and configuring system component hyperparameters.
## Extra
I'm experiencing a tricky mypy issue that I haven't been able to solve. It has to do with the `Builder` attributes not being picked up even though the tests seem to indicate that the system and builder interact as expected. This is also the case with the `CallbackHookMixin`. @KaleabTessera perhaps you might have some ideas once you have looked at the code? 

closes #406

PR #433, opened 2022-02-24
Feature/jax abstract builder class

### LINKED ISSUE #429: [FEATURE] Abstract builder class for Jax-based systems
### Feature
Abstract class for system building for Jax-based Mava systems.

### Proposal
The builder should take care of building essentially elements of a MARL system that run in different processes. These include the data server, variable server, executor (and evaluator) and trainer. 

### Testing
Tests will only consider the proper inheritance of the abstract builder class.

### Definition of done
All abstract methods are defined and have input and return types specified. 

* [x] Code is documented - docstrings for methods and classes, static types for arguments.
* [x] Code is tested - unit, integration and/or functional tests are added.
* [x] All functional tests are green.

arnupretorius:
## What?
Abstract builder class for jax-based Mava systems.
## Why?
To inform the builder for Jax systems through inheritance. 
## How?
Create class with the required functions, including building a data and parameter server, executor and trainer process.
## Extra
@DriesSmit I decided to remove the evaluator as a separate build function since I think we can just use the executor and specify that it is an evaluator process using `executor_id="evaluator"`. Let me know what you think 🙂 

closes #429


mava/core_jax.py
```diff
@@ -0,0 +1,121 @@
...
+import abc
+from types import SimpleNamespace
+from typing import Any, Dict, List
+
+
+class BaseSystem(abc.ABC):
+    """Abstract system object."""
+
+    @abc.abstractmethod
+    def design(self) -> SimpleNamespace:
```
DriesSmit, line 28: There is probably a reason for this, but I am just wondering why this design cannot just happen in init?

DriesSmit (commented):
I really like this setup! Really cool work, thanks @arnupretorius :fire: :fire: Just see my comment.

DriesSmit (approved):

mava/core_jax.py
arnupretorius: The main reason is that there will be things happening in the init that should be in every system for which the user should not have to care about. Therefore, separating out the components specification using the design function made more sense. Plus, this way we can more explicitly make it an abstract function to ensure it must be overwritten. This all might make more sense if you look at the general system class here: #425

DriesSmit (approved):

AsadJeewa (approved):
Looks good. Difficult to review PRs with abstract classes but all seems in order and keen to see concrete implementations. Thanks

mava/core_jax.py
```diff
@@ -69,3 +69,59 @@ def launch(
...
+    def parameter_server(self) -> Any:
+        """Parameter server to store and serve system network parameters.
+
+        Returns:
+            System parameter server
+        """
+
+    @abc.abstractmethod
+    def executor(
+        self, executor_id: str, data_server_client: Any, parameter_server_client: Any
```
sash-a, line 95: Is it not possible to have some sort of abstract data server client and parameter server client just so these methods know what methods these objects will have? 

(Very possible that this can't be done, just curious)

arnupretorius: Thanks Sasha, this is a good point. I think we should definitely change the types to correspond to the data and parameter client classes once they are instantiated. I will have a think about abstract versions for these as well. But I think we can do that in a separate PR.

mava/core_jax.py
```diff
@@ -69,3 +69,59 @@ def launch(
...
+    @abc.abstractmethod
+    def data_server(self) -> List[Any]:
+        """Data server to store and serve transition data from and to system.
+
+        Returns:
+            System data server
+        """
+
+    @abc.abstractmethod
+    def parameter_server(self) -> Any:
```
KaleabTessera, line 86: Are these returning properties? E.g. `return self._parameter_server` ? If so we should add `@property` to the definition.

mava/core_jax.py
```diff
@@ -69,3 +69,59 @@ def launch(
                 are primarily for debugging
             name : name of the system
         """
+
+
+class SystemBuilder(abc.ABC):
+    """Abstract system builder."""
+
+    @abc.abstractmethod
+    def data_server(self) -> List[Any]:
```
KaleabTessera, line 78: Same comment as below.

mava/core_jax.py
KaleabTessera: Would we lose any flexibility if we made `data_server_client` and `parameter_server_client` properties and then just called this like ` executor(executor_id=x)`? 

Also are we calling this function for each executor ? If so, why not call this for all executors and just return a dict with executor id and the corresponding executors?

mava/core_jax.py
```diff
@@ -69,3 +69,59 @@ def launch(
...
+            executor_id : id to identify the executor process for logging purposes
+            data_server_client : data server client for pushing transition data
+            parameter_server_client : parameter server client for pulling parameters
+        Returns:
+            System executor
+        """
+
+    @abc.abstractmethod
+    def trainer(
+        self, trainer_id: str, data_server_client: Any, parameter_server_client: Any
```
KaleabTessera, line 109: Same comments as the `executor(` function.

KaleabTessera (commented):
Thanks @arnupretorius , just some comments/clarifications!

mava/core_jax.py
KaleabTessera: I just saw the comment on `executor_id="evaluator"` . Maybe a bool flag is easier if we using this to differentiate between eval and exec, otherwise someone might enter  `executor_id="executor_0"` thinking this refers to the specific executor.

Issue #407, opened 2022-02-11
[BUG] Docs build failing - Update docs for new vdn and qmix implementation + remove dial


KaleabTessera:
### Describe the bug
- Our docs try to point to qmix and vdn classes that longer exist (changed to value decomposition).
- Remove dial. 

### To Reproduce
1. Go to `docs` and `pip install -r requirements.txt`
2. Run `mkdocs serve`

### Expected behavior
Docs build to pass and new pages to show. 

### Context (Environment)
 - Mava (and other package versions): result of `pip freeze > requirements.txt`
 - OS

### Additional context
-

### Possible Solution
Updated docs pages to point to correct classes + remove dial from docs.

PR #435, opened 2022-02-24
Fix/ Fix docs build.

KaleabTessera:
## What?
Fix docs build. 
## Why?
So that docs build work and docs get updated. 
## How?
Updated files that docs point to. 
## Extra
Closes https://github.com/instadeepai/Mava/issues/407 .


mava/utils/environments/flatland_utils.py
```diff
@@ -22,7 +22,27 @@
 )
 from mava.wrappers.flatland import FlatlandEnvWrapper
 
-try:
+
+def check_flatland_import() -> bool:
```
arnupretorius, line 26: Is this something we only need for Flatland?

arnupretorius (approved):
Thanks @KaleabTessera 🙂

mava/utils/environments/flatland_utils.py
KaleabTessera: So we can create these functions for other imports, flatland is just the one that causes the most issues (mainly because it is so strict on package versions).

arnupretorius: Ok I see, thanks 👍

DriesSmit (approved):
Looks good, @KaleabTessera :slightly_smiling_face:

Issue #436, opened 2022-02-24
[BUG] Flatland docker container doesn't build


KaleabTessera:
### Describe the bug
Flatland's docker container doesn't build due a `gym` version conflict.

Error:
```
The conflict is caused by:
	id-mava[flatland] 0.1.0 depends on gym~=0.22.0
	flatland-rl 3.0.1 depends on gym==0.14.0
 
To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```
### To Reproduce
1. Run `make build version=flatland`

### Expected behavior
Flatland to install correctly.

### Context (Environment)
 - Mava (and other package versions): result of `pip freeze > requirements.txt`
 - OS

### Additional context
-

### Possible Solution
- Use flatland's gym version (for now) and then in future upgrade flatland.

PR #437, opened 2022-02-24
Fix/Flatland Docker Container

KaleabTessera:
## What?
Fix version conflict between mava's gym version and flatland's gym version.
## Why?
To be able to build the flatland container.
## How?
Use flatland's gym version (for now). 
## Extra
Closes https://github.com/instadeepai/Mava/issues/436.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,6 +70,9 @@ ENV SC2PATH /home/app/mava/3rdparty/StarCraftII
 ##########################################################
 # Flatland Image
 FROM tf-core AS flatland
+# Currently mava relies on gym~=0.22.0, while flatland relies gym==0.14.0.
+# This uses flatland's requirement.
+RUN pip uninstall gym -y
 RUN pip install -e .[flatland]
 ##########################################################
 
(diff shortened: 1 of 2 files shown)


mmorris44 (approved):
Still feels to me like we need a better long-term solution to changing package requirements, but this is a nice easy fix for now.

RuanJohn (approved):
Thanks @KaleabTessera! :steam_locomotive:

KaleabTessera:
> Still feels to me like we need a better long-term solution to changing package requirements, but this is a nice easy fix for now.

@mmorris44 Just to clarify, this only impacts mava's flatland container, all other containers use the newer version of gym. Unfortunately even the latest version of flatland requires gym==0.14.0 (https://gitlab.aicrowd.com/flatland/flatland/-/blob/master/requirements_dev.txt) :face_with_head_bandage:

mmorris44 (approved):

KaleabTessera:
I removed the gym version requirements from mava's setup. Pip will resolve to the version when `pip install... ` is run. Usually to the latest, or gym ==0.14.0 for flatland.

arnupretorius (approved):
Thanks @KaleabTessera 👍

PR #441, opened 2022-02-28
Fix/add loss mask to ppo

EdanToledo:
## What?
Loss masking was not correct in PPO.
## Why?
Loss masking needs to consider last timestep.
## How?
Adds a one to the beginning of the dm timestep discounts array and removes the last value.


DriesSmit (approved):
Looks good, thanks @EdanToledo :raised_hands:

DriesSmit (approved):

arnupretorius (approved):
Thanks @EdanToledo, looks good 🙂

Issue #447, opened 2022-03-02
[MAINTAIN] Docs for value decomposition mixer are unclear


sash-a:
### Please describe what needs to be maintained?
Docs and typing don't let you know that you can use the option "qmix" or "vdn" for the mixer argument of value decomposition. 

### Describe the outcome you'd like
Let the user know know "qmix" or "vdn" is an available option [here](https://github.com/instadeepai/Mava/blob/develop/mava/systems/tf/value_decomposition/system.py#L100)

Change typing [here](https://github.com/instadeepai/Mava/blob/develop/mava/systems/tf/value_decomposition/system.py#L50) to `Union[snt.Module, str]`

### How do we know when maintenance is complete?
Checklist:
- [ ] Docs updated
- [ ] Param type updated

PR #448, opened 2022-03-02
docs: updated docs to better represent available options

sash-a:
## What?
issue #447 

Updated docs to better represent available options for mixer networks in value decomposition
 
## Why?
It was not clear that I could use the strings "qmix" or "vdn" when looking at the docs. (Thanks @jcformanek for pointing this out)

## How?
Docs are updated to indicate this is possible and argument type is changed to `Union[snt.Module, str]`


AsadJeewa (approved):
Thanks Sasha

Issue #443, opened 2022-02-28
[FEATURE] Callbacks Mixin for system building


arnupretorius:
### Feature
Mixin class for running callback functions from builder components.

### Tests
Create mock components and test function hooks are called as expected. 

### Definition of done
Call functions for all builder component hooks.

### Mandatory checklist before making a PR
* [ ] The success criteria laid down in “Definition of done” are met.
* [ ] Code is documented - docstrings for methods and classes, static types for arguments.
* [ ] Code is tested - unit, integration and/or functional tests are added.
* [ ] Documentation is updated - README, CONTRIBUTING, or other documentation.
* [ ] All functional tests are green.

Issue #440, opened 2022-02-28
[FEATURE] Callbacks class with builder hooks


arnupretorius:
### Feature
Callback class for Jax-based Mava systems to support hooks within the building, execution and training of systems. This class will be the base for all system components. 

### Proposal
Callbacks class with builder hooks, mainly to support data and parameter server components (executor and trainer hooks will be in a different PR.) 

### Testing
Test each class function. Setup a mock component structure to call different hooks and test their expected behaviour.

### Definition of done
Class is implemented with all required functions to support the creation of data and parameter server components that can be fed to the builder.

### Mandatory checklist before making a PR
* [ ] The success criteria laid down in “Definition of done” are met.
* [ ] Code is documented - docstrings for methods and classes, static types for arguments.
* [ ] Code is tested - unit, integration and/or functional tests are added.
* [ ] Documentation is updated - README, CONTRIBUTING, or other documentation.
* [ ] All functional tests are green.

PR #452, opened 2022-03-09
Feature/jax builder class

### LINKED ISSUE #426: [FEATURE] General system builder for Jax-based systems
### Feature
A builder class for Jax-based Mava systems.

### Proposal
The builder will take as input a list of system components, which are themselves callbacks and build the system using a set of functions consisting of hooks. Functions will include building tables, datasets, adders, variable servers and clients, executor, trainers, evaluators and launching. 

### Testing
All functions should be tested with proper handling of hooks. It is important to handle single instance hooks as well as make sure that calls run when expected and only when expected using the Mixin.  

### Definition of done
A precise outline for the feature to be considered fully implemented, i.e. done/complete.

arnupretorius:
## What?
General builder class for Jax systems.
## Why?
For building systems in a highly flexible way using callback system components.
## How?
Create a builder class with general-purpose builder methods and hooks.

closes #426 #443 #440


mava/callbacks/hook_mixin.py
```diff
@@ -0,0 +1,195 @@
...
+class CallbackHookMixin(ABC):
+
+    ######################
+    # system builder hooks
+    ######################
+
+    callbacks: List
+
+    # INITIALISATION
+    def on_building_init_start(self) -> None:
```
AsadJeewa, line 31: There is a very large number of hooks. Do we handle this as we may at some point need to inject code at any of these points and in that regard, we can add or remove hooks as necessary?

mava/systems/jax/builder.py
```diff
@@ -13,28 +13,188 @@
...
+class Builder(SystemBuilder, CallbackHookMixin):
     def __init__(
         self,
         components: List[Any],
     ) -> None:
         """System building init
 
         Args:
-            components: system callback components
+            components: system callback blocks
```
AsadJeewa, line 35: so blocks are analogous to components that make up a system? why do we need the blocks class exactly?

mava/systems/jax/builder_test.py
```diff
@@ -0,0 +1,440 @@
...
+    def on_building_executor_environment(self, builder: SystemBuilder) -> None:
+        """_summary_"""
+        builder.blocks.env = (
+            builder.blocks.exec_logger,
+            self.config.executor_param_2,
+        )
+
+    def on_building_executor_environment_loop(self, builder: SystemBuilder) -> None:
+        """_summary_"""
+        if builder._executor_id != "evaluator":
```
AsadJeewa, line 191: Nice and clean :smile:

AsadJeewa (dismissed):
Lots of code but everything looks good from my side

mava/systems/jax/system_test.py
```diff
@@ -48,14 +47,18 @@ def dummy_str_plus_bool(self) -> None:
 
 
 # Mock builder
-class MockBuilder(Builder, MockCallbackHookMixin):
-    def __init__(self, components: List[Any]) -> None:
-        """Init for mock builder.
+class MockBuilder(MockCallbackHookMixin):
```
AsadJeewa, line 50: isn't the point to test the actual mixin class so why are we creating a mock class?

mava/callbacks/hook_mixin.py
arnupretorius: Yes, this is exactly right, so in the future, we might add more hooks if needed, but for now, we are creating the ones we think are most necessary.

mava/systems/jax/builder.py
arnupretorius: Oops I think this is a type that creeped in with a global find and replace 😅

arnupretorius: Will fix it.

mava/systems/jax/system_test.py
arnupretorius: This was before we had the actual mixin class but I still think it is good to create mock versions to isolate the testing to just the system. But that said, we must write some integration tests for all the system processes.

arnupretorius:
Closing this PR since it is being subsumed here: #457.

Issue #455, opened 2022-03-15
[BUG] Inconsistent Mypy behaviour


KaleabTessera:
### Describe the bug
Mypy behaviour is inconsistent. Even with the same version used in the pre-commits and setup.py, running the pre-commits/`mypy mava` locally is different to the results from github actions. 

### To Reproduce
1. Run mypy locally - `mypy mava` / `pre-commit run` -> results in no typing errors.
2. Push and create PR. Github actions run `mypy mava`, but this results in typing errors. 

### Expected behavior
Consistent behaviour locally and when run through github actions. 

### Context (Environment)
-

### Proposal for a fix
-

### Definition of done
-

### Mandatory checklist before making a PR
* [ ] The success criteria laid down in “Definition of done” are met.
* [ ] Code is documented - docstrings for methods and classes, static types for arguments.
* [ ] Code is tested - unit, integration and/or functional tests are added.
* [ ] Documentation is updated - README, CONTRIBUTING, or other documentation.
* [ ] All functional tests are green.
* [ ] Link experiment/benchmarking after implementation (optional).

PR #458, opened 2022-03-22
Bugfix/Mypy Inconsistency Issue

KaleabTessera:
## What?
Certain environments need to be updated for the new version of gym - e.g. debugging env. Currently, these envs + new version of gym = mypy issues.
## Why?
-
## How?
-
## Extra
Closes https://github.com/instadeepai/Mava/issues/455 .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,7 @@ setup(
         "matplotlib",
         "dataclasses",
         "box2d-py",
-        "gym",
+        "gym<=0.23.0",
     ],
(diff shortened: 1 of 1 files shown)


AsadJeewa (approved):
Thanks Kale-ab

mmorris44 (approved):
Thanks @KaleabTessera. Will try keep a vague eye out for when these dependency problems are no longer an issue, so we can remove the specific requirement

Issue #358, opened 2022-01-22
[BUG]: Mask out loss for padded experience in MADDPG/MAD4PG


DriesSmit:
**Describe the bug**
The loss for padded experience should still be masked out in MADDPG/MAD4PG.

Issue #357, opened 2022-01-22
[BUG] MADDPG loss arguments incorrect.


DriesSmit:
**Describe the bug**
The reward is fed into the wrong argument here: https://github.com/instadeepai/Mava/blob/develop/mava/systems/tf/maddpg/training.py#L1316

This needs to be fixed.

PR #460, opened 2022-03-23
Bugfix/MADD(4)PG

DriesSmit:
## What?
Implements a few fixes and improvements to MADD(4)PG namely, #358, #357 and 2 more fixes.
## Why?
Bugfixes.
## How?
Changed the trainer files for MADDPG and MAD4PG.
## Extra
.

### DIFF
--- a/mava/systems/tf/mad4pg/training.py
+++ b/mava/systems/tf/mad4pg/training.py
@@ -220,7 +220,12 @@ class MAD4PGBaseTrainer(MADDPGBaseTrainer):
                     dqda_clipping=dqda_clipping,
                     clip_norm=clip_norm,
                 )
-                self.policy_losses[agent] = tf.reduce_mean(policy_loss)
+                # Multiply by discounts to not train on padded data.
+                policy_mask = tf.reshape(step_not_padded, policy_loss.shape)
+                policy_loss = policy_loss * policy_mask
+                self.policy_losses[agent] = tf.reduce_sum(policy_loss) / tf.reduce_sum(
+                    policy_mask
+                )
         self.tape = tape
 
 
@@ -594,7 +599,7 @@ class MAD4PGBaseRecurrentTrainer(MADDPGBaseRecurrentTrainer):
         data: Trajectory = inputs.data
 
         # Note (dries): The unused variable is start_of_episodes.
-        observations, actions, rewards, discounts, _, extras = (
+        observations, actions, rewards, end_of_episode, _, extras = (
             data.observations,
             data.actions,
             data.rewards,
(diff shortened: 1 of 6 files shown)


mava/systems/tf/maddpg/execution.py
```diff
@@ -147,12 +147,19 @@ def select_action(
         # Step the recurrent policy/value network forward
         # given the current observation and state.
         action, policy = self._policy(agent, observation.observation)
-
-        # Return a numpy array with squeezed out batch dimension.
-        action = tf2_utils.to_numpy_squeeze(action)
-        policy = tf2_utils.to_numpy_squeeze(policy)
         return action, policy
 
+    @tf.function
```
arnupretorius, line 152: @DriesSmit @RuanJohn we should just confirm this change didn't cause the memory leak issue.

mava/systems/tf/maddpg/execution.py
```diff
@@ -367,9 +367,32 @@ def select_actions(
...
+        new_states = {}
         for agent, observation in observations.items():
-            actions[agent], policies[agent] = self.select_action(agent, observation)
-        return actions, policies
+            actions[agent], policies[agent], new_states[agent] = self.select_action(
+                agent, observation, states[agent]
+            )
+        return actions, policies, new_states
+
+    def select_action(  # type: ignore
```
arnupretorius, line 377: Why do we have the type ignore here? Seems like basic typing.

mava/systems/tf/mad4pg/training.py
```diff
@@ -594,7 +594,7 @@ def _forward(self, inputs: reverb.ReplaySample) -> None:
         data: Trajectory = inputs.data
 
         # Note (dries): The unused variable is start_of_episodes.
-        observations, actions, rewards, discounts, _, extras = (
+        observations, actions, rewards, end_of_episode, _, extras = (
```
arnupretorius, line 597: Much better naming :)

arnupretorius (commented):
Thanks for the fixes @DriesSmit! 🙂  Just see my few minor comments.

mava/systems/tf/maddpg/execution.py
DriesSmit: I think it was not conforming to the recurrent executor's argument setup. We now use tf.function which means the agent_state should also be passed in as an argument. We could maybe update the recurrent executor definition when all the systems use tf.function over the select_actions function.

KaleabTessera (approved):
Thanks @DriesSmit :fire: :open_hands:

mava/systems/tf/maddpg/execution.py
arnupretorius: Ok thanks, perhaps we can make an issue for this.

mava/systems/tf/maddpg/execution.py
arnupretorius: Discussed offline. Still under investigation but is not particular to MADDPG.

arnupretorius (approved):

mava/systems/tf/maddpg/execution.py
DriesSmit: Will do.

PR #470, opened 2022-04-01
fix: Update black version.

DriesSmit:
## What?
Upgrade the black package version.
## Why?
The current version of black seems to be producing an error.
## How?
Updated the setup.py file.
## Extra
.


KaleabTessera (approved):
Thanks @DriesSmit !!

arnupretorius (approved):
Thanks @DriesSmit 👍

PR #475, opened 2022-04-12
feat: Add fixed sampler capability + bugfixes

DriesSmit:
## What?
Add the capability to fix the sampler for MADDPG/MAD4PG. Also implemented bug fixes with @EdanToledo and @sash-a to get agents with different network architectures working in Mava.
## Why?
@EdanToledo and @sash-a would like the feature of having fixed networks for each agent, while each trainer only trains on one network.
## How?
Added a `fix_sampler` variable to the MADDPG/MAD4PG systems that allow a user to create a fixed sampling order for the agent networks.
## Extra
@sash-a and @EdanToledo please take a look at the new example and see if this would work for you?

### DIFF
--- a/examples/tf/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers_fixed_nets.py
+++ b/examples/tf/debugging/simple_spread/feedforward/decentralised/run_maddpg_scale_trainers_fixed_nets.py
@@ -1,0 +1,118 @@
+# python3
+# Copyright 2021 InstaDeep Ltd. All rights reserved.
+#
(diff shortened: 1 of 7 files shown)


sash-a (approved):
I still think `list_of_networks` is a bit of a vague name, but otherwise looks good

sash-a (approved):
Thanks for changing my nitpicky comment. All looks good :fire:

sash-a (approved):
Even better :fire:

KaleabTessera (approved):
Thanks @DriesSmit :open_hands:

Issue #473, opened 2022-04-11
[FEATURE] piecewise linear epsilon for exploration


nima-siboni:
### Feature
A piecewise linear scheduler for epsilon. With piecewise linear scheduler the user can increase
and decrease the epsilon over the desired time intervals.

### Proposal
Creating a new class for similar to LinearTimestepScheduler

### Testing
The returned value of the epsilon should be tested:
* The returned value right after creation of an instance
* The returned value after updating to a timestep before the smallest timestep in the given list, i.e. the initial fallback value
* The returned value after updating to a timestep beyond the largest timestep in the given list, i.e. the final fallback value
* The returned value interpolating between the given timesteps and epsilons

### Definition of done
A class is implemented and all the tests are passed.

PR #476, opened 2022-04-12
Feature/piecewise linear epsilon for exploration

nima-siboni:
## What?
A new epsilon scheduler is implemented, i.e. a piecewise linear scheduler for epsilon. With piecewise linear scheduler the user can increase and decrease the epsilon over the desired time intervals. Refer to issue [#473](https://github.com/instadeepai/Mava/issues/473).
## Why?
This feature offers more control over scheduling the epsilons.
## How?
A new class is implemented for this feature which, similar to other schedulers, has ```get_epsilon``` method. The class is called ```PieceWiseLinearTimestepScheduler``` and it is in ```exploration_scheduling.py```. Some tests are also added in the corresponding test directory.
## Extra
* In principle it would have been cleaner to implement the new class based on ```BaseExplorationTimestepScheduler```, but the init signature of this new class is very different from the base class.

* Closes the issue #473

### DIFF
--- a/mava/components/tf/modules/exploration/exploration_scheduling.py
+++ b/mava/components/tf/modules/exploration/exploration_scheduling.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import abc
+from typing import List
 
 import numpy as np
 import tensorflow as tf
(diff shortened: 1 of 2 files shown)


mava/components/tf/modules/exploration/exploration_scheduling.py
```diff
@@ -235,3 +236,69 @@ def decrement_epsilon(self) -> float:
...
+        else:
+            for i in range(len(self._timesteps) - 1):
+                if self._timesteps[i] <= time_t <= self._timesteps[i + 1]:
+                    epsilon_begin = self._epsilons[i]
+                    epsilon_end = self._epsilons[i + 1]
+                    timestep_begin = self._timesteps[i]
+                    timestep_end = self._timesteps[i + 1]
+                    epsilon_delta = epsilon_end - epsilon_begin
+                    timestep_delta = timestep_end - timestep_begin
+                    break
```
sash-a, line 291: I think `np.searchsorted(self._timesteps, time_t, side="right")` should return the correct index and is a bit cleaner than looping and breaking

sash-a: In fact doing this should allow you to prepend `initial_fallback_epsilon` and `final_fallback_epsilon` to epsilons and remove the two if statement checks also.

tests/components/tf/modules/exploration/exploration_scheduling_test.py
```diff
@@ -0,0 +1,59 @@
+from mava.components.tf.modules.exploration.exploration_scheduling import (
+    PieceWiseLinearTimestepScheduler as PWLTS,
+)
+
+
+class TestExplorationScheduling:
+    """Testing the exploration schedule."""
+
+    @staticmethod
+    def create_piecewise_linear_exploration_schedule() -> PWLTS:
```
sash-a, line 10: This should probably be a fixture

KaleabTessera: I agree with @sash-a , can we please make this a `pytest.fixture` or use a parametrizing test (https://docs.pytest.org/en/6.2.x/parametrize.html).

mava/components/tf/modules/exploration/exploration_scheduling.py
```diff
@@ -235,3 +236,69 @@ def decrement_epsilon(self) -> float:
...
+class PieceWiseLinearTimestepScheduler:
+    """A piecewise linear scheduler for epsilon.
+
+    Args:
+        timesteps : list of timesteps at which epsilon is changed. The list should be in
+        ascending order.
+        epsilons : list of epsilon values at each timestep.
+        initial_fallback_epsilon : epsilon value to use when timestep is less than the
+        first timestep in the list.
+        final_fallback_epsilon : epsilon value to use when timestep is greater than the
```
KaleabTessera, line 250: I don't think we should use `initial_fallback_epsilon` and `final_fallback_epsilon`. I think we should stick to an implementation close to https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/schedules/PiecewiseConstantDecay, where the first epsilon in the list (`epsilon[0]`) is used until the first `timesteps[0]`. 

It would be great to test that this and the tensorflow scheduler behave similiarily. That way we can use for our jax systems, without need to rely on tf :)

KaleabTessera (changes requested):
Thanks and well done on your first PR @nima-siboni :partying_face: :fire: 

I think we should make some minor changes to the implementation.

mava/components/tf/modules/exploration/exploration_scheduling.py
nima-siboni: Tensorflow has piecewise constant schedule, but for an RL project it makes more sense to have piecewise linear. Something similar to what is implemented in RLlib as in [here](https://docs.ray.io/en/latest/_modules/ray/rllib/utils/schedules/piecewise_schedule.html#PiecewiseSchedule).

That is a choice to be made, I guess it make more sense to have a piecewise linear exploration. Please let me know what you think.

mava/components/tf/modules/exploration/exploration_scheduling.py
nima-siboni: I agree that this way it becomes cleaner.

nima-siboni: Thanks :D

tests/components/tf/modules/exploration/exploration_scheduling_test.py
nima-siboni: done! :)

mava/components/tf/modules/exploration/exploration_scheduling.py
KaleabTessera: Thanks for clarifying @nima-siboni ! Makes more sense now. Could we please follow @sash-a feedback about `prepend [ing] initial_fallback_epsilon and final_fallback_epsilon to epsilons`. This would lead to less possible confusion, e.g. now when I init the component and then call `get_epsilon` , I expect to get `initial_fallback_epsilon` from the description , but instead I believe we get `epsilon[0]`.

Having an `epsilon` and `timesteps` would simplify this a lot and make it easier for people to use.

KaleabTessera:
Closing for now. Will re-open at a later stage.

Issue #479, opened 2022-04-19
[BUG] Importing smac fails due PZ version


KaleabTessera:
### Describe the bug
Importing `smac` fails because it uses a new version of pz (https://github.com/oxwhirl/smac/commit/6d2191b30f1a0a5b6059b6b13cb594baace73341). Mava uses old version and this causes an error.  

E.g. error - https://github.com/instadeepai/Mava/runs/6076208574?check_suite_focus=true 

### To Reproduce
1. Run tests.

### Expected behavior
Tests to pass and code to run.

### Context (Environment)
-

### Additional context
-

### Possible Solution
-

PR #480, opened 2022-04-19
Bugfix/ Update PZ Version and new jax dockerfiles

KaleabTessera:
## What?
- Updated pz env version.
- Added dockerfile and requirements for jax.
## Why?
- Smac wrapper uses latest version of pz, that causes importing smac. 
- New jax images help resolve a bug with version of jax. 
## How?
-
## Extra
Closes https://github.com/instadeepai/Mava/issues/479.


RuanJohn (approved):
Thanks @KaleabTessera :muscle:

PR #483, opened 2022-04-20
Feature/Only init component once.

KaleabTessera:
## What?
Ensure components are only initialized once. 
## Why?
Component init could be computationally expensive and so it should only be done when necessary.
## How?
Made `name` and `config_class` static methods so that they can be accessed without initializing the components. 
## Extra
@arnupretorius @DriesSmit Similiar to what we discussed earlier today.

### DIFF
--- a/mava/components/jax/building/adders.py
+++ b/mava/components/jax/building/adders.py
@@ -16,7 +16,7 @@
 """Commonly used adder components for system builders"""
 import abc
 from dataclasses import dataclass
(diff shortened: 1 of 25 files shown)


arnupretorius (approved):
This looks great, thanks @KaleabTessera! 🙂

DriesSmit (approved):
Thanks @KaleabTessera :fire: This looks great :slightly_smiling_face:

PR #484, opened 2022-04-22
Feature/jax data server tests

KaleabTessera:
## What?
- Test for data server + Relevant Mocks.
- Simplification and decoupling of config e.g. moving some config from Executor Process to a general system init/setup process so we use other component independent of the executor. 
## Why?
- Better testing of redesign.
- Ease of use for new redesign. 
## How?
-
## Extra
-

### DIFF
--- a/mava/components/jax/building/adders_test.py
+++ b/mava/components/jax/building/adders_test.py
@@ -16,18 +16,22 @@
 """Tests for config class for Jax-based Mava systems"""
 
 from dataclasses import dataclass, field
-from typing import Dict, List, Tuple
+from typing import Callable, Dict, List, Optional, Tuple
 
 import numpy as np
 import pytest
-import reverb
 from acme import specs
 
 from mava.components.jax import Component
 from mava.components.jax.building import adders
-from mava.core_jax import SystemBuilder
+from mava.components.jax.building.base import SystemInit
+from mava.components.jax.building.environments import EnvironmentSpec
 from mava.specs import DesignSpec, MAEnvironmentSpec
 from mava.systems.jax.system import System
+from mava.testing.building.mocks import (
+    MockOnPolicyDataServer,
+    make_fake_environment_factory,
+)
 from mava.utils.wrapper_utils import parameterized_restart
 
 agents = {"agent_0", "agent_1", "agent_2"}
@@ -65,44 +69,6 @@ def make_fake_env_specs() -> MAEnvironmentSpec:
 
 
 @dataclass
-class MockConfig:
-    pass
-
-
-class MockTestSetup(Component):
-    """_summary_"""
-
-    def __init__(self, config: MockConfig = MockConfig()) -> None:
-        """_summary_"""
-        self.config = config
-
-    def on_building_data_server(self, builder: SystemBuilder) -> None:
-        """_summary_"""
-        env_spec = make_fake_env_specs()
-        builder.store.server = reverb.Server(
-            [
-                reverb.Table.queue(
-                    name="data_server",
-                    max_size=100,
-                    signature=builder.store.adder_signature_fn(
-                        env_spec, builder.store.sequence_length, {}
-                    ),
-                )
-            ]
-        )
-        builder.store.system_data_server = reverb.Client(
-            f"localhost:{builder.store.server.port}"
-        )
-        builder.store.unique_net_keys = ["network_0"]
-        builder.store.table_network_config = {"table_0": "network_0"}
-
-    @staticmethod
-    def name() -> str:
-        """_summary_"""
-        return "setup"
-
-
-@dataclass
 class DistributorDefaultConfig:
     num_executors: int = 1
     nodes_on_gpu: List[str] = field(default_factory=list)
@@ -130,6 +96,15 @@ class MockDistributor(Component):
         """
         return "distributor"
 
+    @staticmethod
+    def config_class() -> Optional[Callable]:
+        """Config class used for component.
+
+        Returns:
+            config class/dataclass for component.
+        """
+        return DistributorDefaultConfig
+
 
 class TestSystemWithParallelSequenceAdder(System):
     def design(self) -> Tuple[DesignSpec, Dict]:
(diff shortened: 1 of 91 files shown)


mava/components/jax/building/data_server.py
```diff
@@ -44,6 +44,10 @@ def __init__(
 
     def _create_table_per_trainer(self, builder: SystemBuilder) -> List[reverb.Table]:
         data_tables = []
+        # Default table network config - often overwritten by TrainerInit.
+        if not hasattr(builder.store, "table_network_config"):
+            builder.store.table_network_config = {"table_0": "network_0"}
```
DriesSmit, line 49: Shouldn't we rather create a mock init that creates this config? Instead of having a default in the `data_server`?

mava/components/jax/building/environments.py
```diff
@@ -24,11 +24,48 @@
...
+        """Config class used for component.
+
+        Returns:
+            config class/dataclass for component.
+        """
+        return EnvironmentSpecConfig
+
+
+@dataclass
+class ExecutorEnvironmentLoopConfig:
```
DriesSmit, line 68: What was the reason for creating two components (EvironmentSpec and EnvironmentLoop) instead of one?

DriesSmit (commented):
Looks great @KaleabTessera :raised_hands: :fire: See my two comments.

mava/components/jax/building/environments.py
KaleabTessera: @DriesSmit From my perspective it makes sense to split the two. `EvironmentSpec` is only about the spec of the env, while  `EnvironmentLoop` is env loop that gets run e.g. `ParallelEnvironmentLoop`. They serve different uses from my perspective and it also make testing easier, e.g. a component might only need an `EvironmentSpec` to be tested and not nec a whole env loop. Thoughts?

mava/components/jax/building/data_server.py
KaleabTessera: @DriesSmit So I thought about this. Having a mock init would work for the tests, but I felt that a data server component should be independent of a trainer, to increase flexibility. E.g. maybe we create a new style of trainer that possibly doesn't rely on `table_network_config`. 

Having a default in the data server means if no `table_network_config` is set , it can still work similarly to how it worked before multiple trainers. 

Thoughts on this?

mava/components/jax/building/adders_test.py
```diff
@@ -157,11 +131,15 @@ def test_system_parallel_sequence_adder() -> System:
     return TestSystemWithParallelSequenceAdder()
 
 
+# TODO Fix test.
```
arnupretorius, line 134: Should this first get fixed?

mava/components/jax/building/rate_limiters.py
```diff
@@ -37,12 +37,26 @@ def __init__(self, config: RateLimiterConfig = RateLimiterConfig()) -> None:
         self.config = config
 
     @abc.abstractmethod
-    def on_building_rate_limiter(self, builder: SystemBuilder) -> None:
+    def on_building_data_server_rate_limiter(self, builder: SystemBuilder) -> None:
```
arnupretorius, line 40: Did you make sure to also update the callbacks.py and mixin class with this hook change? I don't see any changes to those files?

mava/systems/jax/config.py
```diff
@@ -56,15 +56,23 @@ def add(self, **kwargs: Any) -> None:
                     new_param_names = list(dataclass.__dict__.keys())
                     if set(self._current_params) & set(new_param_names):
                         raise Exception(
-                            "Component configs share a common parameter name. \
-                            This is not allowed, please ensure config \
-                            parameter names are unique."
+                            f"""
+                            Component configs share a common parameter name -
```
arnupretorius, line 60: Nice :)

arnupretorius (commented):
Thanks @KaleabTessera! 🙂 Just see my comments. Also, I made a quick scan of this PR a while back and saw we had a `config_class` function in the callbacks class, which I thought wasn't necessary since we already have it in the abstract component class, but I couldn't see it this time. Did you remove it?

mava/components/jax/building/adders_test.py
KaleabTessera: I was thinking of doing that in another PR, since this one mainly focuses on the data server tests. I will probably create a new one for the adders.

mava/components/jax/building/rate_limiters.py
KaleabTessera: So it is already called `on_building_data_server_rate_limiter` in the callbacks (https://github.com/instadeepai/Mava/blob/8af6b01f67969d5dd4889e00391aa76d943e5b60/mava/callbacks/base.py#L59). I think it was a mistakenly called `on_building_rate_limiter` here.

KaleabTessera:
@arnupretorius Thanks for the review. Yep, I removed the redundant config class method in the callbacks with this commit (https://github.com/instadeepai/Mava/pull/484/commits/9197a2c678b6664f75a6917334d1b881033bf28e).

mava/components/jax/building/environments.py
DriesSmit: If it makes testing easier it sounds fine to me :slightly_smiling_face:

mava/components/jax/building/data_server.py
DriesSmit: I see. I agree that the `data_server` component should be independent of the trainer :+1:  We however need to slightly update the default value as this won't work for agents that don't share weights. Maybe use something like this:

DriesSmit: You can actually just replace line 49 with an error message telling the user to add a component that creates the table network config.

KaleabTessera: I think I get what you are saying @DriesSmit , but not having a default means you need another component to do something that feels like the `data_server` should be able to handle. 

My thinking - have a reasonable default for `table_network_config`, if other components do trickier stuff , then they overwrite. 

Maybe we can solve this by using a better default or do you think something else?

DriesSmit: Okay maybe the following will work then. We update line 49 to be:
`builder.store.table_network_config = {"table_0": sort_str_num(builder.store.agent_net_keys.values())}`. If their is a trainer this gets overwritten. Otherwise this default is used. 

Underneath that line we put:
`assert builder.store.network_sampling_setup == enums.NetworkSampler.fixed_agent_networks`. This is just to make sure that the sampler is the correct setting for this assumed default. How does that sound?

KaleabTessera: @DriesSmit Thanks for the pointers. I have made the updates :)

arnupretorius (approved):

arnupretorius (approved):

PR #490, opened 2022-04-25
fix: Fix parameter_client and batch_size in trainer component.

DriesSmit:
## What?
Fix two bugs found by @EdanToledo. 
1.) Fix a small variable scope issue in the JAX parameter client.
2.) Fix a bug where the epoch batch_size argument is configurable. This variable should not be specified but inferred from the sequence length and the TrajectoryDataset batch size.
## Why?
Bugfixes.
## How?
Made changes to the parameter client and MAPGEpochUpdate component.


KaleabTessera (approved):
Thanks @DriesSmit :handshake: :fire:

PR #491, opened 2022-05-03
Feature/jittable agent ids

EdanToledo:
## What?
Change agent ids to use EntityID dataclass - this dataclass can be easily used to represent an agent and is jittable. The dataclass provides utility functions to convert an entityID object to a string and vice versa. The entityID object hashes in such a way that the string representation can be used to index dictionaries and vice versa.
## Why?
Strings are not a primitive jax type and as such causes a lot of problems when trying to pass agent information to any jitted function. Additionally, the entityID dataclass provides a standardised way to represent agent ids which allow easy access to information such as type and id. This eliminates the need to perform string manipulation which can be a large source of errors.
## How?
An entityId dataclass object was made with the dunder string, dunder hash and dunder equals methods overridden to allow any entityID object to be equally represented as a string. Additionally, utility functions were created to allow users to easily convert between the two. The debugging environment was changed to use entityID objects instead of strings. The mava specs class was changed to use the standardised form of entityID strings.


sash-a:
Just a note because I don't think this will be merged in before I go on leave.

To merge this in we will also have to change env wrappers to be compatible, just two methods need to be changed `possible_agents` and `agents`

As an example you would change `possible_agents` from:
```python
def possible_agents(self):
  return [f"agent_{i} for i in range(self.num_agents)
```

to

```python
def possible_agents(self):
  return [f"EntityId(id=i) for i in range(self.num_agents)
```

I really think this is a nice change to mava, aside from removing one of the bigger obstacles when jitting environments it also adds a lot of stafety when manipulating agent id's as you no longer need to get the substring you can simply call `agent_id.id` or `agent_id.type`

EdanToledo:
I agree

Best Regards,
Edan Toledo
________________________________

DriesSmit:

Issue #492, opened 2022-05-04
[BUG] Inconsistent Logging of Jax Systems


KaleabTessera:
### Describe the bug
Jax system logs look different compared to tf system logs. 

### To Reproduce
1. Run jax example.
2. Run tf example.
3. Compare tensorboard logs. 

### Expected behavior
Same logs. 

### Context (Environment)
-

### Additional context
-

### Possible Solution
Use the same env stats wrapper.

DriesSmit:
This issue has been resolved.

PR #493, opened 2022-05-04
Fix/jax consistent logging

KaleabTessera:
## What?
Makes jax logs consistent with tf logs. 
## Why?
Easier to compare jax and tf systems. 
## How?
Add env stats wrapper to jax parallel env loop component. 
## Extra
- Closes https://github.com/instadeepai/Mava/issues/492 .
- Also ensured jax containers are pushed before tf containers in github actions.

### DIFF
--- a/mava/components/jax/building/environments.py
+++ b/mava/components/jax/building/environments.py
@@ -16,7 +16,7 @@
 """Execution components for system builders"""
 import abc
 from dataclasses import dataclass
-from typing import Callable, Optional
+from typing import Any, Callable, Optional
 
 import acme
 
@@ -25,6 +25,7 @@ from mava.components.jax import Component
 from mava.core_jax import SystemBuilder
 from mava.environment_loop import ParallelEnvironmentLoop
 from mava.utils.sort_utils import sort_str_num
+from mava.wrappers.environment_loop_wrappers import DetailedPerAgentStatistics
 
 
 @dataclass
@@ -67,6 +68,7 @@ class EnvironmentSpec(Component):
 @dataclass
 class ExecutorEnvironmentLoopConfig:
     should_update: bool = True
+    executor_stats_wrapper_class: Any = DetailedPerAgentStatistics
 
 
 class ExecutorEnvironmentLoop(Component):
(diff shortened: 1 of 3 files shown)


RuanJohn (approved):
Thank you for this @KaleabTessera :muscle:

mava/components/jax/building/environments.py
```diff
@@ -67,6 +68,7 @@ def config_class() -> Optional[Callable]:
 @dataclass
 class ExecutorEnvironmentLoopConfig:
     should_update: bool = True
+    executor_stats_wrapper_class: Any = DetailedPerAgentStatistics
```
sash-a, line 71: Should this not be an `Optional[DetailedPerAgentStatistics]` or it's base class?

arnupretorius: I think it is good to have the default here since without it I don't think we track the trainer stats etc. Or are you mostly referring to the type that we should change from `Any`? For that, I agree, the base class could be more appropriate. In general, going forward, I think we should do a bit of a rethink regarding these stats wrappers and find a cleaner way using the redesign instead of passing in wrapper classes in the config.

arnupretorius (approved):
Look good to me, thanks @KaleabTessera 👍 

I think perhaps at some point we can rethink the stats wrappers a bit and try making components for them instead of passing in the wrapper classes in the config to a component.

mava/components/jax/building/environments.py
KaleabTessera: I updated the type @sash-a @arnupretorius . I agree with redesigning the wrappers in future.

KaleabTessera:
This is a temporary solution to get the eval and exec vars logging correctly. In future, once we have deprecated our tf systems, we can simplify the environment wrappers.

mava/components/jax/building/environments.py
sash-a: I was refering to the type to change from `Any`. But I think the current solution is good until we think of a cleaner way.

sash-a:
Looks good to me @KaleabTessera :fire:

mava/wrappers/environment_loop_wrappers.py
```diff
@@ -114,9 +114,7 @@ def _compute_episode_statistics(
 
         # Record counts.
         if hasattr(self._executor, "_counts"):
-            loop_type = "executor"
-            if "_" not in self._loop_label:
-                loop_type = "evaluator"
+            loop_type = "evaluator" if "_" not in self._loop_label else "executor"
```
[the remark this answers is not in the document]
KaleabTessera: @mmorris44 Agreed. I want to use `self._executor.store.is_evaluator` , but for now we need to do this to be compatible with tf.

PR #495, opened 2022-05-06
fix: removed obs flattening from executor's action selection

### LINKED ISSUE #494: [BUG][Jax] Default executor flattens observations
### Describe the bug
The default executor flattens observations which is undesirable for environments which can benefit from spatial knowledge.

### To Reproduce
`observation = executor.store.observation.observation.reshape((1, -1))` <- inside `action_selection.py`

### Expected behavior
It was expected that observations would be fed to neural networks unedited.

### Possible Solution
`observation = utils.add_batch_dim(executor.store.observation.observation)`

sash-a:
Closes #494

Removed implicit obs flattening from executor's action selection. It should rather be done by the networks

## What?
Changed `action_selection.py` to not flatten observations before passing them to the neural network and instead unsqueeze them.
## Why?
Because the current behavior is undesirable for grid world style environments where agents can benefit from spatial knowledge
## How?
The change simply adds a batch dimension without flattening the observations

### DIFF
--- a/mava/components/jax/executing/action_selection.py
+++ b/mava/components/jax/executing/action_selection.py
@@ -18,6 +18,7 @@
 from dataclasses import dataclass
 
 import jax
+from acme.jax import utils
 
 from mava.components.jax import Component
 from mava.core_jax import SystemExecutor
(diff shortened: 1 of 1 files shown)


AsadJeewa:
Thanks @sash-a

AsadJeewa (approved):

KaleabTessera (approved):
Thanks @sash-a !

PR #496, opened 2022-05-06
Feature/jax environment support

EdanToledo:
## What?
New components were added to use JAX Environments - these components are the Jax environment loop and the Jax executor environment loop. Additionally, a seed value can now be passed to the distributor which is used to create random seeds for each executor.
## Why?
JAX environments were not supported. 
## How?
The new components allow any system to use JAX environments so long as they abide by the "state, timestep, extras" API
## Extras:
The commits are going to be slightly incorrect. This is due to branching off a massively changed branch.


mava/systems/jax/launcher.py
```diff
@@ -16,8 +16,10 @@
 """General launcher for systems"""
 from typing import Any, Dict, List, Union
 
+import jax.random as random
 import launchpad as lp
 import reverb
+from chex import PRNGKey
```
DriesSmit, line 22: Is this imported and not used?

mava/systems/jax/launcher.py
```diff
@@ -16,8 +16,10 @@
 """General launcher for systems"""
 from typing import Any, Dict, List, Union
 
+import jax.random as random
```
DriesSmit, line 19: Is this imported and not used?

DriesSmit:

PR #497, opened 2022-05-06
feature: added useful utils for working with stacked pytrees

sash-a:
## What?
Utils for working with stacked pytrees
## Why?
Found that this code was being repeated often and these methods were very useful


mmorris44 (changes requested):
Happy to merge this shortly. Could we just rename the file to `jax_tree_utils` please? Long-term we perhaps even ought to partition the utils into `tf`-specific ones, `jax`-specific ones, and general ones. For now, renaming the file will do, I think.

mmorris44 (approved):
Happy with it now. Thanks @sash-a!

Issue #500, opened 2022-05-06
[FEATURE] Creating epsilon scheduler component and classes


nima-siboni:
### Feature
A component for epsilon scheduler, and some common epsilon scheduling logics are implemented
### Proposal
* A component is created for epsilon scheduling which has the following callbacks:
> ```on_execution_init_end``` which assigns an instance epsilon_scheduler class (for example linear epsilon scheduler) to the store of the executor.
> ```on_execution_update``` which updates the epsilon_scheduler which is in the store. This is commonly called at the end of execution of a observation in an episode. Refer to ```ParallelEnvironmentLoop``` class where ```self._executor.update()``` is called.

* An abstract base class for implementation of different epsilon schedulers is created.
* A number of concrete epsilon schedulers are implemented using the above base class.

### Testing
* Tests are done for concrete epsilon schedulers to return proper epsilon based on the timestep.
* Tests for the component is postponed to the time that they are used in MADQN.

### Definition of done
The component is implemented.
The epsilon_scheduler class is implemented and tested.

### Mandatory checklist before making a PR
* [ ] The success criteria laid down in “Definition of done” are met.
* [ ] Code is documented - docstrings for methods and classes, static types for arguments.
* [ ] Code is tested - unit, integration and/or functional tests are added.
* [ ] Documentation is updated - README, CONTRIBUTING, or other documentation.
* [ ] All functional tests are green.
* [ ] Link experiment/benchmarking after implementation (optional).

PR #501, opened 2022-05-06
feat: epsilon scheduler components and classes are added

nima-siboni:
## What?
Epsilon scheduler component and classes are added.
## Why?
Epsilon scheduler is needed for exploration in DQN-like algorithms.
## How?

* A component is created for epsilon scheduling which has the following callbacks:
> ```on_execution_init_end``` which assigns an instance epsilon_scheduler class (for example linear epsilon scheduler) to the store of the executor.
> ```on_execution_update``` which updates the epsilon_scheduler which is in the store. This is commonly called at the end of execution of a observation in an episode. Refer to ```ParallelEnvironmentLoop``` class where ```self._executor.update()``` is called.

* An abstract base class for implementation of different epsilon schedulers is created.
* A number of concrete epsilon schedulers are implemented using the above base class.

## Extra
Resolves issue #500


DriesSmit:

PR #502, opened 2022-05-06
feature: add optional observation network to ppo default networks

sash-a:
## What?
PPO default networks now takes an optional observation argument
## Why?
To allow for observation networks in cases where it would be beneficial (such as 2D observations)
## How?
The `observation_network` argument was added to the three network creation methods in `networks.py`. It defaults to a `batch_concat` so that the observations are 1D and the current behavior is preserved by default. However, the user can specify a custom network for this, an example can be seen in `run_mappo_obs_net.py`


examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo_obs_net.py
```diff
@@ -0,0 +1,119 @@
...
+
+    Args:
+        _ : _
+    """
+    # Environment.
+    environment_factory = functools.partial(
+        debugging_utils.make_environment,
+        env_name=FLAGS.env_name,
+        action_space=FLAGS.action_space,
+    )
```
sash-a, line 61: Currently there is no env in mava that produces 2D observations (that I could find). So this example does fail, as the network is an Atari net which has convs and thus expects a 2D input.

KaleabTessera:
Hi @sash-a , thanks for this! 

Please use pong or co-op pong in the example (https://github.com/instadeepai/Mava/blob/develop/examples/tf/petting_zoo/atari/pong/recurrent/decentralised/run_madqn.py or https://github.com/instadeepai/Mava/blob/develop/examples/tf/petting_zoo/butterfly/cooperative_pong/feedforward/decentralised/run_mappo.py) . These envs have pixel obs.

mmorris44 (commented):
~~I'm having trouble actually running this (in docker). We can only build one target at a time, so I can't build `jax-core` and `pz` at the same time to be able to run this code.~~ -> use target `pz-jax`. My bad

mmorris44 (changes requested):
Requested changes due to one line I'm confused by. Otherwise looks good!

mmorris44 (dismissed):
Approval pending confirmation that example runs fine in docker

mmorris44 (approved):
Approved! Thanks @sash-a :)

Issue #503, opened 2022-05-06
[FEATURE] Creating the executor for MADQN


nima-siboni:
### Feature
As the first step to have the first MADQN system, the executor is created.

### Proposal
The MAPPO design in jax/systems/mappo and its executor are modified for which :
* ```FeedforwardExecutorSelectAction``` is changed to ```FeedforwardExecutorSelectActionValueBased```
* ```PPONetworks``` is changed to ```DQNNetworks```
* ```EpsilonScheduler``` is added

### Testing
Tests are same as MAPPO system_tests with additional tests:
* steps_counter
* epsilon scheduler 

### Definition of done
A precise outline for the feature to be considered fully implemented, i.e. done/complete.
The executor is implemented and passes the tests.

### Mandatory checklist before making a PR
* [ ] The success criteria laid down in “Definition of done” are met.
* [ ] Code is documented - docstrings for methods and classes, static types for arguments.
* [ ] Code is tested - unit, integration and/or functional tests are added.
* [ ] Documentation is updated - README, CONTRIBUTING, or other documentation.
* [ ] All functional tests are green.
* [ ] Link experiment/benchmarking after implementation (optional).

PR #504, opened 2022-05-06
feat: MADQN executor is added.

nima-siboni:
Resolves Issue #503

## What?
MADQN's executor is implemented.
## Why?
The first step towards a complete MADQN system.
## How?
The MAPPO design in jax/systems/mappo and its executor are modified for which :
* ```FeedforwardExecutorSelectAction``` is changed to ```FeedforwardExecutorSelectActionValueBased```
* ```PPONetworks``` is changed to ```DQNNetworks```
* ```EpsilonScheduler``` is added to the system design
[Describe how the change was made i.e. describe technically what the change does.]
## Extra


DriesSmit:

PR #505, opened 2022-05-10
feat: add PPO action masking for the discrete setting

EdanToledo:
## What?
Changed MAPPO's networks class and the feedforward action selection executor to support discrete action masking using OLT's legal actions.
## Why?
Improved PPO performance.
## How?
The MAPPO network's get_action and forward_function now can take in an optional mask whereby it will create a new distribution with masked logits. The masking has been done such that gradients can still be calculated. 
## Extra?
The action masking has been tested on smac and doesn't crash. It also managed to get a decent score on 3m.


KaleabTessera:
@EdanToledo Looks good, but can you please fix the mypy issues. :pray:

KaleabTessera (approved):
Thanks @EdanToledo :fire: :open_hands:

PR #506, opened 2022-05-16
fix: changed confusing Executor parameter name

sash-a:
## What?
Changed name of parameter in the executor store
## Why?
`config` is a confusing parameter name, especially when looking at the executor creation in the builder where it was

```python
self.store.executor = Executor(
    config=self.store,
    components=self.callbacks,
)
```

Personally was getting confused thinking it wanted a config class and not the store.


AsadJeewa (approved):

KaleabTessera:
Thanks @sash-a !

PR #507, opened 2022-06-01
Allow for multiple launchpad terminal configurations

RuanJohn:
## What?
Added support for displaying processes in separate terminals. 

## Why?
It was possible to do in tensorflow mava, but not yet implemented in the redesigned mava.


mmorris44 (approved):
Perfect

PR #508, opened 2022-06-01
fix: component names not matching design spec keys

mmorris44:
## What?
- Added check in system `__init__` to ensure design spec keys match component names
- Raise Error if abstract `name()` function of `Component` not implemented
- Change MAPPO component names to match MAPPO design spec keys

## Why?
- Component keys are used when adding and updating components, but they are compared to component names, accessed with the abstract `name()` method

## How?
- Added check in system `__init__` to ensure design spec keys match component names. Raise exception if mismatch
- Raise Error if abstract `name()` function of `Component` not implemented
- Change MAPPO component names to match MAPPO design spec keys
- Added superclasses for certain components which have the `name()` method defined

## Extra
- For future work: need to ensure sensible names for components, good use of inheritance for components, and try match component class names to actual names

### DIFF
--- a/mava/systems/jax/system.py
+++ b/mava/systems/jax/system.py
@@ -41,11 +41,7 @@ class System(BaseSystem):
         for key, value in self._design.get().items():
             if key != value.name():
                 raise Exception(
-                    "Component '"
-                    + key
-                    + "' has mismatching name '"
-                    + value.name()
-                    + "'"
+                    f"Component '{key}' has mismatching name '{value.name()}'"
                 )
 
     def _make_config(self) -> None:


mava/systems/jax/system.py
```diff
@@ -37,6 +37,17 @@ def __init__(self) -> None:
...
 
+        # Enforce that design keys match component names
+        for key, value in self._design.get().items():
+            if key != value.name():
+                raise Exception(
+                    "Component '"
+                    + key
+                    + "' has mismatching name '"
+                    + value.name()
+                    + "'"
```
sash-a, line 48: Super pedantic, but fstring would make this a lot more readable: `f"Component '{key}' has mismatching name '{value.name}'"`

sash-a (changes requested):
Nice change should make new components easier to make! :fire: 

Just the one fstring change and then I think it's good

mava/systems/jax/system.py
mmorris44: Agreed, thanks @sash-a. Will change now

sash-a (approved):

RuanJohn (approved):
Looks good to me. Thank you @mmorris44! :call_me_hand:

PR #509, opened 2022-06-01
Feature/jax upgrade networks upgrade acme

KaleabTessera:
## What?
- Change `mlp` to `layernormmlp` in ppo to be consistent with our tf systems.
- Upgrade acme, reverb, tf and launchpad (we will have to benchmark this). 
## Why?
- 
## How?
-
## Extra
-


mmorris44 (changes requested):

mmorris44 (dismissed):
Approved, thanks @KaleabTessera. Merge pending successful benchmarks

sash-a:
Something that's been bugging me related to this change. The `critic_layer_sizes` param in `make_default_networks` isn't used. I think if the default nets are going to use a `CategoricalValueHead` then we should probably just have a `policy_value_layer_sizes` param instead. What do you think?

mmorris44:
Any updates on the benchmark results for this @KaleabTessera @RuanJohn? Are we ready to merge?

KaleabTessera:
@mmorris44 Just pausing this for now. We want to compare the acme,tf,reverb upgrades on tf systems first. We are doing this on another PR.

KaleabTessera:
@arnupretorius @mmorris44 @RuanJohn PR is ready now. Nice and small one :) 

@RuanJohn Can you do a couple of runs on this to see if the performance remains stable on the jax networks with this change (if you already didn't in the past)?

RuanJohn:
> @arnupretorius @mmorris44 @RuanJohn PR is ready now. Nice and small one :)
> 
> @RuanJohn Can you do a couple of runs on this to see if the performance remains stable on the jax networks with this change (if you already didn't in the past)?

@KaleabTessera, will start some runs this afternoon :call_me_hand:

RuanJohn:
> Should we perhaps now use this PR to try update `numpy`, as well as some other packages? That way we can benchmark everything together.

The runs have already been going for about two hours now. So my suggestion would be to do the other updates on a separate PR.

KaleabTessera:
Layernorm seemed to result in a regression in performance, so closing this for now.

PR #511, opened 2022-06-02
Feature/component abstract classes

mmorris44:
## What?
Create abstract classes for components which we will have variants of in our systems (e.g. feed-forward vs recurrent)
## Why?
Abstract classes define templates for components and enforce using consistent names, since the `name()` method comes from the base class
## How?
Defined abstract components for `executor observe`, `executor selection action`, `Network`, `TrainerStep`, `TrainerDataset`, `parameter server`
## Extra
Tested to ensure changes do not break MAPPO system. ~~Wait for [PR #508](https://github.com/instadeepai/Mava/pull/508) to be merged before reviewing and merging this one~~ -> merged


mava/components/jax/building/datasets.py
```diff
@@ -26,6 +27,38 @@
 Transform = Callable[[reverb.ReplaySample], reverb.ReplaySample]
 
 
+class TrainerDataset(Component):
```
RuanJohn, line 30: Very minor, but isn't it a bit more intuitive to call this `Dataset `instead?

mava/components/jax/executing/action_selection.py
```diff
@@ -34,14 +32,29 @@ class ExecutorSelectActionProcessConfig:
 
 class ExecutorSelectAction(Component):
     @abc.abstractmethod
-    def __init__(self, config: Any) -> None:
+    def __init__(
+        self,
+        config: ExecutorSelectActionProcessConfig = ExecutorSelectActionProcessConfig(),
```
RuanJohn, line 37: Also minor, but isn't it more intuitive to have it called `ExecutorSelectActionConfig` to match the component name?

RuanJohn (changes requested):
Thank you @mmorris44! Looks good, just a few very minor naming comments.

mava/components/jax/building/datasets.py
mmorris44: Agreed, changed

mava/components/jax/executing/action_selection.py
mmorris44: Also agreed, changed

RuanJohn (dismissed):

mava/components/jax/building/datasets.py
mmorris44: Actually, switched it back to `TrainerDataset` to match the hook name. Sorry, that's why I named it that originally

RuanJohn: I see, that is fine by me in that case.

RuanJohn (approved):

PR #512, opened 2022-06-02
Feature/jax systems termination condition

RuanJohn:
## What?
Added termination condition to Jax systems so that experiments terminate automatically. 

## Why?
We were having to stop benchmarks manually on the cluster for all Jax systems. 

## How?
Added a terminator component that terminates the program when a certain condition is met. A change from the tensorflow implementation is that a user now has to pass in a parameter name and value to be used for termination instead of a dictionary containing both. The reason for this is to open up future flexibility where systems could be terminated using rewards or some other monitored value. 

## Extra
[EDIT] Made a terminator component


mmorris44 (changes requested):
I would much rather this be its own component, if possible (as you mentioned in the extras). Being able to define arbitrary termination components seems much nicer to me than embedding this in the parameter server, which doesn't have much to do with termination.

sash-a (changes requested):
Totally agree with @mmorris44 would be nicer as a component or a group of components ie each termination condition is it's own component, but then you might need to add a new callback, not quite sure if a fitting one exists. 

Also fstring are much cleaner for this bit
```
print(
      "StepsLimiter: Max",
      server.store.terminal_key,
      "of",
      server.store.terminal_count,
      "reached, terminating.",
  )
```

mmorris44:
> Totally agree with @mmorris44 would be nicer as a component or a group of components ie each termination condition is it's own component, but then you might need to add a new callback, not quite sure if a fitting one exists.
> ```

Yeah we need a new callback for this. Each _type_ of termination condition should be its own component: e.g. `terminate_on_steps`, which you can give a specific step count to look at, would be one component. A termination component like `terminate_at_performance` could stop when a certain model performance is reached.

There should also be a superclass for termination components.

RuanJohn:
These are some good suggestions @mmorris44 and @sash-a. I'll get on it :call_me_hand:

mava/components/jax/updating/terminators.py
```diff
@@ -14,5 +14,92 @@
...
+            "trainer_walltime",
+            "evaluator_steps",
+            "evaluator_episodes",
+            "executor_episodes",
+            "executor_steps",
+        ]
+        assert self.config.parameter_termination_key in valid_options, (
+            "Please give a valid termination condition. "
+            + f"Current valid conditions are {valid_options}"
+        )
```
mmorris44, line 86: Please rather make this an Exception than an assertion. See [here](https://stackoverflow.com/questions/15046695/why-use-assert-instead-of-exception-throwing) for example. Also, it can be formatted entirely with `f`

mava/components/jax/updating/terminators.py
```diff
@@ -14,5 +14,92 @@
...
+        config: Any = ParameterTerminatorConfig(),
+    ):
+        """_summary_
+
+        Args:
+            config : _description_.
+        """
+        self.config = config
+
+        valid_options = [
```
mmorris44, line 75: Make dictionary instead of list

mava/components/jax/updating/terminators.py
```diff
@@ -14,5 +14,92 @@
...
+@dataclass
+class ParameterTerminatorConfig:
+    parameter_termination_key: str = "executor_steps"
+    parameter_termination_value: int = 5000
+
+
+class ParameterTerminator(Terminator):
+    def __init__(
+        self,
+        config: Any = ParameterTerminatorConfig(),
```
mmorris44, line 66: `config: ParameterTerminatorConfig` -> fix type please

mava/components/jax/updating/terminators.py
```diff
@@ -14,5 +14,92 @@
...
+        self, parameter_sever: SystemParameterServer
+    ) -> None:
+        """_summary_"""
+        if (
+            parameter_sever.store.parameters[self.config.parameter_termination_key]
+            > self.config.parameter_termination_value
+        ):
+            print(
+                f"Max {self.config.parameter_termination_key} of "
+                + f"{self.config.parameter_termination_value} reached, terminating."
```
mmorris44, line 98: Format entire string with `f`

examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo.py
```diff
@@ -101,7 +101,8 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
         num_epochs=15,
         num_executors=1,
         multi_process=True,
-        terminal="current_terminal",
+        parameter_termination_key="executor_steps",
+        parameter_termination_value=1500,
```
mmorris44, line 105: I don't see where these values are actually being set in MAPPO. Seems to be using the default `config` values. Also, personal preference, but I would rather these be passed through as a tuple to denote that they have to be supplied together.

mmorris44 (changes requested):
Thanks @RuanJohn! Some changes needed, but mostly looking good.

Also, checks are failing, which needs fixing.

examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo.py
sash-a: Agreed on the tuple or dict and I think also remove the parameter from the variable name, I know we discussed adding it yesterday, but to an end user `parameter_termination_condition` would be more confusing. So maybe something like `termination_condition`

mava/components/jax/updating/terminators.py
```diff
@@ -14,5 +14,92 @@
...
+        Returns:
+            _description_
+        """
+        return "temination_condition"
+
+
+@dataclass
+class ParameterTerminatorConfig:
+    parameter_termination_key: str = "executor_steps"
+    parameter_termination_value: int = 5000
```
sash-a, line 60: I think this default should be much higher, otherwise by default it'll end too soon and might be hard to debug

mava/components/jax/updating/terminators.py
sash-a: No keys, so rather a set?

mava/components/jax/updating/terminators.py
```diff
@@ -14,5 +14,92 @@
...
+        """_summary_
+
+        Returns:
+            _description_
+        """
+        return "temination_condition"
+
+
+@dataclass
+class ParameterTerminatorConfig:
```
sash-a, line 58: Maybe `ParameterSeverTerminatorConig` instead?

sash-a (changes requested):
Looks great, just some minor changes

mava/components/jax/updating/terminators.py
mmorris44: Sorry, that's what I meant haha. i.e. `{item1, item2, item3}`

mava/components/jax/updating/terminators.py
RuanJohn: Sorted, I just need to keep multiple strings due to mypy PEP errors.

mava/components/jax/updating/terminators.py
RuanJohn: Sorted.

mava/components/jax/updating/terminators.py
RuanJohn: Same issue as mentioned above.

examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo.py
RuanJohn: Changed to take in a `termination_condition` dictionary instead.

mava/components/jax/updating/terminators.py
RuanJohn: Agreed, the default is now `None` so it will run indefinitely.

mmorris44 (approved):
Looks great! Thanks @RuanJohn. One small comment, which you can ignore if you disagree with

sash-a (approved):
Looks good! :rocket: 

Nice extension to this in the future would be to allow for multiple terminal conditions and to terminate when the first one is true

PR #514, opened 2022-06-02
fix: mypy fix for chex imports

RuanJohn:
## What?
Fixed `chex` mypy bug 

## Why?
There were mypy issues when trying to make commits.


mmorris44 (approved):
Nice!

PR #551, opened 2022-06-07
test: unit tests for jax logging components

### LINKED ISSUE #525: [TEST] Jax Loggers
### What do you want to test?
Jax logger components

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

mmorris44:
## What?
Unit tests for jax logging components
## Why?
Component currently not tested. Needs to be robust for jax Mava release
## How?
- Test `Logger` component
- Fixed typo in `mocks.py`
## Extras
Closes #525

### DIFF
--- a/tests/jax/components/building/loggers_test.py
+++ b/tests/jax/components/building/loggers_test.py
@@ -1,10 +1,12 @@
-from typing import Any, Callable, Tuple
+import functools
+from typing import Callable
 
 import pytest
 
 from mava.components.jax.building.loggers import Logger, LoggerConfig
 from mava.core_jax import SystemBuilder
 from mava.systems.jax import Builder
+from mava.utils.loggers import logger_utils
 
 
 class TestLogger(Logger):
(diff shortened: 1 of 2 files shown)


tests/jax/components/building/loggers_test.py
```diff
@@ -0,0 +1,158 @@
...
+    """Pytest fixture for TestLogger.
+
+    Args:
+        test_logger_factory: factory to use in logger config.
+
+    Returns:
+        Default TestLogger.
+    """
+    test_logger = TestLogger(test_logger_factory)
+    return test_logger
```
sash-a, line 78: Just `return TestLogger(test_logger_factory)` here

tests/jax/mocks.py
```diff
@@ -473,7 +473,7 @@ class MockLoggerConfig:
     logger_param_1: int = 2
 
 
-class MockLogerClass:
```
sash-a, line 476: Logers > Loggers :joy:

sash-a (changes requested):
Very minor change, but otherwise looks good :fire:

sash-a (dismissed):
Thanks Matthew :+1:

KaleabTessera (commented):
Thanks @mmorris44 . 

It looks good, but I was wondering if we could pass in the logger as a component to the builder and test it like that? Without manually calling `on_building_executor_logger`. That way we are testing that the builder correctly handles the logger. 

I did something similar with the data server tests - https://github.com/instadeepai/Mava/blob/153b3f6cab8150a9acec5f65ae9366f22207598f/tests/jax/mocks.py#L817.

mmorris44:
> It looks good, but I was wondering if we could pass in the logger as a component to the builder and test it like that? Without manually calling `on_building_executor_logger`. That way we are testing that the builder correctly handles the logger.

I did think about doing this, but decided to rather manually call the hooks. This aligns with our goal of testing components in isolation. What you're describing above feels to me like a builder integration test, not a `Logger` test. Perhaps we should test both?

tests/jax/components/building/loggers_test.py
```diff
@@ -0,0 +1,157 @@
...
+    ):
+        """Create TestLogger
+
+        Args:
+            test_logger_factory: factory to use in the logger config.
+        """
+        logger_config = LoggerConfig()
+        logger_config.logger_factory = test_logger_factory
+        logger_config.logger_config = {
+            "trainer": {"time_stamp": "trainer_config"},
```
RuanJohn, line 27: Very minor, but can we make these `trainer_logger_config`, `executor_logger_config` and `evaluator_logger_config` instead? It makes reading this a littel bit easier.

mmorris44: Agreed, done :)

RuanJohn (approved):
Thank you @mmorris44 :muscle:

PR #552, opened 2022-06-07
Bugfix/fix old tf architectures

KaleabTessera:
## What?
- Fix state based, networked and centralised architectures.
- Remove lp as independent dependency. It is a core dependency in acme, so for now, it is auto-downloaded with acme. 
- Updated some core libraries. 
- Change `env_states` to `s_t` for consistency.
- There is still a weird bug that occurs when running tests with `shared_weights=False` with multiple other tests. Full description of bug: 
  ```
  Running tests with shared_weights=False pass when running indepedently
              (other tests commented out), but fail when run with other tests and not
              enough parallel cores (fail with 2 or less cores, pass with 4 cpu cores). This is likely a race condition,
              hangling process from previous tests or related to network sampling
              (TODO @Dries investigate if you have a chance). Only the test fails,
              the examples run.
  ```
  For now, I don't think it worth fixing these kind of tests since tf will be deprecated soon. 

## Why?
- So that tests pass and we can do final tf release. 
## How?
-
## Extra
-

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ RUN pip install -e .[reverb]
 ENV DISPLAY=:0
 RUN if [ "$record" = "true" ]; then \
     ./bash_scripts/install_record.sh; \
-fi
+    fi
 EXPOSE 6006
 ##########################################################
 
@@ -89,6 +89,10 @@ RUN pip install .[open_spiel]
 ##########################################################
 # MeltingPot Image
 FROM tf-core AS meltingpot
+
+# Melting pot requires python>=3.9
+RUN apt-get install python3.9 -y && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.9 1 && rm -rf /root/.cache && apt-get clean
+
 # Install meltingpot
 RUN apt-get install -y git
 RUN ./bash_scripts/install_meltingpot.sh
@@ -153,6 +157,10 @@ RUN pip install .[open_spiel]
 ##########################################################
 # MeltingPot Image
 FROM jax-core AS meltingpot-jax
+
+# Melting pot requires python>=3.9
+RUN apt-get install python3.9 -y && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.9 1 && rm -rf /root/.cache && apt-get clean
+
 # Install meltingpot
 RUN apt-get install -y git
 RUN ./bash_scripts/install_meltingpot.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,10 +89,6 @@ RUN pip install .[open_spiel]
 ##########################################################
 # MeltingPot Image
 FROM tf-core AS meltingpot
-
-# Melting pot requires python>=3.9
-RUN apt-get install python3.9 -y && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.9 1 && rm -rf /root/.cache && apt-get clean
-
 # Install meltingpot
 RUN apt-get install -y git
 RUN ./bash_scripts/install_meltingpot.sh
@@ -157,10 +153,6 @@ RUN pip install .[open_spiel]
 ##########################################################
 # MeltingPot Image
 FROM jax-core AS meltingpot-jax
-
-# Melting pot requires python>=3.9
-RUN apt-get install python3.9 -y && update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.9 1 && rm -rf /root/.cache && apt-get clean
-
 # Install meltingpot
 RUN apt-get install -y git
 RUN ./bash_scripts/install_meltingpot.sh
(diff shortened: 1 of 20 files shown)


Dockerfile
```diff
@@ -23,13 +23,13 @@ WORKDIR ${folder}
 COPY . /home/app/mava
 # For box2d
 RUN apt-get install swig -y
-## Install core dependencies.
-RUN pip install -e .[reverb,launchpad]
+## Install core dependencies + reverb.
```
mmorris44, line 26: Rather `install reverb dependencies`

README.md
```diff
@@ -263,16 +263,16 @@ docker run --gpus all -it --rm  -v $(pwd):/home/app/mava -w /home/app/mava insta
     * Install core dependencies:
 
     ```bash
-    pip install id-mava[tf,reverb,launchpad]
```
mmorris44, line 266: I'm not quite convinced by removing `launchpad` as a dependency, since we use it directly. Unlikely to break anything though, since like you said, it's a core dependency of `acme`. So I don't feel very strongly about keeping it in

mava/components/tf/architectures/networked.py
```diff
@@ -114,12 +114,13 @@ def _get_critic_specs(
...
             for agent in agents:
-                critic_obs_shape = list(copy.copy(self._embed_specs[agent].shape))
+                net_key = self._agent_net_keys[agent]
+                critic_obs_shape = list(copy.copy(self._embed_specs[net_key].shape))
                 critic_act_shape = list(
                     copy.copy(self._agent_specs[agent].actions.shape)
                 )
                 critic_obs_shape.insert(0, len(self._network_spec[agent]))
-                critic_obs_specs[agent] = tf.TensorSpec(
+                critic_obs_specs[net_key] = tf.TensorSpec(
```
mmorris44, line 123: Is it only the `obs` which need to be indexed by the `net_key`? Do the specs (e.g. see below for `critic_act_specs`) not also need to be?

tests/systems/mad4pg_system_test.py
```diff
@@ -167,7 +170,6 @@ def test_centralised_mad4pg_on_debugging_env(self) -> None:
             checkpoint=False,
             architecture=architectures.CentralisedQValueCritic,
             trainer_fn=mad4pg.MAD4PGCentralisedTrainer,
-            shared_weights=False,
```
mmorris44, line 170: I would rather explicitly set `shared_weights=True`, to make it clear that we need this in our tests

tests/systems/mad4pg_system_test.py
```diff
@@ -221,7 +224,6 @@ def test_state_based_mad4pg_on_debugging_env(self) -> None:
             checkpoint=False,
             trainer_fn=mad4pg.MAD4PGStateBasedTrainer,
             architecture=architectures.StateBasedQValueCritic,
-            shared_weights=False,
```
mmorris44, line 224: Likewise

tests/systems/maddpg_system_test.py
```diff
@@ -266,7 +282,6 @@ def test_state_based_maddpg_on_debugging_env(self) -> None:
             checkpoint=False,
             trainer_fn=maddpg.MADDPGStateBasedTrainer,
             architecture=architectures.StateBasedQValueCritic,
-            shared_weights=False,
```
mmorris44, line 269: Likewise

mmorris44 (dismissed):
Looks good @KaleabTessera. Merge pending benchmarks.

Also, I had a few minor comments which you may choose to address if you'd like.

README.md
KaleabTessera: I think the issue with having a separate dependency while it is a core dependency on acme is managing the version. Acme will have its own required version of lp, if ours doesn't match then the install fails due to a conflict. Becomes tricky to always manage the correct version.

mava/components/tf/architectures/networked.py
KaleabTessera: No only the obs specs. Not 100% sure, only Dries does :laughing: 

But this has been tested.

Dockerfile
KaleabTessera: So I think this installs both core (`.` does core) and reverb.

tests/systems/mad4pg_system_test.py
KaleabTessera: Done.

tests/systems/maddpg_system_test.py
KaleabTessera: Done.

README.md
mmorris44: If you just don't give a specific version in our requirements for `launchpad`, then it will use the `acme` version, I think

KaleabTessera: Yeah that is true, but then I don't see the benefit of having it as a separate dependency. If acme removes it as a core dependency, then we can maybe have it as a separate dependency. What do you think?

mmorris44: Sounds good :)

RuanJohn (approved):
Thank you @KaleabTessera! Benchmarks show no system performance degrading.

PR #559, opened 2022-06-09
Feature/jax systems termination condition unit tests 

RuanJohn:
## What?
Unit tests for `CountConditionTerminator` and `TimeTerminator` components
## Why?
Increasing Mava redesign test coverage.


mypy.ini
```diff
@@ -104,3 +104,6 @@ ignore_missing_imports = True
 
 [mypy-chex.*]
 ignore_missing_imports = True
+
+[mypy-terminators_test_data.*]
+ignore_missing_imports = True
```
sash-a, line 109: My mypy-foo isn't the best so just wanna check that this is just for the terminator test and everywhere else it does still check for missing imports?

tests/jax/updating/terminators_test.py
```diff
@@ -0,0 +1,202 @@
...
+
+@dataclass
+class MockParameterStore:
+    parameters: Optional[Dict[str, Any]] = None
+    stopped: bool = False
+
+
+@dataclass
+class MockParameterServer:
+    store: Optional[MockParameterStore] = None
```
sash-a, line 43: Howcome this isn't in `test/jax/mocks.py`?

tests/jax/mocks.py
```diff
@@ -834,3 +839,15 @@ def design(self) -> Tuple[DesignSpec, Dict]:
...
+@dataclass
+class MockTimeTerminatorConfig(TimeTerminatorConfig):
+    termination_function: Callable = lp.stop
+    run_seconds: float = 1.0
+
+
+@dataclass
+class MockCountConditionTerminatorConfig(CountConditionTerminatorConfig):
+    termination_condition: Optional[Dict[str, Any]] = None
+    termination_function: Callable = lp.stop
```
sash-a, line 853: I could be wrong here, but I don't think these need to be mocks, you should just be able to use the original configs. My understanding is that you only mock out something that has undesired functionality, but these are simply datastores, so they have no functionality and I don't think you need to mock them out

tests/jax/updating/terminators_test_data.py
```diff
@@ -0,0 +1,38 @@
...
+    """Data for count condition terminator"""
+
+    return [
+        {"trainer_steps": 10},
+        {"trainer_walltime": 10},
+        {"evaluator_steps": 10},
+        {"evaluator_episodes": 10},
+        {"executor_episodes": 10},
+        {"executor_steps": 10},
+    ]
```
sash-a, line 28: This is much cleaner than what we were discussing yesterday

tests/jax/updating/terminators_test.py
```diff
@@ -0,0 +1,202 @@
...
+) -> None:
+    """Test if time terminator terminates"""
+
+    test_parameter_server = mock_parameter_server
+
+    def _set_stopped() -> None:
+        test_parameter_server.store.stopped = True
+
+    test_terminator = TimeTerminator(
+        config=MockTimeTerminatorConfig(
```
sash-a, line 166: Again, I think this could be a normal `TerminatorConfig`

tests/jax/updating/terminators_test.py
```diff
@@ -0,0 +1,202 @@
...
+            run_seconds=0.5, termination_function=_set_stopped
+        )
+    )
+
+    test_terminator.on_parameter_server_init(test_parameter_server)
+
+    time.sleep(1)
```
sash-a, line 173: So you really don't want to use `time.sleep` in tests, it's considered bad practice. Rather set `run_second` to 0

tests/jax/updating/terminators_test.py
```diff
@@ -0,0 +1,202 @@
...
+    """Utility function for stepping parameters"""
+
+    parameter_dict[key] += 1
+
+
+@pytest.mark.parametrize("condition", count_condition_terminator_data())
+def test_count_condition_terminator_terminated(
+    condition: Dict, mock_parameter_server: SystemParameterServer
+) -> None:
+    """Test if count condition terminator terminates"""
```
sash-a, line 77: "that" instead of "if"

sash-a (changes requested):
Looks good, couple minor changes and then using the real configs if it is possible. :+1:

mypy.ini
RuanJohn: I think this should be the case, but will wait for someone else to confirm.

tests/jax/mocks.py
RuanJohn: Totally agree. Sorted.

tests/jax/updating/terminators_test_data.py
RuanJohn: :rocket:

tests/jax/updating/terminators_test.py
RuanJohn: Good catch! Thank you. Changed it.

tests/jax/updating/terminators_test.py
```diff
@@ -0,0 +1,202 @@
...
+
+    test_terminator = TimeTerminator(
+        config=MockTimeTerminatorConfig(
+            run_seconds=10, termination_function=_set_stopped
+        )
+    )
+
+    test_terminator.on_parameter_server_init(test_parameter_server)
+
+    time.sleep(1)
```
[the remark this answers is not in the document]
RuanJohn: Chaged.

tests/jax/updating/terminators_test.py
```diff
@@ -0,0 +1,202 @@
...
+    test_terminator.on_parameter_server_run_loop_termination(test_parameter_server)
+
+    assert test_parameter_server.store.stopped is True
+
+
+@pytest.mark.parametrize("condition", count_condition_terminator_data())
+def test_count_condition_terminator_not_terminated(
+    condition: Dict, mock_parameter_server: SystemParameterServer
+) -> None:
+    """Test if count condition terminator does not terminate"""
```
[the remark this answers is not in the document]
RuanJohn: Fixed.

tests/jax/updating/terminators_test.py
RuanJohn: Fixed.

tests/jax/updating/terminators_test.py
sash-a: Happy to leave it here

sash-a (approved):
Thanks Ruan! :rocket:

PR #560, opened 2022-06-09
feature: extra info in config common parameter names exception

mmorris44:
## What?
Better exception when common parameter names in the config
## Why?
Currently can't tell which components are causing the conflicts
## How?
- Store dictionary of parameter name to component name
- Provide the components involved in the conflict when displaying the error message
## Extra
Tested by adding `shared_weights: bool = True` to `NetworksConfig` in `mava/components/jax/building/networks.py`


siddarthsingh1 (dismissed):

KaleabTessera (approved):
Nice! Thanks @mmorris44 !

PR #566, opened 2022-06-10
Tests/jax components/action selection

### LINKED ISSUE #529: [TEST] Jax Action Selection
### What do you want to test?
Jax action selection components

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

OmaymaMahjoub:
## What?
Unit tests for jax action_selection components
## Why?
## How?
Test `FeedForwardExecutorSelectSction` component from `components/jax/executing/action_selection.py`
## Extra
Closes #529


tests/jax/components/executing/action_selection.py
```diff
@@ -0,0 +1,237 @@
...
+def get_action(parm_0, parm_1, parm_2):
+    """Function used in the networks.
+
+    Returns:
+        action_info and policy info
+    """
+    return "action_info_after_get_action", "policy_info_after_get_action"
+
+
+class MockExecutor(Executor):
```
mmorris44, line 69: Keep an eye out for where this can be re-used when testing other executor components. If it can be used in others, then this should go in `mocks.py`

tests/jax/components/executing/action_selection.py
```diff
@@ -0,0 +1,237 @@
...
+    """Test adding config as an attribute
+
+    Args:
+        dummy_config
+    """
+    ff_executor_select_action = FeedforwardExecutorSelectAction(config=dummy_config)
+    assert ff_executor_select_action.config.parm_0 == dummy_config.parm_0
+
+
+def test_constructor_empty_config(dummy_empty_config: type) -> None:
```
mmorris44, line 130: This test is not necessary and can be removed. The above test already checks if the config is loaded properly

tests/jax/components/executing/action_selection.py
```diff
@@ -0,0 +1,237 @@
...
+    mock_executor: Executor, dummy_config: type
+) -> None:
+    """Test on_execution_select_actions.
+
+    Args:
+        dummy_config: config
+        mock_executor: Executor
+    """
+
+    ff_executor_select_action = FeedforwardExecutorSelectAction(dummy_config)
```
mmorris44, line 171: This can be made into a fixture. The exact same line is used elsewhere in the code

tests/jax/components/executing/action_selection.py
```diff
@@ -0,0 +1,237 @@
...
+    ff_executor_select_action.on_execution_select_actions(executor=mock_executor)
+
+    assert (
```
mmorris44, line 174: This long list of assertions can be replaced with a loop. Code would be much neater

tests/jax/components/executing/action_selection.py
```diff
@@ -0,0 +1,237 @@
...
+        mock_executor.store.policies_info["agent_1"]
+        == "policy_info_agent_1_observation_1"
+    )
+    assert (
+        mock_executor.store.policies_info["agent_2"]
+        == "policy_info_agent_2_observation_2"
+    )
+
+
+def test_on_execution_select_actions_param(dummy_config: type) -> None:
```
mmorris44, line 201: This test doesn't feel necessary to me. You're just asserting that it crashes when something without `store` and `select_action` is passed through.

tests/jax/components/executing/action_selection.py
```diff
@@ -0,0 +1,237 @@
...
+        dummy_config: config
+        mock_executor: Executor
+    """
+    ff_executor = FeedforwardExecutorSelectAction(dummy_config)
+    ff_executor.on_execution_select_action_compute(executor=mock_executor)
+    assert mock_executor.store.action_info == "action_info_after_get_action"
+    assert mock_executor.store.policy_info == "policy_info_after_get_action"
+
+
+def test_on_execution_select_action_compute_param(dummy_config: type) -> None:
```
mmorris44, line 228: Likewise, this test does not seem necessary

tests/jax/components/executing/action_selection.py
```diff
@@ -0,0 +1,237 @@
...
+    """Test on_execution_select_action_compute.
+
+    Args:
+        dummy_config: config
+        mock_executor: Executor
+    """
+    ff_executor = FeedforwardExecutorSelectAction(dummy_config)
+    ff_executor.on_execution_select_action_compute(executor=mock_executor)
+    assert mock_executor.store.action_info == "action_info_after_get_action"
+    assert mock_executor.store.policy_info == "policy_info_after_get_action"
```
mmorris44, line 225: This does not test if the correct network was retrieved from the store (see `action_selection.py` line 95), since the output is the same no matter which network is retrieved

mmorris44 (changes requested):
Thanks @OmaymaMahjoub. This is a really great first PR!
A couple of comments from my side:
- file should be called `action_selection_test.py`
- and see other comments on the code

tests/jax/components/executing/action_selection.py
```diff
@@ -0,0 +1,237 @@
...
+            }
+        }
+        key = jax.random.PRNGKey(5)
+        action_info = "action_info_test"
+        policy_info = "policy_info_test"
+
+        store = SimpleNamespace(
+            is_evaluator=None,
+            observations=observations,
+            observation=SimpleNamespace(observation=[0, 0, 1, 0], legal_actions=3),
```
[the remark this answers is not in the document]
OmaymaMahjoub: yeah done replace it with [1]

tests/jax/components/executing/action_selection.py
OmaymaMahjoub: I will try to make changes on the mock so that it works based on our goal

tests/jax/components/executing/action_selection.py
OmaymaMahjoub: Still not sure about this since there is already MockExecutor in mocks.py so probably changing the name and then add it to mocks or make mocks.py just for components

tests/jax/components/executing/action_selection_test.py
```diff
@@ -41,13 +42,13 @@ def dummy_config() -> ExecutorSelectActionConfig:
 
 
 @pytest.fixture
-def dummy_empty_config() -> ExecutorSelectActionConfig:
-    """Dummy empty config attribute for FeedforwardExecutorSelectAction class
+def dummy_ff_executor_select_action():
```
mmorris44, line 45: I would rather call this `test_ff_selection_action`, or just `ff_selection_action`, since you're not constructing a dummy version of it like you are for the executor. You're using the actual component, as you should.

mmorris44 (changes requested):
Amazing PR, thanks @OmaymaMahjoub for addressing all of my comments.
I had one final thought while reading over it again and I am otherwise happy :)

mmorris44 (approved):
Very nice! Thanks @OmaymaMahjoub

tests/jax/components/executing/action_selection_test.py
```diff
@@ -0,0 +1,201 @@
...
+
+# Test initiator
+def test_constructor(dummy_config: ExecutorSelectActionConfig) -> None:
+    """Test adding config as an attribute
+
+    Args:
+        dummy_config
+    """
+    ff_executor_select_action = FeedforwardExecutorSelectAction(config=dummy_config)
+    assert ff_executor_select_action.config.parm_0 == dummy_config.parm_0
```
sash-a, line 142: In future I don't think you need to be this rigorous in testing a constructor if it doesn't do anything (this one just assigns the config), it's fine for this PR, but don't worry about it for others (unless there is some complex logic in the constructor). Reason I say this is because it can lead to confusion for someone looking at the tests as a reference for how to use the system, they could get confused as to what `parm_0` is.

sash-a (approved):
Looks good to me, just note that comment for future PRs :rocket:

tests/jax/components/executing/action_selection_test.py
OmaymaMahjoub: Okay, I will remove it from the next Test!

PR #569, opened 2022-06-10
fix: use actual logger instead of test component

mmorris44:
## What?
In Logger component test, use actual component instead of test component
## Why?
Better practice to test the component directly
## How?
Removed `TestLogger`, use `Logger` instead and set the config in the fixture


RuanJohn (approved):
Looks good and the change makes sense to me. Thank you @mmorris44

siddarthsingh1 (approved):

PR #570, opened 2022-06-10
Bugfix/pre commit exclude mava/lib

mmorris44:
## What?
Update pre-commit exclude and baseline docstring coverage required
## Why?
`mava/lib` contained code which was bringing down the coverage
## How?
- added `mava/lib` to `pyproject.toml` exclude
- Increased baseline coverage required (to just above current coverage)
## Extras
Would be nice, if possible, to have the pre-commit hooks by default ignore everything that is ignored by `git` (@KaleabTessera, is this possible?)


AsadJeewa (approved):
How did you decide on 68.5?

pyproject.toml
```diff
@@ -42,8 +42,8 @@ ignore-module = false
 ignore-nested-functions = false
 ignore-nested-classes = true
 ignore-setters = false
-fail-under = 62.5
-exclude = ["setup.py", "docs","tests","examples"]
+fail-under = 68.5
+exclude = ["setup.py", "docs","tests","examples","mava/lib"]
```
AsadJeewa, line 46: Are you sure that this folder is not local to your PC? It does not exist on my side when I use docker or run locally with venv

siddarthsingh1 (approved):

pyproject.toml
mmorris44: The folder is local for me, but it's one which can be created when building mava, so it's good to include it here to prevent local pre-commit checks from failing

PR #571, opened 2022-06-10
Feature/jax test coverage reporting

RuanJohn:
## What?
Added `.coveragerc` file to ignore all code in `/tf` and `/tests/` folders when calculating test code coverage.
## Why?
Makes it possible to easily track testing code coverage progress and can help with adding the code coverage badge to the repo in future.


mmorris44 (approved):
Looks good! Thanks @RuanJohn.

I think we might possibly have some tf-specific things in other folders that don't have `tf` in the path (e.g. utils), but they're tied up into other code, and I don't think there's much we can do about this for now.

AsadJeewa (approved):

PR #573, opened 2022-06-13
feat: first commit for base system init test

RuanJohn:
## What?
Unit tests for the `SystemInit` component in `mava/components/jax/building/base.py`.
## Why?
Expanding the testing of Mava Jax redesign.

### DIFF
--- a/tests/jax/components/building/base_test.py
+++ b/tests/jax/components/building/base_test.py
@@ -1,0 +1,18 @@
+# python3
+# Copyright 2021 InstaDeep Ltd. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from mava.components.jax.building.base import SystemInit, SystemInitConfig
+
+system = SystemInit(config=SystemInitConfig())


tests/jax/components/building/system_init_test.py
```diff
@@ -0,0 +1,471 @@
...
+    ],
+) -> None:
+    """Tests all system network sampling instantiation ValueErrors."""
+
+    (
+        fixed_network_system,
+        fixed_network_builder,
+    ) = fixed_agent_networks_with_incorrect_sampling_setup_type
+    with pytest.raises(ValueError):
+        assert fixed_network_system.on_building_init(fixed_network_builder) == ValueError  # type: ignore # noqa: E501
```
sash-a, line 457: I don't think you need the assert in here as well as the `with pytest.raises(ValueError):`, ditch the `assert ... == Error` and just use the pytest stuff

sash-a (changes requested):
Other than the minor change, I feel like these are kind of superficial tests, you're checking that what you pass in is what you get out. Can you do 1 extra test where there's an edge case like maybe `unique_net_keys` aren't actually all unique, or `random_networks_net_keys_to_ids` assigns two agents to a single network?

arnupretorius (commented):
Looks good, thanks @RuanJohn 🙂 Just see my one comment on the config.

tests/jax/components/building/system_init_test.py
arnupretorius: Might also help with the type issues.

RuanJohn: Thank you for this catch. Updated :call_me_hand:

mava/components/jax/building/system_init.py
```diff
@@ -169,7 +169,7 @@ def on_building_init(self, builder: SystemBuilder) -> None:
                 "Random sampling system init requires random_agent_networks sampling"
             )
 
-        if builder.store.shared_weights:
+        if self.config.shared_weights:
```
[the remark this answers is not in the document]
mmorris44: Then those configs will all have `shared_weights` as a parameter and the system will complain about duplicate entries. I think we should rather just set it in one component and then access it through the global config for the rest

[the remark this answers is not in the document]
RuanJohn: I think that is a good idea, added a `BaseSystemInitConfig` that all the other configs inherit from :call_me_hand:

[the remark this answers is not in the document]
RuanJohn: All tests in `tests/jax/components/building/system_init_test.py` still pass and I can run `examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo.py` with no issues @mmorris44

sash-a (approved):
My mistake about the extra tests, looks good!

mava/components/jax/building/system_init.py
```diff
@@ -131,11 +135,10 @@ def config_class() -> Optional[Callable]:
 
 
 @dataclass
-class RandomSamplingSystemInitConfig:
+class RandomSamplingSystemInitConfig(BaseSystemInitConfig):
```
mmorris44, line 138: As discussed, this one should please not inherit from `BaseSystemInitConfig`, since it can never have shared weights.

RuanJohn: Updated :+1:

mmorris44 (approved):
Looks good, thanks @RuanJohn!

PR #574, opened 2022-06-13
Test/core parameter server

### LINKED ISSUE #548: [TEST] Jax System Core Parameter Server
### What do you want to test?
Jax system core parameter server (in `mava/systems/jax/`)

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

mmorris44:
## What?
Added tests for the core parameter server. Added utils for testing hook call order.
## Why?
Mava components need to be tested before released. Utils will make future core component tests easier.
## How?
- Added `HookOrderTracking` class for testing whether hooks are called (this is where most of the line changes are coming from, actual tests aren't that long)
- Test hook order calling
- Test config loading and parameter store properly updated
- Test that `step()` sleeps
## Extra
Closes #548. Tests passing.


mmorris44:
> What do you think about using the mock to test this?

From what I can see, it allows one to test that a particular method was called from different places in a specific order. However, I am trying to test if a sequence of methods was called in a specific order, which I cannot find functionality for in the library. I've updated my solution to use list appends instead of hashing though, since it's a much simpler solution.

tests/jax/core_system_components/parameter_server_test.py
```diff
@@ -58,13 +62,38 @@ def design(self) -> Tuple[DesignSpec, Dict]:
...
     """Dummy system with zero components."""
     return TestSystem()
 
 
-def test_parameter_server(
+@pytest.fixture
+def test_parameter_server() -> ParameterServer:
+    """Dummy parameter server with no components"""
+    return TestParameterServer(
+        SimpleNamespace(
```
RuanJohn, line 87: Very minor comment, but I find it easier to read code when you explicitly have the keywords included. For example, 
`return TestParameterServer(
        config = SimpleNamespace(
            config_key="expected_value",
            non_blocking_sleep_seconds=1,
            get_parameters="parameter_list",
        ),
        components=[], 
    )`
This is quite nitpicky so please feel free to ignore.

mmorris44: Agreed. Fixed 👍

RuanJohn (commented):
Everything looks good to me @mmorris44 and I like the idea of checking the order of the hook calls. My only question is why we are not testing the `run` method?

mmorris44:
> Everything looks good to me @mmorris44 and I like the idea of checking the order of the hook calls. My only question is why we are not testing the `run` method?

It only calls the `on_parameter_server_run_start` hook and then enters an infinite loop, so I avoided testing it. Will look into a good way to do so.

KaleabTessera (approved):
Thanks @mmorris44 ! This looks good :open_hands:

PR #576, opened 2022-06-13
Tests/jax components/executing/observing

### LINKED ISSUE #531: [TEST] Jax Observing
### What do you want to test?
Jax observing components

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

OmaymaMahjoub:
## What?
Unit tests for Jax  observing component
## Why?
## How?
Test FeedForwardExecutorObserve component from components/jax/executing/observing.py
## Extra
Add issue [TEST] sort_utils to test functions in mava/utils/sort_utils
Close #531


siddarthsingh1 (approved):
Looks good to me!

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,272 @@
...
+
+    Args:
+        mock_feedforward_executor_observe: FeedForwardExecutorObserve,
+        mock_executor_without_adder: Executor
+    """
+    mock_feedforward_executor_observe.on_execution_observe(
+        executor=mock_executor_without_adder
+    )
+
+    assert not mock_executor_without_adder.store.adder
```
mmorris44, line 216: Likewise to above, this assertion also tests nothing about `on_execution_observe`

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,272 @@
...
+
+    for agent in mock_executor.store.policies_info.keys():
+        assert mock_executor.store.next_extras["policy_info"][
+            agent
+        ] == "policy_info_" + str(agent)
+    assert (
+        mock_executor.store.next_extras["network_int_keys"]
+        == mock_executor.store.network_int_keys_extras
+    )
+    assert mock_executor.store.adder.parm == "after_add"
```
mmorris44, line 239: It would be nice to change the `mock_adder` methods so that you can see that the correct arguments are being passed through to the adder, instead of just checking that the method was called.

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,272 @@
...
+
+    Args:
+        mock_feedforward_executor_observe: FeedForwardExecutorObserve,
+        mock_executor_without_parameter_client: Executor
+    """
+    mock_feedforward_executor_observe.on_execution_update(
+        executor=mock_executor_without_parameter_client
+    )
+
+    assert not mock_executor_without_parameter_client.store.executor_parameter_client
```
mmorris44, line 257: Again, not the correct assertion. You want to assert that `get_async()` was not called.

mmorris44 (changes requested):
Thanks for the PR @OmaymaMahjoub. Please see my comments.

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,272 @@
...
+
+    Args:
+        mock_feedforward_executor_observe: FeedForwardExecutorObserve,
+        mock_executor_without_adder: Executor
+    """
+    mock_feedforward_executor_observe.on_execution_observe_first(
+        executor=mock_executor_without_adder
+    )
+
+    assert not mock_executor_without_adder.store.adder
```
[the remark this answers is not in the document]
OmaymaMahjoub: assert not hasattr(mock_executor_without_adder.store,"extras") I think this will solve the problem, what do you think?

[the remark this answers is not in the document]
mmorris44: It won't solve it, I think, because `extras` is already set in the store by the executor. What is set by the method you're trying to test is
```
executor.store.extras[
            "network_int_keys"
        ] = executor.store.network_int_keys_extras
```
You need to ensure that this code is not run if you do not provide an adder.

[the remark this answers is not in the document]
OmaymaMahjoub: mock_executor_without_adder is a function that return a real exacutor not a mock Executor(store=store) so shouldn't just verify that the store has no betwork_int_keys_extra is enough?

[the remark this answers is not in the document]
RuanJohn: I think an easy way to test this is to assert that `executor.store` is the same before and after the hook is called.

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,269 @@
...
+        # agent_net_keys
+        agent_net_keys = {
+            "agent_0": "network_agent_0",
+            "agent_1": "network_agent_1",
+            "agent_2": "network_agent_2",
+        }
+        # network_int_keys_extras
+        network_int_keys_extras = None
+        # network_sampling_setup
+        network_sampling_setup = [
```
RuanJohn, line 88: I think it is better in testing to have to values of variables set explicitly as much as possible instead of having them computed. That way we know exactly what we are working with. In this case this variable can then be set as `network_sampling_setup = [["network_agent_0", "network_agent_1", "network_agent_2"]]`

tests/jax/components/executing/observing_test.py
RuanJohn: I agree with this idea.

tests/jax/components/executing/observing_test.py
RuanJohn: My suggestion here is similar to above.

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,286 @@
...
+            ["network_agent_0"],
+            ["network_agent_1"],
+            ["network_agent_2"],
+        ]
```
RuanJohn, line 94: Just a heads up here, agent networks will now be randomly sampled. Is this intended?

OmaymaMahjoub: Yeah it is if we use    ` [ ["network_agent_0","network_agent_1","network_agent_2"]] `there won't be shuffle when we use the `sample_new_agent_keys` function it will return same network however if we use the above we will get random values between 0 and 2. Is this right or might I be mistaken?

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,286 @@
...
+        # net_keys_to_ids
+        all_samples = []
+        for sample in network_sampling_setup:
+            all_samples.extend(sample)
+        unique_net_keys = list(sort_str_num(list(set(all_samples))))
+        net_keys_to_ids = {net_key: i for i, net_key in enumerate(unique_net_keys)}
```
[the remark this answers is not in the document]
OmaymaMahjoub: okay I will thanks!

tests/jax/components/executing/observing_test.py
RuanJohn: You are correct, I just wanted to double check that you wanted to be doing that :slightly_smiling_face:

tests/jax/components/executing/observing_test.py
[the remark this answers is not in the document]
RuanJohn: Thank you :+1:

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,340 @@
...
+from mava.types import OLT
+from mava.utils.sort_utils import sort_str_num
+
+
+class MockAdder:
+    def __init__(self) -> None:
+        pass
+
+    def add_first(
+        self, timestep: TimeStep, extras: Dict[str, NestedArray] = ...  # type: ignore
```
RuanJohn, line 41: is the `= ...` needed? Removing it might sort out the type issue?

OmaymaMahjoub: I wanted to let the same format of the original method that's why. Should I keep it or not?

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,340 @@
...
+    """
+    feedforward_executor_observe.on_execution_observe_first(executor=mock_executor)
+
+    for agent, net in mock_executor.store.agent_net_keys.items():
+        assert type(agent) == str
+        assert type(net) == str
+        assert agent in ["agent_0", "agent_1", "agent_2"]
+        assert net in ["network_agent_0", "network_agent_1", "network_agent_2"]
+
+    agents = sort_str_num(list(mock_executor.store.agent_net_keys.keys()))
```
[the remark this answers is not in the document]
OmaymaMahjoub: yeah sure it's actually 0, 1 and 2 I will change it!

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,340 @@
...
+
+"""Tests for FeedforwardExecutorObserve class for Jax-based Mava systems"""
+
+
+from types import SimpleNamespace
+from typing import Dict, Tuple
+
+import jax.numpy as jnp
+import pytest
+from acme.types import NestedArray
```
arnupretorius, line 24: I think we should make an issue to remove our dependency on acme's types (I know it is not just here but in a few others places) since we have our own. Also, we should make new type definitions specific for Jax.

tests/jax/components/executing/observing_test.py
OmaymaMahjoub: I removed "=..." to get mypy functional

RuanJohn: I think then keep it in.

arnupretorius: I think the base adders just pass in an empty dict for the extras. Perhaps you can do that?

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,337 @@
...
+            step_type=StepType.FIRST,
+            reward=0.0,
+            discount=1.0,
+            observation=OLT(
+                observation=[0.1, 0.3, 0.7], legal_actions=[1], terminal=[0.0]
+            ),
+        )
+        # extras
+        extras: Dict[str, Any] = {}
+        # Aadder
```
arnupretorius, line 106: adder

tests/jax/components/executing/observing_test.py
OmaymaMahjoub: I removed the test of get_async since we agree on that in one of the previous standups

tests/jax/components/executing/observing_test.py
```diff
@@ -0,0 +1,333 @@
...
+        assert type(agent) == str
+        assert agent in ["agent_0", "agent_1", "agent_2"]
+        assert value in mock_executor.store.net_keys_to_ids.values()
+
+    assert (
+        mock_executor.store.extras["network_int_keys"]
+        == mock_executor.store.network_int_keys_extras
+    )
+
+    # test mock_executor.store.adder.add_first()
```
arnupretorius, line 223: remove?

arnupretorius (approved):
Thanks @OmaymaMahjoub 🙂  Just left a few minor comments.

tests/jax/components/executing/observing_test.py
OmaymaMahjoub: I can remove it from this test since I just wanted to portray how the original function look like and change the type to Any in the functions `add` and `add_first`?

OmaymaMahjoub: I removed it and change the types of the parameters to Any

RuanJohn (approved):
Looks good. Thank you @OmaymaMahjoub!

mmorris44 (approved):
Looks good, thanks @OmaymaMahjoub!

RuanJohn (approved):

PR #577, opened 2022-06-13
Test/components/losses

siddarthsingh1:
## What?
Unit testing for losses component
## Why?
Increasing test coverage for Mava redesign
## Extra
~~Testing for the consistency of the output of the loss is not completed.~~

### DIFF
--- a/tests/jax/components/training/losses_test.py
+++ b/tests/jax/components/training/losses_test.py
@@ -3,6 +3,7 @@ from typing import Callable, Dict, Tuple
 
 import jax
 import jax.numpy as jnp
+import numpy as np
 import pytest
 
 from mava.components.jax.training.losses import (
@@ -17,18 +18,18 @@ class MockNet:
 
     @staticmethod
     def apply(
-        parameters: Dict[str, jnp.ndarray], observation: jnp.array
-    ) -> Tuple[jnp.ndarray, jnp.ndarray]:
+        parameters: Dict[str, np.ndarray], observation: np.array
+    ) -> Tuple[np.ndarray, np.ndarray]:
         """Mock function to apply the network to training data"""
         return observation, parameters["mlp/~/linear_0"]["w"]
 
 
-def log_prob(distribution_params: jnp.array, actions: jnp.array) -> jnp.ndarray:
+def log_prob(distribution_params: np.array, actions: np.array) -> np.ndarray:
     """Mock function to return fixed log probs"""
     return distribution_params + actions
 
 
-def entropy(distribution_params: jnp.array) -> jnp.ndarray:
+def entropy(distribution_params: np.array) -> np.ndarray:
     """Mock function to return fixed entropy"""
 
     return distribution_params
@@ -38,7 +39,7 @@ def entropy(distribution_params: jnp.array) -> jnp.ndarray:
 def mock_trainer() -> Trainer:
     """Creates mock trainer fixture"""
 
-    observation = SimpleNamespace(observation=jnp.array([0.5, 0.5, 0.7, 0.2]))
+    observation = SimpleNamespace(observation=np.array([0.5, 0.5, 0.7, 0.2]))
     observations = {
         "agent_0": observation,
         "agent_1": observation,
--- a/tests/jax/components/training/losses_test.py
+++ b/tests/jax/components/training/losses_test.py
@@ -23,7 +23,7 @@ class MockNet:
 
 
 def log_prob(distribution_params: jnp.array, actions: jnp.array) -> jnp.ndarray:
-    """Mock function to return fixed log probs"""cd 
+    """Mock function to return fixed log probs"""
     return distribution_params + actions
 
 
@@ -37,7 +37,7 @@ def entropy(distribution_params: jnp.array) -> jnp.ndarray:
 def mock_trainer() -> Trainer:
     """Creates mock trainer fixture"""
 
-    observation = SimpleNamespace(observation= jnp.array([0.5, 0.5, 0.7, 0.2]))
+    observation = SimpleNamespace(observation=jnp.array([0.5, 0.5, 0.7, 0.2]))
     observations = {
         "agent_0": observation,
         "agent_1": observation,
@@ -182,9 +182,11 @@ def test_mapg_loss(
     low_agent_0_loss = low_loss_info["agent_0"]
     low_loss_value = low_agent_0_loss["loss_value"]
     low_loss_policy = low_agent_0_loss["loss_policy"]
+    loss_total_low = low_agent_0_loss["loss_total"]
 
     assert loss_entropy == -0.47500002
     assert loss_value == 6.4075003
 
     assert low_loss_value < loss_value
     assert low_loss_policy < loss_policy
+    assert loss_total_low < loss_total
--- a/tests/jax/components/training/losses_test.py
+++ b/tests/jax/components/training/losses_test.py
@@ -3,7 +3,6 @@ from typing import Callable, Dict, Tuple
 
 import jax
 import jax.numpy as jnp
-import numpy as np
 import pytest
 
 from mava.components.jax.training.losses import (
@@ -18,18 +17,18 @@ class MockNet:
 
     @staticmethod
     def apply(
-        parameters: Dict[str, np.ndarray], observation: np.array
-    ) -> Tuple[np.ndarray, np.ndarray]:
+        parameters: Dict[str, jnp.ndarray], observation: jnp.ndarray
+    ) -> Tuple[jnp.ndarray, jnp.ndarray]:
         """Mock function to apply the network to training data"""
         return observation, parameters["mlp/~/linear_0"]["w"]
 
 
-def log_prob(distribution_params: np.array, actions: np.array) -> np.ndarray:
+def log_prob(distribution_params: jnp.ndarray, actions: jnp.ndarray) -> jnp.ndarray:
     """Mock function to return fixed log probs"""
     return distribution_params + actions
 
 
-def entropy(distribution_params: np.array) -> np.ndarray:
+def entropy(distribution_params: jnp.ndarray) -> jnp.ndarray:
     """Mock function to return fixed entropy"""
 
     return distribution_params
@@ -39,7 +38,7 @@ def entropy(distribution_params: np.array) -> np.ndarray:
 def mock_trainer() -> Trainer:
     """Creates mock trainer fixture"""
 
-    observation = SimpleNamespace(observation=np.array([0.5, 0.5, 0.7, 0.2]))
+    observation = SimpleNamespace(observation=jnp.array([0.5, 0.5, 0.7, 0.2]))
     observations = {
         "agent_0": observation,
         "agent_1": observation,
--- a/tests/jax/components/training/losses_test.py
+++ b/tests/jax/components/training/losses_test.py
@@ -1,5 +1,6 @@
 from types import SimpleNamespace
-from typing import Callable, List, Tuple
+#from typing import Callable, List, Tuple
+from typing import Any, Callable, Dict, Optional, Tuple
 
 import jax
 import jax.numpy as jnp
@@ -13,7 +14,7 @@ def mock_trainer() -> Trainer:
     """Creates mock trainer fixture"""
 
     mock_trainer = Trainer(config=SimpleNamespace())
-    mock_trainer.store.gae_fn = None
+    mock_trainer.store.grad_fn = None
 
     return mock_trainer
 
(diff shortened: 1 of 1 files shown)


mmorris44 (changes requested):
Thanks @siddarthsingh1! Please see my comments.

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,210 @@
+from types import SimpleNamespace
+
+# from typing import Callable, List, Tuple
```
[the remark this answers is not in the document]
siddarthsingh1: Updated

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,210 @@
...
+import pytest
+
+from mava.components.jax.training.losses import (
+    MAPGTrustRegionClippingLossConfig,
+    MAPGWithTrustRegionClippingLoss,
+)
+from mava.systems.jax.trainer import Trainer
+
+
+class mock_net:
```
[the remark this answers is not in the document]
siddarthsingh1: Updated

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,210 @@
...
+) -> None:
+    """Test whether mapg loss output is as expected"""
+    mapg_trust_region_clipping_loss.on_training_loss_fns(trainer=mock_trainer)
+    grad_fn = mock_trainer.store.grad_fn
+
+    actions = mock_actions()
+    behaviour_log_probs = mock_behaviour_log_probs()
+    target_values = mock_target_values()
+    advantages = mock_advantages()
+    behavior_values = mock_behavior_values()
```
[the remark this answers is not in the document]
siddarthsingh1: Updated

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,210 @@
...
+) -> None:
+    """Test whether mapg function is successfully created"""
+
+    mapg_trust_region_clipping_loss.on_training_loss_fns(trainer=mock_trainer)
+    assert hasattr(mock_trainer.store, "grad_fn")
+    assert isinstance(mock_trainer.store.grad_fn, Callable)  # type:ignore
+
+
+def test_mapg_config_creation() -> None:
+    """Test whether mapg loss config variables are of correct type"""
```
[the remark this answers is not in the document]
siddarthsingh1: Removed

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,210 @@
...
+
+def entropy(distribution_params: jnp.array) -> jnp.ndarray:
+    """Mock function to return fixed entropy"""
+
+    entropy = jnp.array([-1.7, -1.7, -1.7, -1.7, -1.7])
+
+    return entropy
+
+
+def mock_actions() -> Dict[str, jnp.ndarray]:
```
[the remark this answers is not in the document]
siddarthsingh1: Updated

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,210 @@
...
+            [-1.7, -1.7, -1.7, -1.7, -1.7]
+        )
+
+
+def log_prob(distribution_params: jnp.array, actions: jnp.array) -> jnp.ndarray:
+    """Mock function to return fixed log probs"""
+
+    log_probs = jnp.array([-1.7, -1.7, -1.7, -1.7, -1.7])
+
+    return log_probs
```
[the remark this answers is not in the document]
siddarthsingh1: Updated

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,210 @@
...
+
+class mock_net:
+    """Creates a mock network for loss function"""
+
+    def apply(
+        observation: jnp.array, rng_key: jnp.array
+    ) -> Tuple[jnp.ndarray, jnp.ndarray]:
+        """Mock function to apply the network to training data"""
+        return jnp.array([-1.7, -1.7, -1.7, -1.7, -1.7]), jnp.array(
+            [-1.7, -1.7, -1.7, -1.7, -1.7]
```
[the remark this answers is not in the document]
siddarthsingh1: It's changed to use the variables to create the return.

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,210 @@
...
+    agent_0_loss = loss_info["agent_0"]
+    loss_entropy = agent_0_loss["loss_entropy"]
+    loss_policy = agent_0_loss["loss_policy"]
+    loss_total = agent_0_loss["loss_total"]
+    loss_value = agent_0_loss["loss_value"]
+
+    assert loss_entropy == 1.7
+    assert loss_policy == -2.0
+    assert loss_total == 9.062
+    assert loss_value == 22.09
```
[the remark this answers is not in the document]
siddarthsingh1: For the rlax reliant values it checks if a different set of variables returns a lower value or not.

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,191 @@
...
+import pytest
+
+from mava.components.jax.training.losses import (
+    MAPGTrustRegionClippingLossConfig,
+    MAPGWithTrustRegionClippingLoss,
+)
+from mava.systems.jax.trainer import Trainer
+
+
+class Mock_net:
```
mmorris44, line 15: `MockNet`

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,191 @@
...
+    def apply(
+        parameters: jnp.array, observation: jnp.array
+    ) -> Tuple[jnp.ndarray, jnp.ndarray]:
+        """Mock function to apply the network to training data"""
+        return jnp.array(observation), jnp.array(observation)
+
+
+def log_prob(distribution_params: jnp.array, actions: jnp.array) -> jnp.ndarray:
+    """Mock function to return fixed log probs"""
+    return distribution_params + actions
```
mmorris44, line 27: Can we not just concatenate here as well?

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class Mock_net:
+    """Creates a mock network for loss function"""
+
+    def apply(
+        parameters: jnp.array, observation: jnp.array
+    ) -> Tuple[jnp.ndarray, jnp.ndarray]:
+        """Mock function to apply the network to training data"""
+        return jnp.array(observation), jnp.array(observation)
```
mmorris44, line 22: They're already jnp arrays

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,191 @@
...
+    low_agent_0_loss = low_loss_info["agent_0"]
+    low_loss_value = low_agent_0_loss["loss_value"]
+    low_loss_policy = low_agent_0_loss["loss_policy"]
+
+    assert loss_entropy == -0.47500002
+    assert loss_value == 6.4075003
+    assert loss_total == (loss_entropy * 0.01 + loss_policy + loss_value * 0.5)
+
+    assert low_loss_value < loss_value
+    assert low_loss_policy < loss_policy
```
mmorris44, line 191: This is still confusing me a little bit, since we're asserting for a value of `loss_total` here?

mmorris44 (changes requested):
Thanks @siddarthsingh1 for the changes since last review. Just a couple more things to fix that I found.

tests/jax/components/training/losses_test.py
siddarthsingh1: Changed

tests/jax/components/training/losses_test.py
siddarthsingh1: Since they're both arrays I added them together to pretend we're using them to calculate the log_probs. If they're concatenated it would change the output size.

tests/jax/components/training/losses_test.py
siddarthsingh1: There was an error for `observation = SimpleNamespace(observation= jnp.array([0.5, 0.5, 0.7, 0.2]))`  where I didn't set the observation to be a jnp array. It's been fixed and the extra `jnp.array(...)`s have been removed.

tests/jax/components/training/losses_test.py
siddarthsingh1: I was checking if the `loss_total` was being added together correctly based on the default coefficients in the config. I've removed it if you think it's necessary?

siddarthsingh1: I've changed it to check if the `low_loss_total' is less than the `loss_total` value.

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class MockNet:
+    """Creates a mock network for loss function"""
+
+    def apply(
+        parameters: jnp.array, observation: jnp.array
+    ) -> Tuple[jnp.ndarray, jnp.ndarray]:
+        """Mock function to apply the network to training data"""
+        return observation, observation
```
mmorris44, line 22: `return parameters, observation` surely?

mmorris44 (changes requested):
One final thing that I caught, sorry. Also the unresolved comment on `rlax` / `loss_total`

tests/jax/components/training/losses_test.py
siddarthsingh1: It gets the `policy_loss` value from rlax but it's just a summation of the other 3 losses. So if the other 3 losses are working as intended then the check is just too see that it's being generated as intended.

tests/jax/components/training/losses_test.py
mmorris44: But can't you just assert for the value of the parameters in the output?

tests/jax/components/training/losses_test.py
mmorris44: @siddarthsingh1 this still requires attention. You can add back in the direct assertion for loss values.

siddarthsingh1: Added it back

tests/jax/components/training/losses_test.py
siddarthsingh1: It's actually supposed to be a dictionary. It's been changed to use the parameters value.Thanks for catching that!

tests/jax/components/training/losses_test.py
mmorris44: Awesome, if we're doing this then, do we even still need the `low_loss_total` etc. assertions?

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,194 @@
...
+        agent: jnp.array([1.0, 1.0, 1.0, 1.0])
+        for agent in {"agent_0", "agent_1", "agent_2"}
+    }
+
+    low_target_values = {
+        agent: jnp.array([1.0, 1.0, 1.0, 1.0])
+        for agent in {"agent_0", "agent_1", "agent_2"}
+    }
+    low_advantages = {
+        agent: jnp.array([3.0, 3.0, 3.0, 3.0])
```
RuanJohn, line 153: Maybe set these to be all 1.0 to make them lower than `advantages`

siddarthsingh1: The naming is confusing I think? They're used to generate the lower losses so they need to be higher to create a lower `loss_total`.

tests/jax/components/training/losses_test.py
mmorris44: @siddarthsingh1 what are your thoughts? Seems to me that if we can directly assert for values like above, then we don't need this high / low loss value setup

mmorris44: Thanks for clarifying this morning. I'm happy with how it is now, makes sense!

mmorris44 (dismissed):
Looks good! Thanks @siddarthsingh1

tests/jax/components/training/losses_test.py
RuanJohn: I see. Thank you for clarifying.

tests/jax/components/training/losses_test.py
```diff
@@ -0,0 +1,190 @@
...
+    observations = {
+        "agent_0": observation,
+        "agent_1": observation,
+        "agent_2": observation,
+    }
+
+    agent_net_keys = {
+        "agent_0": "network_agent_0",
+        "agent_1": "network_agent_1",
+        "agent_2": "network_agent_2",
```
RuanJohn, line 51: Please rename the 'agent_net_keys' values here to all be `network_agent` to reflect the shared network defined below.

siddarthsingh1: Updated

RuanJohn (approved):
Thanks @siddarthsingh1!

mmorris44 (approved):

PR #581, opened 2022-06-14
Feature/add sampler remover components

EdanToledo:
## What?
Added Sampler and remover components the same way rate limiter components are done.
## Why?
It fits in better with the JAX redesign and allows the system to know what the priority exponent is. Passing in partial reverb sampler and remover functions is not user friendly.
## How?
Sampler and remover components simply add the function to the builder store and the off policy data server calls it there.


mava/components/jax/building/data_server.py
```diff
@@ -137,10 +280,22 @@ def table(
...
+            warnings.warn(
+                "Sampler has not been expicitly chosen - defaults to uniform sampler."
+            )
+
+        if not hasattr(builder.store, "remover_fn"):
+            builder.store.remover_fn = reverb.selectors.Fifo
+            warnings.warn(
+                "Remover has not been expicitly chosen - defaults to FIFO remover."
+            )
+
```
mmorris44, line 294: Not sure how I feel about defaulting to something here. If you look at the `rate_limiter_fn`, it doesn't have a default. It requires that a `RateLimiter` component be provided.

I think there should rather be checks here which throw an exception, since the user defined a data server component without providing a remover or a sampler component.
Whatever conclusion we go with, the same should be applied to the `rate_limiter`

mava/components/jax/building/data_server.py
```diff
@@ -99,12 +99,155 @@ def name() -> str:
...
+        Returns:
+            config class/dataclass for component.
+        """
+        return SamplerConfig
+
+
+class UniformSampler(Sampler):
+    def on_building_data_server_start(self, builder: SystemBuilder) -> None:
+        """_summary_"""
```
mmorris44, line 144: Everything needs proper docstrings please.

mava/components/jax/building/data_server.py
```diff
@@ -99,12 +99,155 @@ def name() -> str:
         return "data_server"
 
 
+@dataclass
+class SamplerConfig:
+    pass
+
+
+class Sampler(Component):
```
mmorris44, line 107: I'm inclined to say that these two new components and the `rate_limiters` should live in a file `components/jax/building/reverb.py`, or something along those lines.

mmorris44 (changes requested):
Thanks @EdanToledo. Some good, promising changes here! Just a few comments about file structure, docstrings, and the guardrails.

mava/components/jax/building/data_server.py
EdanToledo: I have removed the warnings - I agree an exception should be raised - not sure where we should do this. Maybe in the same place where my warnings were.

mava/components/jax/building/data_server.py
EdanToledo: The rate limiters have no docstrings so I'm not exactly sure the format we are expected to follow.

mava/components/jax/building/data_server.py
EdanToledo: Done - moved it into a file called reverb.py

mava/components/jax/building/data_server.py
mmorris44: This is the format used elsewhere:

"""Short description of what method does.

    Further explanation of method, if needed.

    Args:
        arg1: Explanation about what it is.

    Returns:
        Return type here.
"""

EdanToledo: No other components have docs and i probably will write them incorrectly and they will have to change anyway. I honestly think docs should be written by the mava team.

mava/components/jax/building/data_server.py
EdanToledo: exceptions have been added.

mava/components/jax/building/reverb.py
```diff
@@ -0,0 +1,243 @@
...
+        """_summary_
+
+        Args:
+            config : _description_.
+        """
+        self.config = config
+
+    def on_building_data_server_start(self, builder: SystemBuilder) -> None:
+        """_summary_"""
+        builder.store.priority_exponent = self.config.priority_exponent
```
mmorris44, line 169: This can now be handled through `store.global_config` instead. Please see elsewhere in the code.

mmorris44 (dismissed):
Just one further change after another PR was merged, thanks @EdanToledo

mmorris44 (commented):
Can't approve my own PR

RuanJohn (dismissed):
I think this is a really intuitive implementation!

EdanToledo:
I agree!

RuanJohn (approved):

KaleabTessera (dismissed):
Looks great, thanks @EdanToledo @mmorris44

RuanJohn (approved):

KaleabTessera (approved):

PR #584, opened 2022-06-15
Fix/fix types jax systems + dev merge

KaleabTessera:
## What?
- Merge develop into jax-systems.
- Add mypy checks for examples and tests.
## Why?
- Pick up errors in examples and tests.
## How?
-
## Extra
Most of the changes in this PR are from the merging dev into jax-systems. Please focus on https://github.com/instadeepai/Mava/commit/834a988b2c3b57b82371bcf8fbdc4818eae69c20 and https://github.com/instadeepai/Mava/commit/02e0833706d597182c29ac5401297c410282ad4f.


README.md
```diff
@@ -263,16 +263,16 @@ docker run --gpus all -it --rm  -v $(pwd):/home/app/mava -w /home/app/mava insta
...
 
     ```bash
-    pip install id-mava[tf,reverb,launchpad]
+    pip install id-mava[tf,reverb]
     ```
 
     * Or for the latest version of mava from source (**you can do this for all pip install commands below for the latest depedencies**):
 
     ```bash
-    pip install git+https://github.com/instadeepai/Mava#egg=id-mava[reverb,tf,launchpad]
```
RuanJohn, line 272: Just a minor quesiton, does launchpad get installed somewhere else now? Or why is it removed?

tests/tf/systems/maddpg_system_test.py
```diff
@@ -185,6 +190,16 @@ def test_centralised_maddpg_on_debugging_env(self) -> None:
         for _ in range(2):
             trainer.step()
 
+    @pytest.mark.skip(
+        reason="""
+            Running tests with shared_weights=False pass when running indepedently
+            (other tests commented out), but fail when run with other tests and not
+            enough parallel cores (fail with 2 or less cores, pass with 4 cpu cores).
+            This is likely a race condition, hangling process from previous tests
```
RuanJohn, line 198: Very minor, but `hanging` instead of `hangling`

README.md
AsadJeewa: I thought that it would have been installed by one of the other dependencies i.e. tf or reverb but does not seem to be the case

AsadJeewa (approved):
As you said, there is a lot going on here but all looks good to me, particularly the 2 commits you highlighted.

README.md
AsadJeewa: we are getting launchpad via the acme install now

mmorris44 (commented):
It's in general a bit concerning to be that so many `# type: ignore`'s were added. Some of the places I saw them added really feel like they shouldn't be required.

README.md
RuanJohn: Thank you @AsadJeewa :)

tests/tf/systems/maddpg_system_test.py
KaleabTessera: Done.

KaleabTessera:
> It's in general a bit concerning to be that so many `# type: ignore`'s were added. Some of the places I saw them added really feel like they shouldn't be required.

@mmorris44 Agreed. Unfortunely, most of it is unavoidable due the problem mentioned here https://github.com/instadeepai/Mava/blob/834a988b2c3b57b82371bcf8fbdc4818eae69c20/tests/jax/updating/terminators_test.py#L73.

KaleabTessera:
@OmaymaMahjoub Is looking at the new mypy issues after I merged.

RuanJohn (dismissed):
All good from my side. Thank you @KaleabTessera :call_me_hand:

bash_scripts/install_meltingpot.sh
```diff
@@ -20,4 +20,7 @@ pip install https://github.com/deepmind/lab2d/releases/download/release_candidat
 mkdir ../packages
 cd ../packages
 git clone -b main https://github.com/deepmind/meltingpot
-pip install meltingpot/
+# Temp - Pin version before python 3.9 requirement.
```
arnupretorius, line 23: Do you think it will only be temporary and that meltingpot will support older versions in the future? I also wonder for how long we should support all these? I.e. how common is it that other libraries do not support python 3.7 for example meaning that we would have to either not use them, or selectively support them based on the version of python you're using Mava with.

arnupretorius (approved):
Looks good to me, thanks @KaleabTessera 👍

bash_scripts/install_meltingpot.sh
KaleabTessera: I think meltingpot has said goodbye to older versions of python, not 100% sure why. For us, I think we should support them as long as they don't prove to be an obstacle (like python 3.6 was with that weird extra dependency we had to manually down).

PR #585, opened 2022-06-15
feat: split SystemInit component into three different components.

RuanJohn:
## What?
Renamed `mava/components/jax/building/base.py` to `mava/components/jax/building/system_init.py` to better explain what the file is resposible for. 
Split the `SystemInit` component into three components: `FixedNetworkSystemInit`, `RandomSamplingSystemInit` and `CustomSamplingSystemInit`. 
Access parameters in the config directly from the config and avoid adding them to any stores. 
## Why?
Randomly network sampling was not needed as a default in our systems and it made more sense to have it as an optional component. Separating out components also makes the code easier to read. 

## Extra
Tested by running `examples/jax/debugging/simple_spread/feedforward/decetralised/run_mappo.py`. 
All three components contain the same checks after their main logic, I opted to keep it like this to allow for following the logic, but this could be stripped out into a function that can get called inside the `on_building_init` hook. 
Also happy to change any component names to make them more descriptive.


mava/components/jax/building/system_init.py
```diff
@@ -0,0 +1,247 @@
+from dataclasses import dataclass, field
+from typing import Callable, List, Optional, Union
+
+from mava.components.jax import Component
+from mava.core_jax import SystemBuilder
+from mava.utils import enums
+from mava.utils.sort_utils import sample_new_agent_keys, sort_str_num
+
```
AsadJeewa, line 8: I like the idea of renaming this file as a billion bases is confusing

mava/components/jax/building/system_init.py
```diff
@@ -0,0 +1,247 @@
...
+        Returns:
+            config class/dataclass for component.
+        """
+        return FixedNetworkSystemInitConfig
+
+
+@dataclass
+class RandomSamplingSystemInitConfig:
+    network_sampling_setup: enums.NetworkSampler = (
+        enums.NetworkSampler.random_agent_networks
```
AsadJeewa, line 96: As discussed, these enums seem redundant now

AsadJeewa (approved):
Thanks Ruan. All looks good on my side. Cleaner solution than Enums

mava/components/jax/building/system_init.py
```diff
@@ -0,0 +1,247 @@
...
+        for sample in builder.store.network_sampling_setup:
+            all_samples.extend(sample)
+        builder.store.unique_net_keys = list(sort_str_num(list(set(all_samples))))
+
+        # Create mapping from ints to networks
+        builder.store.net_keys_to_ids = {
+            net_key: i for i, net_key in enumerate(builder.store.unique_net_keys)
+        }
+
+    @staticmethod
```
AsadJeewa, line 78: I think we should create a baseSystemInit for the 3 classes with these shared methods (as done with the adder components)

AsadJeewa (changes requested):
One small change request. Thanks @RuanJohn  :D

AsadJeewa (commented):
Thanks Ruan. All good on my side besides removing the enums

AsadJeewa (dismissed):

mava/components/jax/building/system_init.py
```diff
@@ -0,0 +1,272 @@
...
+        for i in range(1, len(self.config.network_sampling_setup)):
+            assert len(self.config.network_sampling_setup[i]) == sample_length
+
+        # Get all the unique agent network keys
+        all_samples = []
+        for sample in self.config.network_sampling_setup:
+            all_samples.extend(sample)
+        builder.store.unique_net_keys = list(sort_str_num(list(set(all_samples))))
+
+        # Create mapping from ints to networks
```
mmorris44, line 260: From networks to ints

mava/components/jax/building/system_init.py
```diff
@@ -0,0 +1,272 @@
...
+        """_summary_"""
+        return "system_init"
+
+
+@dataclass
+class FixedNetworkSystemInitConfig:
+    shared_weights: bool = True
+    network_sampling_setup: enums.NetworkSampler = (
+        enums.NetworkSampler.fixed_agent_networks
+    )
```
mmorris44, line 55: Seems like these enums are still necessary for now? I will shortly be working on the dependency guardrails. Up to you whether or not you'd like to remove the enum checks, but I'm happy to leave them in for now.

mmorris44 (changes requested):
One minor change to a comment and a question about the enums.

mava/components/jax/building/system_init.py
RuanJohn: Sorted.

mava/components/jax/building/system_init.py
RuanJohn: The enums are still needed for now and need to be added to the builder store due to the `DataServer` component needing to access `builder.store.network_sampling_setup_type` [here](https://github.com/instadeepai/Mava/blob/ddcd87db9ece5aa975d51420440dd63b158e8d62/mava/components/jax/building/data_server.py#L54).

mmorris44: Is that something I should change in my global configs PR then? If this one gets merged first

RuanJohn: Yes, I think that would be good.

mmorris44 (approved):
Looks good, thanks @RuanJohn. I'll try remember to do the config change in my PR

mava/components/jax/building/system_init.py
AsadJeewa: Resolved below

AsadJeewa (approved):
Thanks Ruan

PR #587, opened 2022-06-20
Feature/global config in store

mmorris44:
## What?
Components are now only given local configs. Global config can be accessed through the store.
## Why?
It becomes clear when components are accessing global configs, which other components have provided. This:
- makes the code is more readable
- is easier to understand for newcomers
- stops the linter from complaining about accessing parameters which don't exist
- clearly shows which components are dependent on the presence of others.
## How?
- Updated docstrings to match signatures changes
- Wrote new `config.py` method, `get_local_config()`, to provide only the local config for a component
- In `jax/system.py`, provide only the local config when constructing components
- Access `environment_factory` from `builder.store.global_config` in `ExecutorEnvironmentLoop` to fix error
- Builder gets global config during init and saves it to the store
## Extra
Tested on all existing `jax` unit tests and `run_mappo.py`. Ensured that local configs only contain the parameters specified in the config class.

### DIFF
--- a/mava/components/jax/building/environments.py
+++ b/mava/components/jax/building/environments.py
@@ -93,8 +93,9 @@ class ExecutorEnvironmentLoop(Component):
         Args:
             builder : _description_
         """
-        builder.store.executor_environment = self.config.environment_factory(
-            evaluation=False
+        # Global config set by EnvironmentSpec component
+        builder.store.executor_environment = (
+            builder.store.global_config.environment_factory(evaluation=False)
         )  # type: ignore
 
     @abc.abstractmethod
--- a/mava/systems/jax/config.py
+++ b/mava/systems/jax/config.py
@@ -17,8 +17,9 @@
 
 from dataclasses import is_dataclass
 from types import SimpleNamespace
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Type
 
+from mava.components.jax import Component
 from mava.utils.config_utils import flatten_dict
 
 
@@ -189,3 +190,32 @@ class Config:
             )
 
         return SimpleNamespace(**self._built_config)
+
+    def get_local_config(self, component: Type[Component]) -> SimpleNamespace:
+        """Get built config for a single component.
+
+        Args:
+            component: component to provide config for.
+
+        Returns:
+            built config for a single component.
+        """
+        if not self._built:
+            raise Exception(
+                "The config must first be built using .build()"
+                "before calling .get_local_config()."
+            )
+
+        global_config = self._built_config
+        local_config: Dict[str, Any] = {}
+
+        config_class = component.config_class()
+        if not config_class:  # no config class for component
+            return SimpleNamespace()
+
+        # Check global config for names which appear in the config class
+        for parameter_name, value in global_config.items():
+            if hasattr(config_class, parameter_name):
+                local_config[parameter_name] = global_config[parameter_name]
+
+        return SimpleNamespace(**local_config)
--- a/mava/systems/jax/builder.py
+++ b/mava/systems/jax/builder.py
@@ -17,7 +17,7 @@
 # have been created.
 
 """Jax-based Mava system builder implementation."""
-
+from types import SimpleNamespace
 from typing import Any, List
 
 from mava.callbacks import BuilderHookMixin, Callback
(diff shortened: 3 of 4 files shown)


mava/systems/jax/config.py
```diff
@@ -189,3 +190,32 @@ def get(self) -> SimpleNamespace:
...
+                "The config must first be built using .build()"
+                "before calling .get_local_config()."
+            )
+
+        global_config = self._built_config
+        local_config: Dict[str, Any] = {}
+
+        config_class = component.config_class()
+        if not config_class:  # no config class for component
+            return SimpleNamespace()
```
sash-a, line 214: Nice quality of life change! Does the `config_class` method in the base component class return `None`? Because then adding configs is no longer a requirement for creating components

mava/systems/jax/config.py
```diff
@@ -189,3 +190,32 @@ def get(self) -> SimpleNamespace:
...
+
+        # Check global config for names which appear in the config class
+        for parameter_name, value in global_config.items():
+            if hasattr(config_class, parameter_name):
+                local_config[parameter_name] = global_config[parameter_name]
```
sash-a, line 219: I think this would be cleaner if you looped through `config_class` because then you don't have to do the hasattr check. See the `dataclass.fields`:https://stackoverflow.com/questions/68668930/how-do-i-pull-out-the-attributes-or-field-names-from-a-dataclass

mava/systems/jax/config.py
```diff
@@ -189,3 +190,32 @@ def get(self) -> SimpleNamespace:
...
+
+        return SimpleNamespace(**local_config)
```
sash-a, line 221: Would `component.config_class(**local_config)` work, cause then the types will even be the same?

sash-a (changes requested):
Looks good to me other than the minor change, think this is definitely the way to go with configs.

Disclaimer: I have not run this locally

mava/systems/jax/config.py
mmorris44:
```
    def config_class() -> Optional[Callable]:
        pass
```
Yes, this is it. But I think writing empty config classes might still be good practice to show future users where to add config is they need it.

sash-a: Sure, but it is convenient for quickly testing stuff

mava/systems/jax/config.py
mmorris44: Nope, but `component.config_class()(**local_config)` works, and I have made the change

mava/systems/jax/config.py
mmorris44: Agreed, done. Also more efficient, because we don't have to loop through the entire global config.

sash-a (dismissed):
Looks good, feels like a great improvement :fire:

mava/components/jax/building/environments.py
```diff
@@ -93,8 +93,9 @@ def on_building_executor_environment(self, builder: SystemBuilder) -> None:
         Args:
             builder : _description_
         """
-        builder.store.executor_environment = self.config.environment_factory(
-            evaluation=False
+        # Global config set by EnvironmentSpec component
+        builder.store.executor_environment = (
+            builder.store.global_config.environment_factory(evaluation=False)
```
AsadJeewa, line 98: small thing but please update mocks.py to also make use of global config

mava/systems/jax/parameter_server.py
```diff
@@ -93,7 +93,7 @@ def add_to_parameters(self, add_to_params: Dict[str, Any]) -> None:
     def step(self) -> None:
         """_summary_"""
         # Wait {non_blocking_sleep_seconds} seconds before checking again
-        non_blocking_sleep(self.store.non_blocking_sleep_seconds)
+        non_blocking_sleep(self.store.global_config.non_blocking_sleep_seconds)
```
AsadJeewa, line 96: much better :D

AsadJeewa (approved):
Happy on my side bar a small change. Thanks, Matthew. Important change

mava/components/jax/building/environments.py
mmorris44: Done. Well spotted.

AsadJeewa (approved):

AsadJeewa (approved):

PR #588, opened 2022-06-20
fix: move terminators test to correct folder

mmorris44:
## What?
Move termination component tests to correct folder
## Why?
They were stored in the wrong location
## How?
Moved the files, changed some import statements
## Extra
Tested by running the tests


AsadJeewa (approved):
Thanks Matthew

PR #591, opened 2022-06-20
fix: Small fix to the decentralised architecture.

DriesSmit:
## What?
Implement a small fix to the decentralised architecture to get the RoboCup example working. Let me know if the fix seems reasonable.

## Why?
The RoboCup example is breaking.

## How?
Small change to the decentralised architecture.


mmorris44 (changes requested):
This is a change to a `tf` system but it's merging into `jax-systems`. Should you not rather be merging this change into `develop`?

PR #596, opened 2022-06-21
Feature/dependency guardrails

mmorris44:
## What?
Guardrails for component dependencies.
## Why?
Currently no way to know at build time that you're missing components which other components depend on.
## How?
- Added `ComponentDependencyGuardrails` component which, when included in the system, checks to see that component dependencies are satisfied
- Added static `required_components()` method to `Component`, which by default returns an empty list
- Set up dependencies for every existing component (bulk of the changes) -> **I don't expect reviewers to check in detail that the correct dependencies have been set up. Easiest way to review this is to check the generated dependency graph for strange-looking connections and then follow that up in the code**
- Created `ComponentDependencyDebugger` component which produces a graph of the system dependencies. `networkx` imported to achieve this
- Added comments to store variables which are created by core system modules
- Moved many variables from the store to the global config
- Remove `shared_weights` from `RandomSamplingSystemInitConfig`, since that system cannot have shared weights
- Move `Loss` and `Step` components to their proper classes
## Extra
Tested on `run_mappo` by added satisfied and unsatisfied requirements to components used in the system. Add in debugging / guardrails components to `mappo` system to see them in use.

component_dependency_map


mava/components/jax/building/adders.py
```diff
@@ -152,7 +178,7 @@ def on_building_executor_adder(self, builder: SystemBuilder) -> None:
 
         adder = reverb_adders.ParallelNStepTransitionAdder(
             priority_fns=builder.store.priority_fns,
-            client=builder.store.data_server_client,
+            client=builder.store.data_server_client,  # Created by builder
```
RuanJohn, line 181: Can probably remove this comment?

mava/components/jax/building/adders.py
```diff
@@ -238,7 +256,7 @@ def on_building_executor_adder(self, builder: SystemBuilder) -> None:
 
         adder = reverb_adders.ParallelSequenceAdder(
             priority_fns=builder.store.priority_fns,
-            client=builder.store.data_server_client,
+            client=builder.store.data_server_client,  # Created by builder
```
RuanJohn, line 259: Similarly here

mava/components/jax/building/data_server.py
```diff
@@ -107,6 +112,21 @@ def name() -> str:
...
 
+    @staticmethod
+    def required_components() -> List[Type[Callback]]:
+        """List of other Components required in the system for this Component to function.
+
+        BaseSystemInit required to set up builder.store.agent_net_keys
+        and for config network_sampling_setup.
+        EnvironmentSpec required to set up builder.store.environment_spec
+        and builder.store.extras_spec.
+        AdderSignature required to set up builder.store.adder_signature_fn.
```
RuanJohn, line 123: General comment also applicable elsewhere, but I really appreciate the detailed descriptions of which components are required.

mava/components/jax/building/environments.py
```diff
@@ -127,7 +152,7 @@ def on_building_executor_environment_loop(self, builder: SystemBuilder) -> None:
         """
         executor_environment_loop = ParallelEnvironmentLoop(
             environment=builder.store.executor_environment,
-            executor=builder.store.executor,
+            executor=builder.store.executor,  # Set up by builder
```
RuanJohn, line 155: Comment can be removed.

mava/components/jax/building/adders.py
mmorris44: I included these comments in a bunch of places to mark where store variables are created by core systems, instead of by components. Do you not find them to be useful when reading and trying to understand the code?

mava/components/jax/building/loggers.py
```diff
@@ -40,11 +41,13 @@ def __init__(
     def on_building_executor_logger(self, builder: SystemBuilder) -> None:
         """[summary]"""
         logger_config = self.config.logger_config if self.config.logger_config else {}
+        # is_evaluator set by builder
```
RuanJohn, line 44: I just realised that all the comments could have been put in on purpose and aren't left over from you reading through the code. If that is the case, please ignore my other comments.

mava/components/jax/building/adders.py
RuanJohn: I tihnk they are useful in that case. Please ignore my comments regarding the comments :sweat_smile:

mava/components/jax/training/base.py
```diff
@@ -51,33 +51,3 @@ class Utility(Component):
...
-        """[summary]"""
-
-    @staticmethod
-    def name() -> str:
-        """_summary_
-
-        Returns:
-            _description_
-        """
-        return "sgd_step"
```
RuanJohn, line 83: Is there a reason we are moving the base components to 'step.py' and `losses.py`?

mmorris44: Yup, I noticed that those particular base components were only used in those particular files, and didn't see why we should keep them entirely separate

RuanJohn: Ah I see, what are your thoughts on moving it back? Becuase once we have more systems they will probably all inheret from the same base `Step` and `Loss` components.

mava/components/jax/training/step.py
```diff
@@ -339,3 +403,18 @@ def config_class() -> Optional[Callable]:
...
+    def required_components() -> List[Type[Callback]]:
+        """List of other Components required in the system for this Component to function.
+
+        GAE required to set up trainer.store.gae_fn.
+        MAPGEpochUpdate required for config num_epochs and num_minibatches.
+
+        Returns:
+            List of required component classes.
+        """
+        return Step.required_components() + [
```
RuanJohn, line 417: This would be impacted by the above comment, if we opt to change it.

mava/components/jax/training/trainer.py
```diff
@@ -110,6 +128,7 @@ def on_building_init_end(self, builder: SystemBuilder) -> None:
                     matches = most_matches
                     builder.store.table_network_config[trainer_key] = sample
 
+        # TODO (Matthew): networks need to be created on the nodes instead?
```
RuanJohn, line 131: Were you able to reach a conclusion on this?

mava/systems/jax/mappo/system.py
```diff
@@ -86,5 +90,7 @@ def design(self) -> Tuple[DesignSpec, Any]:
             **trainer_process,
             distributor=building.Distributor,
             logger=building.Logger,
+            component_dependency_guardrails=ComponentDependencyGuardrails,
+            component_dependency_debugger=ComponentDependencyDebugger,
```
RuanJohn, line 94: I'm not sure whether this should be the default. What are your thoughts?

tests/jax/components/building/adders_test.py
```diff
@@ -149,11 +149,6 @@ def test_parallel_sequence_adder(
     assert parallel_sequence_adder.config.use_next_extras is True
 
     parallel_sequence_adder.on_building_init_start(builder=mock_builder)
-    assert (
-        mock_builder.store.sequence_length
-        == parallel_sequence_adder.config.sequence_length
-    )
-
```
RuanJohn, line 156: Did you remove this assert because it is redundant? Similarly for the other assert removals.

tests/jax/components/executing/base_test.py
```diff
@@ -67,18 +61,6 @@ def mock_executor() -> Executor:
...
-def test_on_building_init(mock_builder: Builder) -> None:
-    """Test on_building_init method from ExecutorInit
-
-    Args:
-        mock_builder:Builder
-    """
-    executor_init = ExecutorInit()
-    executor_init.on_building_init(builder=mock_builder)
-
-    assert mock_builder.store.networks == "after_network_factory"
```
RuanJohn, line 79: Why was this test removed?

mava/components/jax/building/guardrails.py
```diff
@@ -0,0 +1,277 @@
...
+from mava.components.jax import Component
+from mava.core_jax import SystemBuilder
+
+
+@dataclass
+class ComponentDependencyGuardrailsConfig:
+    pass
+
+
+class ComponentDependencyGuardrails(Component):
```
RuanJohn, line 17: This component reads very easily. Good stuff :muscle:

mava/components/jax/building/guardrails.py
```diff
@@ -0,0 +1,277 @@
...
+    ):
+        """Save the config"""
+        self.config = config
+
+    def on_building_init_end(self, builder: SystemBuilder) -> None:  # noqa: C901
+        """Compute component dependencies and save graph."""
+        components: List = builder.callbacks[:]
+
+        # Add required components to list
+        additional_required_components = []
```
RuanJohn, line 105: The naming is a bit difficult to follow. Maybe something like `all_required_components`? Otherwise it seems like there are additional components required in the system

RuanJohn: I realise now that we are actually checking for additional components. Could you please give and example? Becuase it seems like all the components should be included in the list created on line 102? Otherwise, shouldn't this cause an error because we are requiring components that aren't in the system?

mava/components/jax/training/base.py
mmorris44: It feels very weird to me to have `Step` and `Loss` not be in the files `step.py` and `losses.py` respectively. That's also the approach we take to all other abstract component classes.

mava/components/jax/building/guardrails.py
```diff
@@ -0,0 +1,277 @@
...
+    fig_inch_width: float = 18.5
+    fig_inch_height: float = 10.5
+    component_name_fontsize: int = 12
+    node_label_font_size: int = 8
+    required_node_color: str = "#ADD8E6"
+    leaf_required_node_color: str = "#90EE90"
+    detached_node_color: str = "#F0F0F0"
+    dependent_node_color: str = "#D3D3D3"
+    cycle_edge_color: str = "#FF0000"
+    default_edge_color: str = "#000000"
```
RuanJohn, line 85: Something that would make debugging/reading easier would be to please add the colour names next to the hex strings? 
eg. `required_node_color: str = "#ADD8E6" # light blue`

mava/components/jax/training/step.py
```diff
@@ -113,6 +137,45 @@ def on_training_step(self, trainer: SystemTrainer) -> None:
...
+            List of required component classes.
+        """
+        return [
+            TrainerDataset,
+            mava.components.jax.building.adders.ParallelSequenceAdder,
+            BaseTrainerInit,
+            mava.components.jax.training.model_updating.EpochUpdate,
+            mava.components.jax.training.model_updating.MinibatchUpdate,
+            Networks,
+        ]
```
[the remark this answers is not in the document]
mmorris44: See elsewhere for what I've done when there are specific components of the abstract that require extra dependencies. I can do the same here?

`MinibatchUpdate` and `EpochUpdate` are already the abstract versions.
`ParallelSequenceAdder` is specific, so maybe I should move that dependency to the `MAPGWithTrustRegionStep` class?

mava/components/jax/training/step.py
mmorris44: Noted, thanks! This is where I am proposing moving `ParallelSequenceAdder` to.
All the other requirements seem necessary for any trainer step one might implement.

RuanJohn (changes requested):
Thank you for all the work @mmorris44! Everything looks really good just a few minor questions/suggestions from my side.

mava/components/jax/training/trainer.py
mmorris44: Afraid not, I think this ties into the Jax network init issue though. Do we currently have an issue out for this?
I don't want to remove the comment if we do not.

mava/systems/jax/mappo/system.py
mmorris44: I think that the debugger should not be there by default, but that the guardrails should. I have just removed the debugger.

mava/components/jax/training/step.py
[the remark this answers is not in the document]
RuanJohn: What you have done elsewhere makes sense. The reason I think the update components should still move is because they are abstract for policy gradient algorithms but a system like dqn might not need them at all. What are yout thoughts?

mava/systems/jax/mappo/system.py
RuanJohn: Sounds good. Thank you

mava/components/jax/training/base.py
RuanJohn: In that case, I agree with you. Please keep as is :)

mava/components/jax/training/trainer.py
RuanJohn: I am not sure whether is an issue out yet. Please keep in that case.

tests/jax/components/executing/base_test.py
mmorris44: See `ExecutorInit`. It now no longer creates the networks since they are just created in `TrainerInit`

mava/components/jax/building/guardrails.py
mmorris44: Agreed, will rename to `all_required_components`, it was named badly by me

tests/jax/components/executing/base_test.py
RuanJohn: Make sense. Thank you

tests/jax/components/building/adders_test.py
RuanJohn: Makes sense, thank you for clarifying :)

mava/components/jax/building/guardrails.py
mmorris44: Agreed, done

RuanJohn: Thank you

mava/components/jax/training/step.py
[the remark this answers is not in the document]
mmorris44: Any `Step` will always need `BaseTrainerInit` and `Networks`, for sure.
Also, it seems like both datasets we have defined a config of `sample_batch_size`, so that's also always needed, no?

Would we not also need an `EpochUpdate` and a `MinibatchUpdate` for DQN?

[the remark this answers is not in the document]
RuanJohn: `BaseTrainerInit`, `Networks` and `TrainerDataset` seem general enough to me. But `MinibatchUpdate` and `EpochUpdate` seem a bit too PPO specific to me.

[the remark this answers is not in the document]
mmorris44: Agreed, thanks for the feedback. I've made the requested changes.

mava/components/jax/building/data_server.py
AsadJeewa: Agreed :smile:

mava/components/jax/building/data_server.py
```diff
@@ -136,6 +141,21 @@ def name() -> str:
...
+
+        Returns:
+            List of required component classes.
+        """
+        return [EnvironmentSpec, AdderSignature, BaseSystemInit]
```
AsadJeewa, line 157: with regards to the diagram showing 32 (ParallelSequenceAdderSignature) as having no dependencies, isn't that misleading? since the data_server will access the signature_fn from the store, which would either be set by the  transition or sequence adder signature component, depending on what component the system has.

AsadJeewa: The diagram is a little confusing with both abstract classes and subclasses are shown. For example, since the MAPPO system has parallelSequenceAdder, it should have an implied dependency on ParallelSequenceAdderSignature

mava/components/jax/building/adders.py
```diff
@@ -264,7 +286,7 @@ def on_building_executor_adder(self, builder: SystemBuilder) -> None:
 
         adder = reverb_adders.ParallelSequenceAdder(
```
AsadJeewa, line 287: according to the diagram, this component has a dependency on GAE so should add a required components method (the other 2 deps are covered by the base class)

mava/components/jax/training/step.py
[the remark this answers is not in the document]
RuanJohn: Thank you :)

mava/components/jax/training/step.py
RuanJohn: Can you also please move `ParallelSequenceAdder` here?

mava/components/jax/building/guardrails.py
RuanJohn: > 

Just following up re this question for incase it got lost in the mix.

mava/components/jax/training/step.py
AsadJeewa: I agree that parallelsequenceadder should be moved here. Perhaps step could have Adder but maybe that is redundant

mava/components/jax/building/adders.py
```diff
@@ -43,6 +46,18 @@ def name() -> str:
...
+    def required_components() -> List[Type[Callback]]:
+        """List of other Components required in the system for this Component to function.
+
+        BaseTrainerInit required to set up builder.store.table_network_config.
+        BaseSystemInit required to set up builder.store.unique_net_keys.
+
+        Returns:
+            List of required component classes.
+        """
+        return [BaseTrainerInit, BaseSystemInit]
```
AsadJeewa, line 59: The diagram is a little confusing with both abstract classes and subclasses are shown. I have been trying to see why the circular dependency exists but it does not make sense that Step has so many dependencies when it is an abstract class with no implementations. Do you agree?

mava/components/jax/building/guardrails.py
```diff
@@ -0,0 +1,277 @@
...
+    dependent_node_color: str = "#D3D3D3"  # light gray
+    cycle_edge_color: str = "#FF0000"  # red
+    default_edge_color: str = "#000000"  # black
+    layout_scale: int = 2
+    node_size_increment: int = 80
+    min_node_size: int = 190
+    dependency_graph_save_path: str = "./component_dependency_map.png"
+
+
+class ComponentDependencyDebugger(Component):
```
AsadJeewa, line 92: very useful class :smile:

mava/components/jax/training/step.py
```diff
@@ -113,6 +137,41 @@ def on_training_step(self, trainer: SystemTrainer) -> None:
...
+        BaseTrainerInit required to set up trainer.store.networks
+        and trainer.store.trainer_agent_net_keys.
+        Networks required to set up trainer.store.key.
+
+        Returns:
+            List of required component classes.
+        """
+        return [
+            TrainerDataset,
+            mava.components.jax.building.adders.ParallelSequenceAdder,
```
AsadJeewa, line 169: Step should just depend on Adders as per my first comment on abstract components

mava/components/jax/training/step.py
mmorris44: Done @RuanJohn. Thanks for the feedback!
@AsadJeewa I don't see why an adder would be directly required by Step? If it's required, then it's required implicitly by an upstream dependency.

mava/components/jax/training/step.py
```diff
@@ -140,8 +199,9 @@ def on_training_init_start(self, trainer: SystemTrainer) -> None:
             None.
         """
         # Note (dries): Assuming the batch and sequence dimensions are flattened.
```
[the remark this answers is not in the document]
AsadJeewa: or are you of the opinion that the order of callbacks handles this?

mava/components/jax/building/guardrails.py
```diff
@@ -0,0 +1,277 @@
...
+        for component in builder.callbacks:
+            for required_component in component.required_components():
+                required_component_present = False
+
+                for system_component in builder.callbacks:
+                    if isinstance(system_component, required_component):
+                        required_component_present = True
+                        break
+
+                if not required_component_present:
```
AsadJeewa, line 36: NIce code! Clear error messages :+1:

AsadJeewa (changes requested):
Thanks, @mmorris44 
This is amazing!!!! Few comments/ queries

mava/components/jax/training/step.py
RuanJohn: Thanks @mmorris44 :muscle:

RuanJohn (approved):

mava/components/jax/training/step.py
AsadJeewa: What I mean is that a concrete extension of step (MAPGWithTrustRegionStep) requires parallelsequenceadder so the abstract super class (Step) would depend on adder

mava/components/jax/training/step.py
```diff
@@ -56,6 +65,21 @@ def name() -> str:
...
+        and trainer.store.trainer_counts.
+        Logger required to set up trainer.store.trainer_logger.
+
+        Returns:
+            List of required component classes.
+        """
+        return [TrajectoryDataset, Step, TrainerParameterClient, Logger]
+
 
 class DefaultTrainerStep(TrainerStep):
```
AsadJeewa, line 84: Just a general comment: there are too many step components in this class and made more confusing since TrainerStep is not an extension of Step and has a dependency on it

mava/components/jax/building/data_server.py
AsadJeewa: Please add comments #todo to look in those redundant components. Perhaps issues are overkill

mava/components/jax/training/step.py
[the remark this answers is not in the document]
mmorris44: There's no functional issue with the circular dependency, but it just means the components are more interleaved than perhaps they should be: i.e. it might be poor design

mava/components/jax/training/step.py
mmorris44: Might be good to create an issue to rename some of the `Step` classes? `TrainerStep ` vs `Step` are not great names for disambiguation, I agree.

mava/components/jax/building/adders.py
AsadJeewa: Discussed this: abstract classes do depend on other abstract classes but only as required

mava/components/jax/training/step.py
AsadJeewa: Yeah that would be important

mava/components/jax/building/data_server.py
AsadJeewa: We discussed `ParallelSequenceAdder` but we should revisit the necessity of the `ExecutorInit` and I think `ExtrasLogProbs`

mava/components/jax/training/step.py
AsadJeewa: @mmorris44 https://github.com/instadeepai/Mava/issues/678

mava/components/jax/building/data_server.py
AsadJeewa: Created an issue @mmorris44 
https://github.com/instadeepai/Mava/issues/679

AsadJeewa (dismissed):
Happy on my side! :partying_face: :fireworks: :smile: 
Thanks @mmorris44

RuanJohn (approved):

AsadJeewa (approved):

PR #605, opened 2022-06-24
Tests/update testing pipeline

RuanJohn:
## What?
Update the CI testing pipeline to run all Jax and TF tests when merging into `feature/jax-systems`, `develop` or `main`. 
## Why?
We were currently only running the tf tests.
## How?
Rename `bash_scripts/tf_tests.sh` to `bash_scripts/tests.sh` and run all tests in the `tests` directory. 
## Extra
Kept the lightweight `bash_scripts/jax_tests.sh` for now, but we can delete it. 
There are some Jax and TF tests that are faulty and break the current testing pipeline I have skipped these tests using `@pytest.mark.skip` these tests need fixing and can be fixed in separate PRs.


bash_scripts/tests.sh
```diff
@@ -46,7 +46,7 @@ apt-get -y install git
 apt-get install swig -y
 
 # Install depedencies
-pip install .[tf,envs,reverb,testing_formatting,record_episode]
+pip install .[jax,envs,reverb,testing_formatting,record_episode]
```
AsadJeewa, line 49: as you mentioned, we should delete jax_tests.sh

AsadJeewa, line 49: and if all the tests are being run now (including tf ones), why is the tf dependency removed?

tests/tf/systems/madqn_system_test.py
```diff
@@ -44,7 +44,7 @@ def test_madqn_on_debugging_env(self) -> None:
 
         # networks
         network_factory = lp_utils.partial_kwargs(
-            madqn.make_default_networks, value_networks_layer_sizes=(64, 64)
+            madqn.make_default_networks, value_networks_layer_sizes=(32, 32)
```
AsadJeewa, line 47: comment applies elsewhere: why were hyparparam changed? to make test faster?

AsadJeewa (commented):
Thanks, Ruan. Few comments/ questions

tests/tf/systems/madqn_system_test.py
RuanJohn: Yes, just to make the tests a bit more lightweight. Happy to increase layer sizes and batch sizes if you think that is a better idea?

bash_scripts/tests.sh
RuanJohn: Sorted. :call_me_hand:

tests/jax/core_system_components/executor_test.py
```diff
@@ -145,6 +147,8 @@ def test_executor_parameter_server_system() -> System:
     return TestSystemExecutorAndParameterSever()
 
 
+# TODO: fix test.
+@pytest.mark.skip(reason="test is currently breaking ci pipeline")
```
AsadJeewa, line 151: Are these tests breaking because they were created a while ago and changes were made to the repo since then?

bash_scripts/tests.sh
AsadJeewa: That is misleading. One would think that JAX and TF are separate but I guess it is a temporary solution until we deprecate JAX

tests/tf/systems/madqn_system_test.py
AsadJeewa: Pros and cons to both but I get the need to make the tests lightweight.

siddarthsingh1 (commented):
Thanks for the update @RuanJohn. The changes seem to make sense so I only have one question.

tests/tf/systems/mappo_system_test.py
```diff
@@ -50,7 +50,7 @@ def test_mappo_on_debugging_env(self) -> None:
             environment_factory=environment_factory,
             network_factory=network_factory,
             num_executors=1,
-            batch_size=32,
+            batch_size=2,
```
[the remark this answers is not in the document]
RuanJohn: Yes, it was to avoid reverb issues.

AsadJeewa (approved):
Thanks Ruan. Happy on my side

mmorris44 (changes requested):
Thanks @RuanJohn! Confirmed that `make run-tests` works on my side after `make build`.
I think that [this](https://adamj.eu/tech/2020/07/28/how-to-fix-a-pytest-collection-warning-about-web-tests-test-app-class/) is why all the warnings are showing up.

It's not a breaking issue, since we don't put our tests within classes, but the warnings are frustrating. Calling the classes `TestSomething` causes `pytest` to try test them, so we can solve this by either renaming the offending classes or adding `__test__ = False` to the class (see the bottom of this [page](https://pytest.org/en/7.1.x/example/pythoncollection.html)). I.e.:
```
# Will not be discovered as a test
class TestClass:
    __test__ = False
```

RuanJohn:
> Thanks @RuanJohn! Confirmed that `make run-tests` works on my side after `make build`. I think that [this](https://adamj.eu/tech/2020/07/28/how-to-fix-a-pytest-collection-warning-about-web-tests-test-app-class/) is why all the warnings are showing up.
> 
> It's not a breaking issue, since we don't put our tests within classes, but the warnings are frustrating. Calling the classes `TestSomething` causes `pytest` to try test them, so we can solve this by either renaming the offending classes or adding `__test__ = False` to the class (see the bottom of this [page](https://pytest.org/en/7.1.x/example/pythoncollection.html)). I.e.:
> 
> ```
> # Will not be discovered as a test
> class TestClass:
>     __test__ = False
> ```

Incorporated these changes @mmorris44 :+1:

AsadJeewa (dismissed):
Thanks, @RuanJohn Happy with the latest changes. Some of these tests will be fixed with existing PRs (such as the adders)

mmorris44 (dismissed):
Approved, thanks @RuanJohn. I still get one error when running the system tests, but happy to merge this for now to get our pipeline working. We can keep an eye out for it in the future.

AsadJeewa (approved):
Just had to resolve an adders merge conflict. I also ran make run-tests locally and all looks good (240 passed, 17 skipped in 50.47s ). Happy to merge :smile:

mmorris44:
> Just had to resolve an adders merge conflict. I also ran make run-tests locally and all looks good (240 passed, 17 skipped in 50.47s ). Happy to merge 😄

@AsadJeewa please run `make run-integration-tests` as well to see if it works.

mmorris44 (approved):

AsadJeewa (approved):

PR #607, opened 2022-06-27
Feature: episode monitoring

sash-a:
## What?
Added environment monitoring (recording videos of episodes)
## Why?
Very useful for debugging
## How?
Added a `MonitorExecutorEnvironmentLoop` component which creates a `MonitorParallelEnvironmentLoop` on evaluators and a regular `ParallelEnvironmentLoop` on executors
## Extra
* Had to edit `MonitorParallelEnvironmentLoop` to get the correct counter. The current way the env loops and loop wrappers support both tf and jax mava makes them incredibly messy, I'll open an issue and get on this ASAP
* I changed `checkpoint_subpath` config in the `ParameterServerConfig` to `experiment_path` this allows it to be a more general path for other things that we want to save there (e.g the recordings)


mava/components/jax/updating/parameter_server.py
```diff
@@ -29,7 +29,7 @@
 @dataclass
 class ParameterServerConfig:
     checkpoint: bool = True
-    checkpoint_subpath: str = "~/mava/"
+    experiment_path: str = "~/mava/"
```
sash-a, line 32: I'm open to suggestions on this, but I think changing this to be a more general name is useful

mava/wrappers/environment_loop_wrappers.py
```diff
@@ -361,8 +361,11 @@ def _retrieve_render(self) -> Optional[np.ndarray]:
 
     def _append_frame(self) -> None:
         """Appends a frame to the sequence of frames."""
+        # THIS IS UGLY, we need to separate jax and tf functionality
```
AsadJeewa, line 364: Rather change this comment to a TODO

mava/components/jax/updating/parameter_server.py
AsadJeewa: I agree with this. The comment we have everwhere `# Checkpointer appends "Checkpoints" to checkpoint_dir` is not ideal but this is because we are using Acme's checkpointer

mava/components/jax/building/environments.py
```diff
@@ -137,3 +138,68 @@ def on_building_executor_environment_loop(self, builder: SystemBuilder) -> None:
...
+    def on_building_executor_environment_loop(self, builder: SystemBuilder) -> None:
+        """
+        Builds a `ParallelEnvironmentLoop` on all executors and a
+        `MonitorParallelEnvironmentLoop` on the executor and stores it
+         in the `builder.store.system_executor`
+
+        Args:
+            builder : the system builder
+        """
+        if builder.store.is_evaluator:
```
AsadJeewa, line 171: this if statement is not ideal but it is how we do things elsewhere when we only want this on the evaluator (previously we had eval_loop_fn)

AsadJeewa (commented):
Thanks Sasha. Few small comments

examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo.py
```diff
@@ -68,7 +69,7 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
         )
 
     # Checkpointer appends "Checkpoints" to checkpoint_dir
```
AsadJeewa, line 71: change this comment to say experiment_path

examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo.py
```diff
@@ -88,19 +89,20 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
 
     # Create the system.
     system = mappo.MAPPOSystem()
-
+    system.update(MonitorExecutorEnvironmentLoop)
```
AsadJeewa, line 92: I like this as this clears shows a case when we do not want PPO system to have recording but the specific run file that add the feature :smile:

AsadJeewa (changes requested):
Thanks Sasha!. Few comments/ changes requested

mava/components/jax/updating/parameter_server.py
AsadJeewa: But you would need to update all other run files

AsadJeewa:
Also, there is a small error in the build which should disappear with a local pre-commit

mava/components/jax/building/environments.py
```diff
@@ -137,3 +138,68 @@ def on_building_executor_environment_loop(self, builder: SystemBuilder) -> None:
...
+            executor_environment_loop = MonitorParallelEnvironmentLoop(
```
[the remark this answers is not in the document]
sash-a: Discussed offline and the conclusion we came to is that it is the environments fault for not handling the render mode of `rgb_array` correctly. In order to not show rendering during training for these types of envs:

```python
import pyvirtualdisplay
display = Display(visible=False)
display.start()
```

AsadJeewa:
docker run debugging env:` [evaluator/0] Render frames exception: HINT: you can install pyglet directly via 'pip install pyglet'. But if you really just want to install all Gym dependencies and not have to think about it, 'pip install -e .[all]' or 'pip install gym[all]' will do it.`

mava/components/jax/building/environments.py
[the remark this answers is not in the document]
AsadJeewa: with `make build record=true` and `make run-record` the render does not appear. It is only with venv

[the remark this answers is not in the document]
AsadJeewa: @sash-a please create an issue for this then I can approve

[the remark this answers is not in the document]
sash-a: Created, issue #629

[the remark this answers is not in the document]
AsadJeewa: Thank you

AsadJeewa (approved):
Thanks for the changes @sash-a 
Looks good

examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo.py
RuanJohn: Does this mean that our example will now always record episodes? If so, I think it would be better to make it a separate example.

mava/wrappers/environment_loop_wrappers.py
AsadJeewa: one outstanding change

RuanJohn (approved):
Looks good! Thank you @sash-a! :call_me_hand:

AsadJeewa (approved):

PR #610, opened 2022-06-28
feat: split trainer init into multiple components.

RuanJohn:
## What?
Split `TrainerInit` component into three components. 
## Why?
Avoid extra if statements and allows for setting the default to being a single trainer that trains all networks. 
## How?
Refactored single `TrainerInit` component to be `SingleTrainerInit`, `OneTrainerPerNetworkInit` and `CustomTrainerInit` components. 
## Extra
Kept the trainer enums, but happy to remove them as they are not used downstream in the code. 
Tested by running `examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo.py` with `reverb==0.7.3`.


mava/components/jax/training/trainer.py
```diff
@@ -15,57 +15,89 @@
...
+@dataclass
+class SingleTrainerInitConfig(BaseTrainerInitConfig):
+    pass
+
+
+class SingleTrainerInit(BaseTrainerInit):
+    def __init__(self, config: SingleTrainerInitConfig = SingleTrainerInitConfig()):
+        """Initialises a single trainer.
+
+        Here a single trainer is used to train all networks.
```
AsadJeewa, line 74: small thing but I think you can remove "here a" from the comment (elsewhere too)

mava/components/jax/training/trainer.py
```diff
@@ -15,57 +15,89 @@
...
         Raises:
-            ValueError: _description_
+            ValueError: Raises an error when trainer_networks is not
+                        set to single_trainer.
         """
         trainer_networks = self.config.trainer_networks
         unique_net_keys = builder.store.unique_net_keys
 
         # Setup trainer_networks
-        if not isinstance(trainer_networks, dict):
```
AsadJeewa, line 55: Glad to see this code gone lol :smile:

mava/components/jax/training/trainer.py
```diff
@@ -15,57 +15,89 @@
...
-                    for i in range(len(unique_net_keys))
-                }
-            else:
-                raise ValueError(
-                    "trainer_networks does not support this enums setting."
-                )
-        else:
-            builder.store.trainer_networks = trainer_networks
+
+        if trainer_networks != enums.Trainer.single_trainer:
```
AsadJeewa, line 95: I cannot see any use in the JAX systems. I think you should create an issue to remove enums once we deprecate TF and maybe add a TODO comment here also

mava/components/jax/training/trainer.py
```diff
@@ -113,9 +145,197 @@ def on_training_utility_fns(self, trainer: SystemTrainer) -> None:
...
+            builder : the system builder
+        Raises:
+            ValueError: Raises an error when trainer_networks is not
+                        passed in as a dictionary.
+        """
+        trainer_networks = self.config.trainer_networks
+        unique_net_keys = builder.store.unique_net_keys
+
+        # Setup trainer_networks
+        if not isinstance(trainer_networks, dict) or trainer_networks == {}:
```
AsadJeewa, line 289: More of a question but why do we use dict in (various places in) mava instead of just using {} notation (from what I read online, dict is slower)

mava/components/jax/training/trainer.py
```diff
@@ -15,57 +15,89 @@
...
     trainer_networks: Union[
         Dict[str, List], enums.Trainer
     ] = enums.Trainer.single_trainer
 
 
-class TrainerInit(Component):
-    def __init__(self, config: TrainerInitConfig = TrainerInitConfig()):
-        """_summary_
+@dataclass
+class SingleTrainerInitConfig(BaseTrainerInitConfig):
```
AsadJeewa, line 66: If subclass configs add nothing, shouldn't we just use the base and remove this?

mava/components/jax/training/trainer.py
```diff
@@ -124,4 +344,4 @@ def config_class() -> Optional[Callable]:
         Returns:
             config class/dataclass for component.
         """
-        return TrainerInitConfig
+        return CustomTrainerInitConfig
```
AsadJeewa, line 347: I see you got a draft PR. In the tests for this, it would be important to test the custom trainer init as it requires a dictionary that conforms to unique_net_keys

AsadJeewa (changes requested):
Thanks, Ruan. Small changes requested

mava/components/jax/training/trainer.py
RuanJohn: Definitely, that draft PR is currently testing the old `TrainerInit`, so once this PR goes through I'll test the custom cases separately.

mava/components/jax/training/trainer.py
RuanJohn: I had a think about this, and I actually think we can remove configs for `SingleTrainerInit` and for `OneTrainerPerNetworkInit` that way the components don't need the enums check either.

mava/components/jax/training/trainer.py
RuanJohn: Sounds good. Will do.

mava/components/jax/training/trainer.py
RuanJohn: The components do not require any enums any more, but I will still make an issue for investigating the use of enums after deprecating tf.

AsadJeewa (approved):
Thanks, @RuanJohn. Happy on my side :D

AsadJeewa (approved):
re-approve due to base branch change

PR #611, opened 2022-06-28
test: testing launcher class for Jax-based Mava systems

### LINKED ISSUE #546: [TEST] Jax System Core Launcher
### What do you want to test?
Jax system core launcher (in `mava/systems/jax/`)

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

OmaymaMahjoub:
## What?
Test: Unit tests for Jax system core launcher
## Why?
## How?
Unit test for Launcher component from `mava/systems/jax/launcher.py`
## Extra
Close #546 
Integration test for `launch()` method within `Launcher`


tests/jax/core_system_components/launcher_test.py
```diff
@@ -15,4 +15,381 @@
...
+
+    assert launcher._nodes[-1] == launcher._node_dict["data_server"]
+    assert data_server == launcher._nodes[-1]
+
+    assert isinstance(launcher._node_dict["data_server"], Client)
+    assert launcher._node_dict["data_server"]._server_address == (
+        f"localhost:{launcher._replay_server._port}"
+    )
+    assert isinstance(launcher._node_dict["data_server"]._client, pybind.Client)
+    assert launcher._node_dict["data_server"]._signature_cache == {}
```
sash-a, line 263: You've already tested this above, you can remove it from here.

tests/jax/core_system_components/launcher_test.py
```diff
@@ -15,4 +15,381 @@
...
+        parameter_server = launcher.add(
+            mock_parameter_server_fn,
+            node_type=NodeType.corrier,
+            name="parameter_server_test",
+        )
+    parameter_server = launcher.add(
+        mock_parameter_server_fn,
+        node_type=NodeType.corrier,
+        name="parameter_server",
+    )
```
sash-a, line 283: I'm not understanding this, why does the above raise an exception but below it doesn't? It looks like the same code?

Also please be more explicit with error type

tests/jax/core_system_components/launcher_test.py
```diff
@@ -15,4 +15,381 @@
...
+    with pytest.raises(Exception):
+        parameter_server_2 = launcher.add(
+            mock_parameter_server_second_fn,
+            node_type=NodeType.corrier,
+            name="parameter_server",
+        )
```
sash-a, line 354: Ah here this makes sense, you would expect an error from 2 of the same nodes. I assume there is some reason that multi-process allows for two nodes with the same name?

Probably worth looking into @mmorris44

tests/jax/core_system_components/launcher_test.py
```diff
@@ -15,4 +15,381 @@
...
+
+
+def test_get_nodes_multi_process() -> None:
+    """Test get_nodes method in case of multi process"""
+    launcher = Launcher(multi_process=True)
+    with pytest.raises(Exception):
```
sash-a, line 360: Specific exception type

sash-a (changes requested):
Looks good! Just a few general fixes. Combine the functions and their mocks and catch more specific exceptions.

Also I'm confused about when/why you are allowed to add nodes with the same name, in my opinion you probably shouldn't ever be able to, but it seems that if the launcher is multi-process then it is ok. Paging @arnupretorius @mmorris44

tests/jax/core_system_components/launcher_test.py
```diff
@@ -15,4 +15,381 @@
...
+    )
+    assert [parameter_server] == launcher._program._groups["parameter_server_test"][
+        -1
+    ]._created_handles
+
+    assert not hasattr(launcher, "_nodes")
+    assert not hasattr(launcher, "_node_dict")
+
+
+def test_add_multi_process_two_add_same_name(
```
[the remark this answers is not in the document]
OmaymaMahjoub: In multi process case if we have two nodes with same name it append the nodes under same label and in this case it is parameter server:
Here is the code:  `self._groups[label].append(node)`
And the test verified that it works:
```
assert (
        launcher._program._groups["parameter_server_test"][0]._constructor
        == mock_parameter_server_fn
    )
    assert (
        launcher._program._groups["parameter_server_test"][1]._constructor
        == mock_parameter_server_second_fn
    )
```

tests/jax/core_system_components/launcher_test.py
OmaymaMahjoub: it is the case of non-multi-process and with nodes of type reverb_node so we need to make a test since it is a different case.
What do you think?

tests/jax/core_system_components/launcher_test.py
OmaymaMahjoub: The name of the node must be one of these 5 ones:
```
self._node_dict: Dict = {
                "data_server": None,
                "parameter_server": None,
                "executor": None,
                "evaluator": None,
                "trainer": None,
            }
```

tests/jax/core_system_components/launcher_test.py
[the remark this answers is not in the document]
sash-a: aha I see, well then I'll assume that is the desired behavior and you can leave the tests as is

tests/jax/core_system_components/launcher_test.py
sash-a: Ok sure, feel free to keep it then

tests/jax/core_system_components/launcher_test.py
```diff
@@ -15,4 +15,381 @@
...
+    """
+    launcher = Launcher(multi_process=True)
+    data_server = launcher.add(
+        mock_data_server_fn,
+        node_type=NodeType.reverb,
+        name="data_server_test",
+    )
+    parameter_server = launcher.add(
+        mock_parameter_server_fn,
+        node_type=NodeType.corrier,
```
mmorris44, line 159: This was changed to `NodeType.courier` by Asad in another PR, so this code should not work.

mmorris44, line 159: This was changed to `NodeType.courier` by Asad in another PR, so this code should not work.

mmorris44 (changes requested):
Found a change which is breaking the pipeline. Should be an easy fix

tests/jax/core_system_components/launcher_test.py
mmorris44: Thanks for flagging this @sash-a. I can't run launchpad locally and don't have the most experience with it, so @arnupretorius or @KaleabTessera are probably best suited to advise. Here's what I think though:
For multi-processing, it adds nodes to a `lp.Program`, grouping the node by the given name (i.e. you can have multiple executor nodes which will be grouped together), and `lp` handles this just fine. See line 97 of `mava/systems/jax/launcher.py`.

I don't see why we shouldn't support this for single-process systems though.

tests/jax/core_system_components/launcher_test.py
OmaymaMahjoub: oh I see thank you @mmorris44!

mmorris44 (approved):
Thanks @OmaymaMahjoub! Happy with the PR now :)
Also, please don't forget to create an issue for the launcher behaviour around nodes with the same name (i.e. it's weird that multi-processing supports it but single-process does not, it's also weird to me that the node name gets set in the `arguments` arg of `add()` in `launcher.py`, see line 77 of `distributor.py`).

OmaymaMahjoub:
> Thanks @OmaymaMahjoub! Happy with the PR now :) Also, please don't forget to create an issue for the launcher behaviour around nodes with the same name (i.e. it's weird that multi-processing supports it but single-process does not, it's also weird to me that the node name gets set in the `arguments` arg of `add()` in `launcher.py`, see line 77 of `distributor.py`).

Thanks @mmorris44  I made an [issue](https://github.com/instadeepai/Mava/issues/636) about it. Yeah, I found this strange, especially that in the multi-process case we have the full freedom to use any names for the nodes while in single-process nope all the details are within the issue!

tests/jax/core_system_components/launcher_test.py
sash-a: I think either way works as long as it is consistent, so either both allow for multiple nodes with the same name or neither do, unless there's a reason I'm missing where it makes sense in one scenario, but doesn't make sense in another

PR #613, opened 2022-06-29
Test/core executor

### LINKED ISSUE #545: [TEST] Jax System Core Executor
### What do you want to test?
Jax system core executor (in `mava/systems/jax/`)

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

mmorris44:
## What?
Test for core Mava executor.
## Why?
Robust testing for Mava release.
## How?
- Wrote unit tests for methods in the class
- Used `HookOrderTracking` to test hook orders
- Moved some integration tests out of existing files and into a new folder
- Fixed some issues with the parameter server test that a merged PR created
## Extra
Closes #545. Tests pass and example runs.
**Note:** tests in the `integration` folder are not new, they're just moved from previous locations.


RuanJohn (approved):
Looks good. I am not reviewing the integration tests now since we want to handle these when we move on to integration tests. All tests are passing and faulty executor integration tests are skipped.

tests/jax/core_system_components/executor_test.py
```diff
@@ -1,5 +1,5 @@
 # python3
-# Copyright 2021 InstaDeep Ltd. All rights reserved.
+# Copyright 2022 InstaDeep Ltd. All rights reserved.
```
arnupretorius, line 2: I think we should keep the year 2021 (the year Mava was released) for all files. This is also what I have seen done in other repos. However, I'm of course not an expert on this, so could be wrong. But keeping the year the same would make it easier to maintain.

tests/jax/core_system_components/executor_test.py
```diff
@@ -13,266 +13,234 @@
...
+from typing import Dict, List
 
-import functools
-from typing import Dict, Tuple
-
-import acme
+import dm_env
 import numpy as np
 import pytest
+from acme.types import NestedArray
```
arnupretorius, line 23: Let's please use Mava types

tests/jax/core_system_components/executor_test.py
```diff
@@ -13,266 +13,234 @@
...
+    """Test that store is handled properly in update"""
+    test_executor.update(wait=False)
+    assert not test_executor.store._wait
+    test_executor.update(wait=True)
+    assert test_executor.store._wait
+
+
+def test_init_hook_order(test_executor: TestExecutor) -> None:
+    """Test if init hooks are called in the correct order"""
+    assert test_executor.hook_list == [
```
arnupretorius, line 164: This is great, nice that we are testing hook order. We should also do this in other classes.

tests/jax/core_system_components/parameter_server_test.py
```diff
@@ -15,55 +15,17 @@
 
 """Tests for parameter server class for Jax-based Mava systems"""
 
-import functools
 import time
 from types import SimpleNamespace
-from typing import Dict, List, Tuple
+from typing import List
```
arnupretorius, line 20: Not sure why some code for the parameter server tests made it into this PR. But would be good to keep things separate as much as possible.

tests/jax/integration/__init__.py
```diff
@@ -1 +1,15 @@
-"""Jax systems integration tests"""
+# python3
+# Copyright 2022 InstaDeep Ltd. All rights reserved.
```
arnupretorius, line 2: Same as earlier, I think we can just keep all files at 2021

tests/jax/core_system_components/parameter_server_test.py
arnupretorius: Ah wait I see, it is because of the integration test. In that case, we could've perhaps made separate PRs for unit/integration testing. This instance is minor, I suspect it was just moving things around and so on, so no worries. But just something to keep in mind going forward, that we separate PRs sensibly.

arnupretorius (commented):
Look good to me 🙂 Thanks for this! 

Just left some minor comments.

tests/jax/core_system_components/executor_test.py
RuanJohn: Make sense to me. Updated.

tests/jax/core_system_components/executor_test.py
RuanJohn: Sorted.

tests/jax/core_system_components/executor_test.py
RuanJohn: It really is a nice way to go about testing the core components. We have been doing this for all core components where the hooks get called sequentially.

tests/jax/core_system_components/parameter_server_test.py
RuanJohn: Indeed, it looks like it was just because things were getting moved around.

mmorris44 (commented):
I'm happy with all the changes suggested by @arnupretorius and made by @RuanJohn. Would approve if I could :)

RuanJohn (approved):
Thanks @mmorris44!

PR #617, opened 2022-07-05
Test/jax environments

### LINKED ISSUE #524: [TEST] Jax environments
### What do you want to test?
Jax environments components

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

mmorris44:
## What?
Test jax environment components.
## Why?
Robust testing of components before jax systems release.
## How?
- Tested `EnvironmentSpec`, `ExecutorEnvironmentLoop`, and `ParallelExecutorEnvironmentLoop`
## Extra
Closes #524


tests/jax/components/building/environments_test.py
```diff
@@ -0,0 +1,217 @@
...
+        assert environment.environment.num_agents == 10
+
+    def test_on_building_init_start(
+        self, test_environment_spec: EnvironmentSpec, test_builder: SystemBuilder
+    ) -> None:
+        """Test by manually calling the hook and checking the store."""
+        test_environment_spec.on_building_init_start(test_builder)
+
+        # Assert for type and extra spec
+        environment_spec = test_builder.store.environment_spec
```
RuanJohn, line 100: Due to the env specs renaming PR this test currently has errors. Please update the names to match the new naming :)

tests/jax/components/building/environments_test.py
```diff
@@ -0,0 +1,219 @@
...
+    """Pytest fixture for executor environment loop"""
+
+    class ExecutorStatsWrapperClass:
+        def __init__(self, executor_environment_loop: Any) -> None:
+            self.executor_environment_loop = executor_environment_loop
+            self.wrapped = True
+
+    config = ExecutorEnvironmentLoopConfig(
+        should_update=False,
+        executor_stats_wrapper_class=ExecutorStatsWrapperClass,  # type: ignore
```
OmaymaMahjoub, line 63: Why the executor_stats_wrapper_class is `ExecutorStatsWrapperClass` while it is mentioned as an object of `EnvironmentLoopStatisticsBase`?

tests/jax/components/building/environments_test.py
```diff
@@ -0,0 +1,219 @@
...
+    """Pytest fixture for executor environment loop"""
+    config = ExecutorEnvironmentLoopConfig(should_update=False)
+    return AbstractExecutorEnvironmentLoop(config=config)
+
+
+@pytest.fixture
+def test_parallel_executor_environment_loop() -> ParallelExecutorEnvironmentLoop:
+    """Pytest fixture for executor environment loop"""
+
+    class ExecutorStatsWrapperClass:
```
OmaymaMahjoub, line 56: How about making this class inherits from `EnvironmentLoopStatisticsBase` so that you won't need `#type: ignore` in line 63

OmaymaMahjoub (dismissed):
Looks good to me just a small optional change in line 56 if it is necessary!

tests/jax/components/building/environments_test.py
mmorris44: Thanks @OmaymaMahjoub. That's a good suggestion, but I think it might be quite involved, since I have to call to the __init__ of `EnvironmentLoopStatisticsBase`, which requires an actual environment loop. Happy to go for this if you'd like me to though?

OmaymaMahjoub: Oh, I see yeah it makes sense that it needs to be a similar initiator if we do that, yeah as you like we can keep it like that, other than that I think it is a good PR thank you @mmorris44 !

OmaymaMahjoub (approved):

siddarthsingh1 (approved):
All looks good @mmorris44 . Everything seems to be covered :smile: .

PR #618, opened 2022-07-05
Feature/local testing script

### LINKED ISSUE #616: [BUG] Running integration tests locally freezes laptop/PC due to overuse of resources
### Describe the bug
Running integration tests locally (via `make run-integration-tests`) freezes laptop/PC due to overuse of resources

### To Reproduce
Steps to reproduce the behaviour:
1.`make run-integration-tests`

### Expected behavior
Integration tests should not use all remaining resources and freeze system but rather complete as expected using a reasonable ampunt of compute piwer

### Context (Environment)
 - MAVA develop branch

### Possible Solution
change `N_CPU=$(grep -c ^processor /proc/cpuinfo)` is tests.sh to use fewer CPUs/ create a local_tests.sh for this

RuanJohn:
## What?
Made a local testing bash script for running all tests or all non-integration tests in a local docker container.
## Why?
The current testing bash script was using all available CPU cores on a machine leading to integration tests hanging/failing on some machines due to `launchpad` also trying to access multiple CPU cores. 
## How?
Made `local_tests.sh` script that uses all available CPU cores when all non-integration tests are run but uses less CPU cores when integration tests are run alongside all other tests. The `Makefile` as also been updated to run `local_tests.sh` when `make run-tests` or `make run-integration-tests` are called. 
## Extra
At the moment the solution is really basic and will just use `NUM_CPU * 0.75` CPU cores for running the integration test suite where `NUM_CPU` is the number of CPU cores available on a particular machine. 
closes #616

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,10 +77,10 @@ bash:
 	$(DOCKER_RUN) bash
 
 run-tests:
-	$(DOCKER_RUN) /bin/bash bash_scripts/tests.sh
+	$(DOCKER_RUN) /bin/bash bash_scripts/local_tests.sh
 
 run-integration-tests:
-	$(DOCKER_RUN) /bin/bash bash_scripts/tests.sh true
+	$(DOCKER_RUN) /bin/bash bash_scripts/local_tests.sh true
 
 run-checks:
 	$(DOCKER_RUN) /bin/bash bash_scripts/check_format.sh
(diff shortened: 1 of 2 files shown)


AsadJeewa (approved):
Looks good :smile: 
Very useful. My laptop no longer freezes!

sash-a (approved):
looks good :rocket:

PR #619, opened 2022-07-06
feat: match MAPPO performance

RuanJohn:
## What?
Updates to current Jax MAPPO implementation to match the Tensorflow MAPPO implementation. 
## Why?
We firstly want to be able to compare the Jax system against the Tensorflow system fairly and secondly we want the Jax implementation to match and ultimately exceed the Tensorflow system's performance. 
## How?
* Added flag for normalising advantages during training
* Added support for separate critic and policy networks with separate optimizers 
## Extra
The PR is quite large but a totally separate system `MAPPOSystemSeparateNetworks` as well as two new examples on the debug environment and smac were created. The main differences are the new separate networks components as well as the `PPOSeparateNetworks` class in `mava/systems/jax/mappo/networks.py`. The easiest way to find the new components is to look at all components ending in `SeparateNetworks` that are in the `MAPPOSystemSeparateNetworks` system design in `mava/systems/jax/mappo/system.py`.

Benchmarking is currently underway on the debug environment, 3s5z and Flatland just to make sure no bugs crept in after the final changes which were to merge cases where there were two for loops into one loop. 

The debug and 3s5z benchmarks have completed and the Flatland benchmark ran up to ~70% completion [due a problem on the cluster causing the experiments to terminate early]. To see the benchmarking results please [look here. ](https://www.notion.so/instadeep/Further-unnecessary-for-loop-removal-b0ccf9b90f9448e183b274a89f6c7755)


sash-a:
Linking a possible cause #628 (there is no masking of dead agents in value loss)

mava/components/jax/building/parameter_client.py
```diff
@@ -194,3 +194,171 @@ def on_building_trainer_parameter_client(self, builder: SystemBuilder) -> None:
...
         """Component type name, e.g. 'dataset' or 'executor'."""
         return "trainer_parameter_client"
+
+
+####################
+# SEPARATE NETWORKS
+###################
+
+
+class ExecutorParameterClientSeparateNetworks(BaseParameterClient):
```
DriesSmit, line 204: What is the issue we face when trying to implement a general ParameterClient that works with both network setups?

mava/components/jax/updating/parameter_server.py
```diff
@@ -204,3 +204,58 @@ def on_parameter_server_run_loop(self, server: SystemParameterServer) -> None:
             server.store.system_checkpointer.save()
             server.store.last_checkpoint_time = time.time()
             print("Updated variables checkpoint.")
+
+
+class ParameterServerSeparateNetworks(DefaultParameterServer):
```
DriesSmit, line 209: Same with the ParameterServer. What is the reason we can't build a generic one that works with both network setups? Is it that there would be too many if statements?

mava/components/jax/building/parameter_client.py
```diff
@@ -194,3 +194,171 @@ def on_building_trainer_parameter_client(self, builder: SystemBuilder) -> None:
...
+        """_summary_
+
+        Args:
+            builder : _description_
+        """
+        # Create policy parameters
+        params = {}
+        get_keys = []
+        net_type_key = "networks"
+        for agent_net_key in builder.store.networks[net_type_key].keys():
```
DriesSmit, line 227: E.g. could we not just specify the networks in a specific way (in the networks components) so that this can be one for loop over all networks?

mava/components/jax/updating/parameter_server.py
RuanJohn: I don't see there being any issues with doing this. The only reason I did it was to avoid if statements (and potential bugs) but very happy to change it so that we only have generic ones for both.

mava/components/jax/building/parameter_client.py
RuanJohn: I think we might even be able to already have this done in one loop as is.

mava/components/jax/updating/parameter_server.py
DriesSmit: I was wondering if it would be possible without it statements. I agree that if statements is what we are trying to avoid :)

RuanJohn: I couldn't find a nice way of doing this without adding in an extra if statement. But happy to make changes if you can spot a way :)

mava/components/jax/building/parameter_client.py
RuanJohn: Indeed, it is now in only one loop.

mava/components/jax/building/parameter_client.py
RuanJohn: Discussed below :)

siddarthsingh1 (approved):
Thanks @RuanJohn!

DriesSmit (approved):
Thanks, @RuanJohn :slightly_smiling_face: Let's get this one in and then solve the multiple parameter_server and data_server separately :+1:

mava/components/jax/building/parameter_client.py
DriesSmit: Great thanks!

mava/components/jax/updating/parameter_server.py
DriesSmit: It think it is fine for now :)

PR #621, opened 2022-07-07
chore: better store name for env spec

AsadJeewa:
## What?
Rename store environment spec
## Why?
For the variable to depict what it stores, readibility
## How?
Rename store environment spec
## Extra
Integration tests pass, examples still run


RuanJohn (approved):
:+1:

mmorris44 (approved):
Thank you for this @AsadJeewa! New name makes much more sense to me. Ran the jax tests and example: everything seems to be working fine.

PR #624, opened 2022-07-11
test: parameter client unit test 

### LINKED ISSUE #527: [TEST] Jax Parameter Client
### What do you want to test?
Jax paramater client components

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

RuanJohn:
## What?
Unit tests for Jax parameter client component. 
## Why?
Ongoing work to increase testing code coverage for Jax Mava redesign. 
## Extra
closes #527


tests/jax/components/building/parameter_client_test.py
```diff
@@ -0,0 +1,238 @@
...
+# limitations under the License.
+
+"""Unit tests for parameter client components"""
+
+from types import SimpleNamespace
+
+import pytest
+
+from mava.components.jax.building.parameter_client import (
+    ExecutorParameterClient,
```
OmaymaMahjoub, line 23: Missing `BaseParameterClient` test

tests/jax/components/building/parameter_client_test.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+
+    mock_builder = mock_builder_with_parameter_client
+    mock_builder.store.is_evaluator = False
+    exec_param_client = ExecutorParameterClient(
+        config=ExecutorParameterClientConfig(executor_parameter_update_period=100)
+    )
+    exec_param_client.on_building_executor_parameter_client(builder=mock_builder)
+
+    assert hasattr(mock_builder.store, "executor_parameter_client")
```
OmaymaMahjoub, line 74: I think this is unnecessary since in the next lines you  verify that `mock_builder.store.executor_parameter_client...` exists

tests/jax/components/building/parameter_client_test.py
```diff
@@ -0,0 +1,238 @@
...
+    assert mock_builder.store.executor_parameter_client._get_keys == [
```
OmaymaMahjoub, line 75: I don't know if this necessary but there are some missing verifications like: 
`mock_builder.store.executor_parameter_client._all_keys`
`mock_builder.store.executor_parameter_client._client`... (Check this file mava/systems/jax/parameter_client.py)

OmaymaMahjoub (changes requested):
Looks good to me, just small changes!

tests/jax/components/building/parameter_client_test.py
RuanJohn: Good catch! I initially thought they were not too important but it makes sense to include them.

tests/jax/components/building/parameter_client_test.py
RuanJohn: I initially left it out since it is only there to create count parameters, but I have now included an extra mock class and test case for it. Better to test than to not test.

tests/jax/components/building/parameter_client_test.py
OmaymaMahjoub: Sounds great at least the ones that are included as parameter when it initiate an instance from the class like ._all_keys... 🙌

OmaymaMahjoub (approved):
Looks good to me! Thank you @RuanJohn

tests/jax/components/building/parameter_client_test.py
```diff
@@ -0,0 +1,315 @@
...
+        "networks-network_agent_0",
+        "networks-network_agent_1",
+        "networks-network_agent_2",
+        "trainer_steps",
+        "trainer_walltime",
+        "evaluator_steps",
+        "evaluator_episodes",
+        "executor_episodes",
+        "executor_steps",
+    ]
```
sash-a, line 137: I think it might be a bit better to do:

```
expected_keys = set([
        "evaluator_episodes",
        "evaluator_steps",
        "executor_episodes",
        "executor_steps",
        "networks-network_agent_0",
        "networks-network_agent_1",
        "networks-network_agent_2",
        "trainer_steps",
        "trainer_walltime",
    ]
)

assert all([key in expected_keys for key in mock_builder.store.executor_parameter_client._all_keys])
assert all([key in expected_keys for key in mock_builder.store.executor_parameter_client._get_keys])
```

Unless I'm missing something but it seems like `get_keys` and `all_keys` are the same, but in a different order. So this should do the same, but it's a bit shorter and order agnostic.

tests/jax/components/building/parameter_client_test.py
```diff
@@ -0,0 +1,315 @@
...
+    mock_builder = mock_builder_with_parameter_client
+    mock_builder.store.is_evaluator = True
+    exec_param_client = ExecutorParameterClient(
+        config=ExecutorParameterClientConfig(executor_parameter_update_period=100)
+    )
+    exec_param_client.on_building_executor_parameter_client(builder=mock_builder)
+
+    assert hasattr(mock_builder.store, "executor_parameter_client")
+
+    assert mock_builder.store.executor_parameter_client._all_keys == [
```
sash-a, line 180: Everything below seems the same as above test. Do you need to retest it? I would imagine you're testing the same code, could you just remove it? If I'm wrong and it's different code generating the same data then feel free to ignore this

sash-a (changes requested):
I think you could clean this up a lot if you just had one big method that tests some kind of parameter client ie `def check_parameter_client(parameter_client):`. Where you put all the asserts that you are repeating in three of the test cases. And then you can pass into this function either `mock_builder.store.executor_parameter_client` or `mock_builder.store.trainer_parameter_client` and it should do all the asserts.

But otherwise than this it looks good! :rocket:

tests/jax/components/building/parameter_client_test.py
RuanJohn: Totally agree. This definitely cleans things up really well. I've done this for all test cases.

tests/jax/components/building/parameter_client_test.py
RuanJohn: It is indeed different (although slightly) since the below case is for an executor to be used for evaluation and the above was for a 'vanilla' executor.

RuanJohn:
> I think you could clean this up a lot if you just had one big method that tests some kind of parameter client ie `def check_parameter_client(parameter_client):`. Where you put all the asserts that you are repeating in three of the test cases. And then you can pass into this function either `mock_builder.store.executor_parameter_client` or `mock_builder.store.trainer_parameter_client` and it should do all the asserts.
> 
> But otherwise than this it looks good! rocket

Thanks for the feedback! You are right there are quite a few repeated asserts throughout the tests, but I think it still makes more sense to keep all the cases separate. There are 3 cases for the `ExecutorParameterClient` (executor with parameter client passed in, evaluator with a parameter client passed in and a case where no parameter client is passed in through the builder store) and there are 2 for the `TrainerParameterClient` (with and without a parameter client passed in through the builder store). 

It seems to me that having a big method would require multiple fixtures or multiple set ups inside the method which would add a bunch of code. If there is a cleaner way please let me know? Happy to change things :)

sash-a:
> > I think you could clean this up a lot if you just had one big method that tests some kind of parameter client ie `def check_parameter_client(parameter_client):`. Where you put all the asserts that you are repeating in three of the test cases. And then you can pass into this function either `mock_builder.store.executor_parameter_client` or `mock_builder.store.trainer_parameter_client` and it should do all the asserts.
> > But otherwise than this it looks good! rocket
> 
> Thanks for the feedback! You are right there are quite a few repeated asserts throughout the tests, but I think it still makes more sense to keep all the cases separate. There are 3 cases for the `ExecutorParameterClient` (executor with parameter client passed in, evaluator with a parameter client passed in and a case where no parameter client is passed in through the builder store) and there are 2 for the `TrainerParameterClient` (with and without a parameter client passed in through the builder store).
> 
> It seems to me that having a big method would require multiple fixtures or multiple set ups inside the method which would add a bunch of code. If there is a cleaner way please let me know? Happy to change 

I think the change you made with extracting the expected variables has cleaned this up a whole lot. I'm happy to keep it the way it is :rocket:

sash-a (approved):
:1st_place_medal:

OmaymaMahjoub (approved):

PR #626, opened 2022-07-12
Test/jax components datasets unit test

### LINKED ISSUE #522: [TEST] Jax Datasets
### What do you want to test?
Jax dataset components

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

OmaymaMahjoub:
## What?
Test: Unit tests for Jax datasets components
## Why?
## How?
Test `TransitionDataset` and `TrajectoryDataset` components from `mava/components/jax/building/datasets.py`
## Extra
Close #522

### DIFF
--- a/tests/jax/components/building/datasets_test.py
+++ b/tests/jax/components/building/datasets_test.py
@@ -1,43 +1,52 @@
 from types import SimpleNamespace
+
 import pytest
+import reverb
+from reverb import item_selectors, rate_limiters
+
 from mava.components.jax.building.datasets import TransitionDataset
 from mava.systems.jax.builder import Builder
-from reverb import item_selectors, rate_limiters
-import reverb
+
 
 class MockBuilder(Builder):
     def __init__(self) -> None:
-        simple_server=reverb.Server(tables=[reverb.Table(
-                name="table_0",
-                sampler=item_selectors.Prioritized(priority_exponent=1),
-                remover=item_selectors.Fifo(),
-                max_size=1000,
-                rate_limiter=rate_limiters.MinSize(1),
-            )
-        ],port=2000)
+        simple_server = reverb.Server(
+            tables=[
+                reverb.Table(
+                    name="table_0",
+                    sampler=item_selectors.Prioritized(priority_exponent=1),
+                    remover=item_selectors.Fifo(),
+                    max_size=1000,
+                    rate_limiter=rate_limiters.MinSize(1),
+                )
+            ],
+            port=2000,
+        )
         """ https://colab.research.google.com/github/deepmind/reverb/blob/master/examples/demo.ipynb#scrollTo=SQFSZJkyroFX """
-        data_server_client=SimpleNamespace(server_address=f'localhost:{simple_server.port}')
-        trainer_id= "table_0"
+        data_server_client = SimpleNamespace(
+            server_address=f"localhost:{simple_server.port}"
+        )
+        trainer_id = "table_0"
         self.store = SimpleNamespace(
-            data_server_client=data_server_client,
-            trainer_id=trainer_id)
-        
+            data_server_client=data_server_client, trainer_id=trainer_id
+        )
+
 
 @pytest.fixture
-def mock_builder()->MockBuilder:
+def mock_builder() -> MockBuilder:
     return MockBuilder()
 
-def test_init_transition_dataset()->None:
-    transition_dataset=TransitionDataset()
 
-    assert transition_dataset.config.sample_batch_size== 256
-    assert transition_dataset.config.prefetch_size==None
-    assert transition_dataset.config.num_parallel_calls==12
-    assert transition_dataset.config.max_in_flight_samples_per_worker==None
-    assert transition_dataset.config.postprocess==None
+def test_init_transition_dataset() -> None:
+    transition_dataset = TransitionDataset()
 
-def test_on_building_trainer_dataset(mock_builder:MockBuilder):
-    transition_dataset=TransitionDataset()
+    assert transition_dataset.config.sample_batch_size == 256
+    assert transition_dataset.config.prefetch_size == None
+    assert transition_dataset.config.num_parallel_calls == 12
+    assert transition_dataset.config.max_in_flight_samples_per_worker == None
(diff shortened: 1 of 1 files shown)


tests/jax/components/building/datasets_test.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Args:
+        mock_builder: Builder
+    """
+    trajectory_dataset = TrajectoryDataset()
+    trajectory_dataset.on_building_trainer_dataset(builder=mock_builder)
+
+    assert mock_builder.store.sample_batch_size == 256
+
+    dataset = mock_builder.store.dataset_iterator._iterator._dataset
```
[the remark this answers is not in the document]
AsadJeewa: This will also need to be made clearer when we add comments to the Datasets class

tests/jax/components/building/datasets_test.py
```diff
@@ -0,0 +1,145 @@
...
+) -> None:
+    """Test on_building_trainer_dataset of TransitionDataset Component
+
+    Args:
+        mock_builder: Builder
+    """
+    transition_dataset = TransitionDataset()
+    transition_dataset.on_building_trainer_dataset(builder=mock_builder)
+
+    dataset = mock_builder.store.dataset._dataset._map_func._func(1)._dataset
```
AsadJeewa, line 94: Please add a comment here to explain how this gives us the dataset via TF. Same for `TrajectoryDataset()`

AsadJeewa (changes requested):
Thanks, @OmaymaMahjoub. A few changes requested

tests/jax/components/building/datasets_test.py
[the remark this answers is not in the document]
OmaymaMahjoub: I didn't get which one should we change it to dataset_iterator so I will be thankful if you can change it! Also about the question that you have I am curious to know why too!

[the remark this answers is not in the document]
AsadJeewa: I made a commit to change the transition dataset store value to dataset iterator. We are not currently using the dataset and will probably do so in the future. Will investigate (it comes from TF) but that is not related to this PR

AsadJeewa:
Thanks, Omayma. I made a small commit as per our discussion to fix the pre-commit and change the var name (as per your comment. One more small change requested

tests/jax/components/building/datasets_test.py
```diff
@@ -0,0 +1,192 @@
...
+    max_in_flight_samples_per_worker: int = 1024
+    num_workers_per_iterator: int = -2
+    max_samples_per_stream: int = -2
+    rate_limiter_timeout_ms: int = -2
+    get_signature_timeout_secs: Optional[int] = None
+
+
+def test_init_transition_dataset() -> None:
+    """Test initiator of TransitionDataset component"""
+    transition_dataset = TransitionDataset(config=TransitionDatasetConfigTest)  # type: ignore
```
AsadJeewa, line 94: Please change this to a fixture (as we did in other tests such as the reverb_compoents_test) then it can be of type `TrajectoryDatasetConfig` and you can remove the type: ignore and make it multiline so that it is not > 88 characters as per the pre-commit checks

tests/jax/components/building/datasets_test.py
```diff
@@ -0,0 +1,192 @@
...
+            transition_dataset.config.num_parallel_calls,
+            dtype=dtypes.int64,
+            name="num_parallel_calls",
+        )
+    )
+
+
+def test_init_trajectory_dataset() -> None:
+    """Test initiator of TrajectoryDataset component"""
+    trajectory_dataset = TrajectoryDataset(config=TrajectoryDatasetConfigTest)  # type: ignore
```
AsadJeewa, line 141: same as above

AsadJeewa (changes requested):
Thanks, Omayma. I made a small commit as per our discussion to fix the pre-commit and change the var names (as per your comment). I also shortened the comments, as we discussed. 
One more small change requested

tests/jax/components/building/datasets_test.py
OmaymaMahjoub: The error from mypy here is not about the length of the line but it is for `Argument "config" to "TransitionDataset" has incompatible type "TransitionDatasetConfigTest"; expected "TransitionDatasetConfig"` which means that the dataconfig we gave should be of type TransitionDatasetConfig.
I will try to make mock and update the values from TransitionDatasetConfig inside it!

AsadJeewa (dismissed):
Thanks for the changes @OmaymaMahjoub 
Happy on my side

tests/jax/components/building/datasets_test.py
```diff
@@ -0,0 +1,201 @@
...
+
+    dataset = mock_builder.store.dataset_iterator._dataset._map_func._func(1)._dataset
+    assert (
+        dataset._input_dataset._server_address
+        == mock_builder.store.data_server_client.server_address
+    )
+    assert dataset._input_dataset._table == mock_builder.store.trainer_id
+    assert dataset._batch_size == transition_dataset.config.sample_batch_size
+    assert (
+        dataset._input_dataset._max_in_flight_samples_per_worker
```
KaleabTessera, line 137: I don't think we can always assume `_max_in_flight_samples_per_worker = 2 * transition_dataset.config.sample_batch_size`.

OmaymaMahjoub: Yeah, you were right it is the case where max_in_flight_sample_per_worker=None 
```
if max_in_flight_samples_per_worker is None and batch_size is None:
    max_in_flight_samples_per_worker = 100
  elif max_in_flight_samples_per_worker is None:
    max_in_flight_samples_per_worker = 2 * batch_size
```
I will make 3 cases for it (batch_size and non max_in.., non batch_size and non max_in, max_in_flight non none)

OmaymaMahjoub: Done!

KaleabTessera (approved):
Thanks @OmaymaMahjoub :fire: :fire:

AsadJeewa (approved):
Thanks @OmaymaMahjoub

PR #627, opened 2022-07-12
Test/jax core parameter client unit tests

### LINKED ISSUE #547: [TEST] Jax System Core Parameter Client
### What do you want to test?
Jax system core parameter client (in `mava/systems/jax/`)

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

RuanJohn:
## What?
Unit tests for core parameter client
## Why?
Ongoing efforts to increase testing code coverage for callback redesign. 
## Extra
closes #547


tests/jax/core_system_components/parameter_client_test.py
```diff
@@ -15,4 +15,406 @@
...
+    }
+
+    assert parameter_client._set_get_call_counter == 0
+
+    parameter_client.set_and_get_async()
+
+    assert parameter_client._set_get_future is None
+
+
+def test_add_async(parameter_client: ParameterClient) -> None:
```
mmorris44, line 317: I can't find a test here for what happens when `self._add_future is None`

tests/jax/core_system_components/parameter_client_test.py
```diff
@@ -15,4 +15,406 @@
...
+
+    parameter_client.add_async(params={"new_key_1": 1})
+
+    assert parameter_client._async_add_buffer == {"new_key_1": 2}
+
+    # set add_future to dummy variable other than none
+    # assert that new parameter is added to async buffer
+    parameter_client._add_future._state = "not_done"  # type: ignore
+    parameter_client._async_add_buffer = {}
+    parameter_client.add_async(params={"new_key_2": 1})
```
mmorris44, line 336: This seems to test again what was just tested

tests/jax/core_system_components/parameter_client_test.py
```diff
@@ -15,4 +15,406 @@
...
+                "networks-network_key_0": {
+                    "layer_0": {"weights": "new_weights", "biases": "new_biases"}
+                }
+            },
+        )
+
+
+def test_copy_array_with_device(parameter_client: ParameterClient) -> None:
+    """Test that new parameters are set on a device when a device \
+        is give."""
```
mmorris44, line 388: given

mmorris44 (changes requested):
Very thorough and well-written PR, thanks @RuanJohn!
Just a few minor comments.

tests/jax/core_system_components/parameter_client_test.py
```diff
@@ -1,5 +1,5 @@
 # python3
-# Copyright 2021 InstaDeep Ltd. All rights reserved.
+# Copyright 2022 InstaDeep Ltd. All rights reserved.
```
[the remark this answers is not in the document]
RuanJohn: Sorted.

tests/jax/core_system_components/parameter_client_test.py
RuanJohn: By default `self._add_future` is set to `None`, please see [here](https://github.com/instadeepai/Mava/blob/9a0835a4998d17cab580a7269443c12cc16cf6d1/mava/systems/jax/parameter_client.py#L87). So we technically check for that condition in lines 319 - 322. I have also added an extra assert to verify that `self._add_future.done()` is `True` after adding the parameters. I can also explicitly set the value of `self.add_future` to `None` in the test if you think that would make it more explicit?

mmorris44: Ah I see, no worries then!
Maybe just add an assert before calling `add_async` that `self._add_future is None`, to ensure that the default is set correctly?

tests/jax/core_system_components/parameter_client_test.py
mmorris44: Keep both of them. Your explanation makes sense

tests/jax/core_system_components/parameter_client_test.py
RuanJohn: Added it :call_me_hand:

tests/jax/core_system_components/parameter_client_test.py
```diff
@@ -15,4 +15,410 @@
...
+import jax
+import numpy as np
+import pytest
+
+from mava.callbacks.base import Callback
+from mava.systems.jax.parameter_client import ParameterClient
+from mava.systems.jax.parameter_server import ParameterServer
+
+
+class MockParameterServer(ParameterServer):
```
sash-a, line 31: Too late now, but it's for things like this where I wonder if we shouldn't have used monkeypatch

tests/jax/core_system_components/parameter_client_test.py
```diff
@@ -15,4 +15,410 @@
...
+    assert parameter_client._parameters == {
+        "key_0": np.array(1, dtype=np.int32),
+        "key_1": np.array(2, dtype=np.float32),
+        "key_2": np.array(3, dtype=np.int32),
+        "key_3": np.array(4, dtype=np.int32),
+        "key_4": np.array(5, dtype=np.int32),
+        "networks-network_key_0": {"layer_0": {"weights": 1, "biases": 1}},
+        "networks-network_key_1": {"layer_0": {"weights": 2, "biases": 2}},
+        "networks-network_key_2": {"layer_0": {"weights": 3, "biases": 3}},
+        "allkey_0": np.array(0, dtype=np.int32),
```
sash-a, line 307: You expect this 3 times maybe extract into `expected_get_params`? But I can see an argument to leave it as is too.

sash-a (approved):
Super clear tests @RuanJohn, was a very easy review! Only 1 minor change, but I don't think it's strictly necessary so you can leave it as is.

tests/jax/core_system_components/parameter_client_test.py
RuanJohn: I think it is a bit easier to leave it as is for now.

sash-a (approved):
:rocket:

mmorris44 (approved):
Re-approved. Still happy! Thanks @RuanJohn

PR #630, opened 2022-07-14
Test/jax components/training/step

### LINKED ISSUE #536: [TEST] Jax Trainer Step
### What do you want to test?
Jax step components

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

OmaymaMahjoub:
## What?
Test: Unit tests for Jax train step components
## Why?
## How?
Make unit test for `DefaultTrainerStep` and `MAPGWithTrustRegionStep `components from `mava/components/jax/training/step.py` file
## Extra
Close #536


tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,475 @@
...
+
+@pytest.fixture
+def mock_trainer() -> MockTrainer:
+    """Build fixture from MockTrainer"""
+    return MockTrainer()
+
+
+@pytest.fixture
+def dummy_sample() -> DummySample:
+    """Build fixture from DUmmySample"""
```
AsadJeewa, line 345: small u Dummy

OmaymaMahjoub: Done thanks!

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,475 @@
...
+
+    def __init__(self) -> None:
+        self.data = SimpleNamespace(
+            observations={
+                "agent_0": OLT(
+                    observation=jnp.array(
+                        [
+                            [
+                                [0.1, 0.5, 0.7, 0.1, 0.5, 0.7],
+                                [0.1, 0.5, 0.7, 0.1, 0.56, 0.87],
```
RuanJohn, line 154: If `sequence_length = 3` we need an extra observation for here. For example it should be: 
```
"agent_0": OLT(
                    observation=jnp.array(
                        [
                            [
                                [0.1, 0.5, 0.7, 0.1, 0.5, 0.7],
                                [0.1, 0.5, 0.7, 0.1, 0.56, 0.87], 
                                [0.2, 0.3, 0.4, 0.5, 0.6, 0.7], 
                            ],
                            [
                                [0.1, 0.5, 0.7, 0.1, 0.5, 0.7],
                                [0.1, 0.5, 0.7, 0.13, 0.5, 0.78],
                                [0.2, 0.3, 0.4, 0.5, 0.6, 0.7], 
                            ],
                        ]
                    ),
```
Similarly for the other observations.

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,475 @@
...
+                            [
+                                [1, 1, 1, 1, 0, 0],
+                                [0, 1, 1, 0, 1, 0],
+                            ],
+                            [
+                                [0, 0, 0, 1, 0, 1],
+                                [1, 1, 1, 0, 1, 0],
+                            ],
+                        ]
+                    ),
```
RuanJohn, line 173: Similarly `legal_actions` should also be slightly extended please 
```
                    legal_actions=jnp.array(
                        [
                            [
                                [1, 1, 1, 1, 0, 0],
                                [0, 1, 1, 0, 1, 0],
                                [1, 1, 1, 1, 1, 1],
                            ],
                            [
                                [0, 0, 0, 1, 0, 1],
                                [1, 1, 1, 0, 1, 0],
                                [1, 1, 1, 1, 1, 1],
                            ],
                        ]
                    ),
```

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,475 @@
...
+                                [0, 0, 0, 0, 0, 1],
+                                [0, 1, 0, 0, 0, 0],
+                            ],
+                            [
+                                [0, 1, 1, 0, 1, 1],
+                                [0, 0, 0, 1, 1, 0],
+                            ],
+                        ]
+                    ),
+                ),
```
RuanJohn, line 186: From what I have seen from the debug environment samples it seems that `terminal` should instead take on this form 
```
terminal=jnp.array([
                           [[0.],
                            [0.],
                            [0.]],
                            [[0.],
                             [0.],
                             [0.]]
                      ], )),
```

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,475 @@
...
+            },
+            actions={
+                "agent_0": jnp.array(
+                    [
+                        [[1, 2, 1, 1, 3, 1], [4, 2, 0, 0, 0, 0]],
+                        [[1, 1, 1, 1, 2, 1], [1, 3, 0, 4, 0, 0]],
+                    ]
+                ),
```
RuanJohn, line 270: Again, according to the output data from the example on the debug environment it seems that `actions` should take on the following form 
```
"agent_0": jnp.array(
                    [
                        [[1, 2, 1,], [4, 2, 0, ]],
                        [[1, 1, 1,], [1, 3, 0,]],
                        [[1, 2, 1], [2, 1, 0,]], 
                    ]
                ),
```

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,475 @@
...
+                "agent_2": jnp.array(
+                    [
+                        [[1, 2, 1, 4, 0, 1], [0, 0, 0, 3, 0, 0]],
+                        [[1, 1, 1, 2, 0, 1], [0, 3, 0, 0, 0, 0]],
+                    ]
+                ),
+            },
+            rewards={
+                "agent_0": jnp.array([[0.96, 0.34], [0.22, 0.34]]),
```
RuanJohn, line 285: `rewards` just needs an extra value:
```
"agent_0": jnp.array([[0.96, 0.34,0.67], [0.22, 0.34, 0.54]]),
```

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,475 @@
...
+                "agent_1": jnp.array([[0.22, 0.33], [0.25, 0.84]]),
+                "agent_2": jnp.array([[0.43, 0.45], [0.43, 0.25]]),
+            },
+            discounts={
+                "agent_0": jnp.array([[1, 0, 0, 1, 1, 0], [1, 0, 0, 1, 1, 1]]),
```
RuanJohn, line 290: `discounts` should take on a simialr shape to `rewards` and `actions`:
```
"agent_0": jnp.array([[1., 1., 1.],[1., 1., 1.]]),
```

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,475 @@
...
+                "agent_2": jnp.array([[1, 0, 0, 1, 1, 1], [0, 0, 0, 1, 0, 1]]),
+            },
+            extras={
+                "policy_info": {
+                    "agent_0": jnp.array(
+                        [
+                            [
+                                [-1.23, -1.3, -2.45, -1.45, -0.21, -1.34],
+                                [-2.11, -0.55, -1.17, -1.11, -1.15, -1.57],
+                            ],
```
RuanJohn, line 301: and then `policy_info` should as well be for the shape: 
```
'agent_0': jnp.array([[-1.63, -1.78, -1.44], [-1.69, -1.48, -1.66  ]]), 
```

RuanJohn (changes requested):
Thank you for all the hard work @OmaymaMahjoub! Some minor comments about the shapes of the data so that they mirror the data in the ppo examples with a similar `sequence_length` and `sample_batch_size`.

OmaymaMahjoub:
> 

Thank you @RuanJohn for the reviewing efforts I will try to make all the necessary changes based on your suggestions!

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,723 @@
...
+            opt_states=opt_states,
+            key=jax.random.PRNGKey(5),
+            num_minibatches=1,
+            epoch_update_fn=epoch_update,
+            num_epochs=2,
+        )
+        self.store = store
+
+
+class DummySample:
```
RuanJohn, line 158: Thank you for all these updates! Just because the dummy sample class is now so big can we maybe move this to a separate file in the same directory and import it from there to make the test a bit cleaner to read?

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,723 @@
...
+    def __init__(self) -> None:
+        self.data = SimpleNamespace(
+            observations={
+                "agent_0": OLT(
+                    observation=array(
```
RuanJohn, line 163: Also, a minor detail but could we please have all these arrays be `jnp.array` please?

OmaymaMahjoub: I import array from jax.numpy to more clarification I will change it to jnp.array

RuanJohn: Thank you :)

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,723 @@
...
+def test_on_training_init_start(mock_trainer: MockTrainer) -> None:
+    """Test on_training_init_start method from MAPGWITHTrustRegionStep component"""
+    mapg_with_trust_region_step = MAPGWithTrustRegionStep()
+    mapg_with_trust_region_step.on_training_init_start(trainer=mock_trainer)
+
+    assert (
+        mock_trainer.store.full_batch_size
+        == mock_trainer.store.sample_batch_size
+        * (mock_trainer.store.sequence_length - 1)
+    )
```
RuanJohn, line 666: Can we please set this to an expected value? Because it seems like it using the code as used in the component

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,723 @@
...
+    assert trainer_step.config.random_key == 42
+
+
+def test_on_training_step_with_timestamp(mock_trainer: Trainer) -> None:
+    """Test on_training_step method from TrainerStep case of existing timestamp"""
+    trainer_step = DefaultTrainerStep()
+    old_timestamp = mock_trainer.store.timestamp
+    trainer_step.on_training_step(trainer=mock_trainer)
+
+    assert int(mock_trainer.store.timestamp) == int(time.time())
```
RuanJohn, line 611: Is there not a way to rather just check that the timestamp value has increased? I am concerned that the test might break due the `time.time()` somehow not being able to align with the timestamp.

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,723 @@
...
+    assert float(list(metrics["rewards_mean"].values())[1]) == 0.08261970430612564
+    assert float(list(metrics["rewards_mean"].values())[2]) == 0.07156813144683838
+
+    assert list(metrics["rewards_std"].keys()) == ["agent_0", "agent_1", "agent_2"]
+    assert float(list(metrics["rewards_std"].values())[0]) == 0.0
+    assert float(list(metrics["rewards_std"].values())[1]) == 0.07048596441745758
+    assert float(list(metrics["rewards_std"].values())[2]) == 0.07740379124879837
+
+    random_key, _ = jax.random.split(old_key)
+    assert list(mock_trainer.store.key) == list(random_key)
```
RuanJohn, line 717: We could just assert that new key is different from the old key here

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,723 @@
...
+        "rewards_mean",
+        "rewards_std",
+    ]
+
+    updates = {
+        net_key: mock_trainer.store.networks["networks"][net_key].params
+        for net_key in mock_trainer.store.networks["networks"].keys()
+    }
+
+    assert metrics["norm_params"] == optax.global_norm(jax.tree_leaves(updates))
```
RuanJohn, line 700: I think we could assert for a value here as well.

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,723 @@
...
+    assert float(list(metrics["rewards_std"].values())[2]) == 0.07740379124879837
+
+    random_key, _ = jax.random.split(old_key)
+    assert list(mock_trainer.store.key) == list(random_key)
+
+    assert mock_trainer.store.opt_states == {
+        "network_agent_0": 0,
+        "network_agent_1": 1,
+        "network_agent_2": 2,
+    }
```
RuanJohn, line 723: I know this would be quite a bit of work, but an important part of the step component is that the network parameters and the optimizer states get updated/changed. I think a way we could do this is to update the mock epoch update function to not return the same carry as was passed in but updating the parameters and states. Even if we just increment everything by 1 each time the epoch update is called. That way we can verify that parameters and states are being update and that the epoch update funciton is getting called the correct number of times.

tests/jax/components/training/step_test.py
OmaymaMahjoub: Another solution could be just to check it is bigger then old_timestamp, I will try to make it instead of time.time

tests/jax/components/training/sample.py
```diff
@@ -0,0 +1,445 @@
+from types import SimpleNamespace
```
RuanJohn, line 1: Please update the name to be `step_test_sample.py` and add the licence to the file :)

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,284 @@
...
+
+    assert list(mock_trainer.store.trainer_parameter_client.params.keys()) == [
+        "trainer_steps",
+        "trainer_walltime",
+    ]
+    assert mock_trainer.store.trainer_parameter_client.params["trainer_steps"] == 1
+    elapsed_time = int(time.time() - old_timestamp)
+    assert (
+        int(mock_trainer.store.trainer_parameter_client.params["trainer_walltime"])
+        == elapsed_time
```
RuanJohn, line 189: This might run into the same `time.time()` issue as the previous check could have. Is there another way of checking this as well?

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,284 @@
...
+    assert float(list(metrics["rewards_std"].values())[0]) == 0.0
+    assert float(list(metrics["rewards_std"].values())[1]) == 0.07048596441745758
+    assert float(list(metrics["rewards_std"].values())[2]) == 0.07740379124879837
+
+    assert list(mock_trainer.store.key) != list(old_key)
+
+    assert sorted(list(mock_trainer.store.opt_states.keys())) == sorted(
+        ["network_agent_0", "network_agent_1", "network_agent_2"]
+    )
+    assert mock_trainer.store.opt_states != {
```
RuanJohn, line 280: Since we know how many epochs we are updating for and what the updates should be doing we can check for these values directly to be 
```
assert mock_trainer.store.opt_states != {
        "network_agent_0": 2,
        "network_agent_1": 3,
        "network_agent_2": 4,
    }
```
SImilarly, we could also be able check directly for the network parameters values in `mock_trainer.store.networks`

tests/jax/components/training/step_test.py
OmaymaMahjoub: We can change the assert to check that is different of zero

tests/jax/components/training/step_test.py
OmaymaMahjoub: the problem that the function change the order of the keys and values I will try to figure out how to sort it else  I will just put what the output exactly look like

OmaymaMahjoub: Actually the assert works thanks!
```
assert mock_trainer.store.opt_states == {
        "network_agent_0": 2,
        "network_agent_1": 3,
        "network_agent_2": 4,
    }
```

tests/jax/components/training/step_test.py
RuanJohn: Sounds good.

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,284 @@
...
+    assert float(list(metrics["rewards_std"].values())[0]) == 0.0
+    assert float(list(metrics["rewards_std"].values())[1]) == 0.07048596441745758
+    assert float(list(metrics["rewards_std"].values())[2]) == 0.07740379124879837
+
+    assert list(mock_trainer.store.key) != list(old_key)
+
+    for i, net_key in enumerate(mock_trainer.store.networks["networks"]):
+        assert jnp.array_equal(
+            mock_trainer.store.networks["networks"][net_key].params["key"],
+            jnp.array([i + 2.0, i + 2.0, i + 2.0]),
```
RuanJohn, line 277: Instead of 2 here we could have the numner of updating epochs. This way we are checking two things at once and not hardcoding a value.

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,284 @@
...
+        )
+
+    assert mock_trainer.store.opt_states == {
+        "network_agent_0": 2,
+        "network_agent_1": 3,
+        "network_agent_2": 4,
+    }
```
RuanJohn, line 284: Similarly, these are the original values that have been incremented `mock_trainer.store.num_epochs * mock_trainer_store.num_minibatches` times. Is there are way to check this? Then we again verify that step was called the correct number of times.

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,290 @@
...
+    assert float(list(metrics["rewards_mean"].values())[1]) == 0.08261970430612564
+    assert float(list(metrics["rewards_mean"].values())[2]) == 0.07156813144683838
+
+    assert list(metrics["rewards_std"].keys()) == ["agent_0", "agent_1", "agent_2"]
+    assert float(list(metrics["rewards_std"].values())[0]) == 0.0
+    assert float(list(metrics["rewards_std"].values())[1]) == 0.07048596441745758
+    assert float(list(metrics["rewards_std"].values())[2]) == 0.07740379124879837
+
+    assert list(mock_trainer.store.key) != list(old_key)
+    value = mock_trainer.store.num_epochs * mock_trainer.store.num_minibatches
```
RuanJohn, line 273: Can we change the name here to be more descriptive? eg. `value` -> `num_expected_update_steps`?

OmaymaMahjoub: I was looking for it name and I couldn't find thanks!

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,292 @@
...
+
+def test_step(mock_trainer: MockTrainer, dummy_sample: DummySample) -> None:
+    """Test step function"""
+    mapg_with_trust_region_step = MAPGWithTrustRegionStep()
+    del mock_trainer.store.step_fn
+    mapg_with_trust_region_step.on_training_step_fn(trainer=mock_trainer)
+    old_key = mock_trainer.store.key
+    with jax.disable_jit():
+        metrics = mock_trainer.store.step_fn(dummy_sample)
+
```
RuanJohn, line 247: Maybe we can just have some comments in each section of the test make it easier to read later on. So here we could have `# Check that metrics were correclty computed`

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,292 @@
...
+    assert list(metrics["rewards_mean"].keys()) == ["agent_0", "agent_1", "agent_2"]
+    assert float(list(metrics["rewards_mean"].values())[0]) == 0.0
+    assert float(list(metrics["rewards_mean"].values())[1]) == 0.08261970430612564
+    assert float(list(metrics["rewards_mean"].values())[2]) == 0.07156813144683838
+
+    assert list(metrics["rewards_std"].keys()) == ["agent_0", "agent_1", "agent_2"]
+    assert float(list(metrics["rewards_std"].values())[0]) == 0.0
+    assert float(list(metrics["rewards_std"].values())[1]) == 0.07048596441745758
+    assert float(list(metrics["rewards_std"].values())[2]) == 0.07740379124879837
+
```
RuanJohn, line 271: Similarly, `# check that trainer random key has been updated`

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,292 @@
...
+    assert list(mock_trainer.store.key) != list(old_key)
+    num_expected_update_steps = (
```
RuanJohn, line 273: Similarly, `# check that network parameters and optimzer states were updated the correct number of times`

tests/jax/components/training/step_test.py
OmaymaMahjoub: Maybe a better idea I will look if there is a way when a test fail it gives you a message like py files with try and except

tests/jax/components/training/step_test.py
```diff
@@ -0,0 +1,292 @@
...
+    assert callable(mock_trainer.store.step_fn)
+
+
+def test_step(mock_trainer: MockTrainer, dummy_sample: DummySample) -> None:
+    """Test step function"""
+    mapg_with_trust_region_step = MAPGWithTrustRegionStep()
+    del mock_trainer.store.step_fn
+    mapg_with_trust_region_step.on_training_step_fn(trainer=mock_trainer)
+    old_key = mock_trainer.store.key
+    with jax.disable_jit():
```
DriesSmit, line 245: Just out of curiosity. Why is JIT disabled here?

OmaymaMahjoub: because before step function we have @jit and this made the dummy sample we have made which is the input of the step function not being accepted and I had this error:

RuanJohn (approved):
Thank you for all the work @OmaymaMahjoub! All good from my side :muscle:

tests/jax/components/training/step_test.py
DriesSmit: Ah, I see. Should we not rather try and fix the dummy sample then? As the real system uses jit and therefore we are slightly misaligned with our test here.

DriesSmit (commented):
Thanks, @OmaymaMahjoub :fire:  The code looks good! Just see my one comment.

tests/jax/components/training/step_test.py
DriesSmit: Or do we test the jitted step function at some later stage?

OmaymaMahjoub: I tried that before and I couldn't find any solution 😓 I will try to look for further solutions!

OmaymaMahjoub: nope we didn't however, good updates I find a way to run samples without jit thanks to the data sample that @RuanJohn send it previously! Still have small issue I will try to fix it and commit

OmaymaMahjoub: Done! Thanks for the help

DriesSmit: Great, thanks @OmaymaMahjoub !

DriesSmit (approved):

RuanJohn (approved):
Thank you for all the work @OmaymaMahjoub!

PR #632, opened 2022-07-14
Tests/core system unit tests

### LINKED ISSUE #549: [TEST] Jax System Core System
### What do you want to test?
Jax system core system (in `mava/systems/jax/`)

### Outline of test structure
* Unit tests
* Test components and hooks

### Definition of done

RuanJohn:
## What?
Unit tests for core `System` component.
## Why?
Part of ongoing work to increase testing code coverage for Jax redesign Mava. 
## Extra
Expanded existing unit tests. 
closes #549


tests/jax/core_system_components/system_component_test.py
```diff
@@ -282,6 +283,26 @@ def test_system_launch_with_build(
     assert system_with_two_components._builder.store.int_plus_str == 3
     assert system_with_two_components._builder.store.float_plus_bool == 1.2
     system_with_two_components.launch()
+    assert system_with_two_components._builder.store.global_config == SimpleNamespace(
+        multi_process=True,
```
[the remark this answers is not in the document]
RuanJohn: Good catch. Sorted. It now tests for the defaults from the the defaults dataclasses unless the parameters are overwritten. Just one thing to note, I have to instantiate the `DistributorDefaultConfig` in order to get access to the `nodes_on_gpu` attribute. I think it is because its default is given using `field`. Happy to change it to just check for an empty list instead.

siddarthsingh1 (changes requested):
Looks good @RuanJohn. Just one of the function descriptions looks a bit odd.

tests/jax/core_system_components/system_component_test.py
```diff
@@ -296,6 +323,26 @@ def test_system_update_with_existing_component(
...
+        param_3=ComponentOneDefaultConfig.param_3,
+        param_4=ComponentTwoDefaultConfig.param_4,
+        param_5=ComponentTwoDefaultConfig.param_5,
+    )
+
+
+def test_system_update_when_built(
+    system_with_two_components: System,
+) -> None:
+    """Test that exception is raised when system is updated are being built."""
```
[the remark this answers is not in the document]
RuanJohn: Fixed it :)

tests/jax/core_system_components/system_component_test.py
```diff
@@ -272,16 +278,37 @@ def system_with_two_components() -> System:
...
+    )
+
+
+def test_launch_when_not_built(
+    system_with_two_components: System,
+) -> None:
+    """Test that exception is thrown when launch is called before a system\
+        has been built."""
+
+    with pytest.raises(Exception):
```
sash-a, line 310: 2 things here, in general I think you should expect a more specific exception, but for this case I think we raise a regular `Exception` which I think would be greatly improved if we made a couple custom exceptions:

* `DuplicateConfigError`
* `NotYetBuiltError`
* `AlreadyBuiltError`

Not sure if you want to add it to this PR or make it in a separate PR, but these would likely be pretty common issues for new users and would be nice to have good error messages for them

RuanJohn: Thanks for the feedback! In the code we are just raising regular Exeptions at the moment, hench the generic checks in the test. I agree with the custom exceptions idea and think it would be an improvement but I think a separate PR for that is probably a better idea. What do you think?

siddarthsingh1 (approved):
I'm happy with to merge the PR @RuanJohn . :+1:

tests/jax/core_system_components/system_component_test.py
sash-a: Ye they are very descriptive, so maybe it's fine to just leave it. I'm honestly not sure what python convention is, maybe if we keep descriptive error messages then it's fine. Custom exceptions do seem quite _java-like_ so maybe we just leave it

sash-a (approved):
Woops meant to approve this when I replied this morning :rocket:

KaleabTessera (approved):
Thanks @RuanJohn!

Issue #631, opened 2022-07-14
[FEATURE] Continuous IPPO


siddarthsingh1:
### Feature
Implementation of continuous IPPO algorithm into Jax Mava. This will allow the Jax version of Mava to be used in continuous settings like PettingZoo's Multi-Walker environment.

### Proposal
- Add `make_continuous_networks` function to `class PPONetworks`.

### Testing
Testing is the same as discrete IPPO.

### Benchmarking (Optional)
Describe how and why this feature needs to be benchmarked with regards to:
* Comparable performance to the TF continuous PPO implementation in settings that are not the debug environment.

### Definition of done
Continuous IPPO is able to successfully solve environments to a similar level as the TF implementation.

### Mandatory checklist before making a PR
* [ ] The success criteria laid down in “Definition of done” are met.
* [ ] Code is documented - docstrings for methods and classes, static types for arguments.
* [ ] Code is tested - unit, integration and/or functional tests are added.
* [ ] Documentation is updated - README, CONTRIBUTING, or other documentation.
* [ ] All functional tests are green.
* [ ] Link experiment/benchmarking after implementation (optional).

PR #635, opened 2022-07-19
Feature/make continuous ppo

siddarthsingh1:
## What?
Added continuous network option for IPPO.
## Why?
Extending functionality for IPPO implementation.
## How?
Added function to create IPPO network with a clipped Gaussian function instead of a categorical policy.
## Extra
Closes issue #631 
-~~The `get_action(...)` function needs an if statement to check the network type in use since the discrete actions are of type `int64` and continuous actions are of type `float32`~~
-  ~~Need to make custom networks to clip distribution. Functionality is present in `acme.tf` but not `acme.jax`~~
- NaN's generated in Multiwalker


mmorris44:
@siddarthsingh1 please create and link an issue for this within the correct milestone (IPPO)

DriesSmit:

PR #654, opened 2022-07-24
feat: full ippo system integration test

### LINKED ISSUE #542: [TEST] Jax MAPPO System Integration
### What do you want to test?
Jax MAPPO full integration test

### Outline of test structure
* Unit tests (if possible)
* Test component interactions
* Ensure not crashing during standard executor and trainer steps

### Definition of done
Passing checks, cover all basic component interactions, edge cases considered

RuanJohn:
## What?
Full distributed Jax ippo system integration test.
## Why?
## Extra
closes #542 
In order to be able to set `lp_launch_type=lp.LaunchType.TEST_MULTI_THREADING` we have to copy the node function [here](https://github.com/instadeepai/Mava/blob/fce79fd6a52a59d8dbf585e2630f78dffd4b6428/mava/systems/jax/launcher.py#L122) as well which is similar to what is done in the single process case.


tests/jax/systems/mappo_system_test.py
```diff
@@ -1,30 +1,161 @@
...
+        multi_process: bool,
+        nodes_on_gpu: List = [],
+        sp_trainer_period: int = 10,
+        sp_evaluator_period: int = 10,
+        name: str = "System",
+        terminal: str = "current_terminal",
+    ) -> None:
+        """Create mock launcer component in order to be able to \
+            overwrite the launch method which enables for testing \
+                a full distribured Mava system."""
```
mmorris44, line 37: typo

tests/jax/systems/mappo_system_test.py
```diff
@@ -1,30 +1,161 @@
...
+    def __init__(
+        self,
+        multi_process: bool,
+        nodes_on_gpu: List = [],
+        sp_trainer_period: int = 10,
+        sp_evaluator_period: int = 10,
+        name: str = "System",
+        terminal: str = "current_terminal",
+    ) -> None:
+        """Create mock launcer component in order to be able to \
```
mmorris44, line 35: typo

tests/jax/systems/mappo_system_test.py
```diff
@@ -1,30 +1,161 @@
...
+            overwrite the launch method which enables for testing \
```
mmorris44, line 36: `for testing` -> `the testing of`

tests/jax/systems/mappo_system_test.py
```diff
@@ -1,30 +1,161 @@
...
+        super().__init__(
+            multi_process,
+            nodes_on_gpu,
+            sp_trainer_period,
+            sp_evaluator_period,
+            name,
+            terminal,
+        )
+
+    def launch(self) -> None:
```
mmorris44, line 47: I'm not the biggest fan of mocking out these methods, since it means we're not actually testing some of the launchpad code. However, I suspect that you may have found the best way to do this, so feel free to disregard this if there's no better way to test this using the actual Mava launcher.

tests/jax/systems/mappo_system_test.py
```diff
@@ -1,30 +1,161 @@
...
+                if episode % self._sp_evaluator_period == 0:
+                    _ = evaluator.run_episode_and_log()
+                    print("Performed evaluator run.")
+
+                print(f"Episode {episode} completed.")
+                episode += 1
+                executor_steps += executor_stats["episode_length"]
+
+
+class MockDistributor(Distributor):
```
mmorris44, line 86: This is just identical to our standard distributor, barring the call to `MockLauncher` instead of `Launcher`. Perhaps just include a comment referring the reader to where to find the actual code, in case it changes?

Also perhaps add a note in the distributor that people should update this section of the test if they change the distributor code.

tests/jax/systems/mappo_system_test.py
```diff
@@ -1,30 +1,161 @@
...
+            episode = 1
+            executor_steps = 0
+
+            _ = self._node_dict["data_server"]
+            _ = self._node_dict["parameter_server"]
+            executor = self._node_dict["executor"]
+            evaluator = self._node_dict["evaluator"]
+            trainer = self._node_dict["trainer"]
+
+            while True:
```
mmorris44, line 71: I'm probably missing something obvious, but I don't see where this loop gets terminated.

mmorris44 (changes requested):
Thanks @RuanJohn! This looks amazing.
Just some minor comments from me and some queries about the approach taken.

The test passes for me after 30 seconds.

tests/jax/systems/mappo_system_test.py
RuanJohn: thanks for catching all of these :sweat_smile: 
sorrted them out.

tests/jax/systems/mappo_system_test.py
RuanJohn: I think that `Mock` is a misleading way of referring to these. I can change the naming to reflect what is really happening. In order to be able to test launchpad systems `launch_type` has to be set tp `"test_mt"` as is done [here[(https://github.com/instadeepai/Mava/blob/45e9178906d368baaa6c67bd1d6cf08cbe39ff77/tests/jax/systems/mappo_system_test.py#L57). In the [tf system integration tests](https://github.com/instadeepai/Mava/blob/45e9178906d368baaa6c67bd1d6cf08cbe39ff77/tests/tf/systems/mappo_system_test.py#L71) we were able to set this directly in the tests due to the way were calling the `lp.launch` method, but in the redesign this method is called inside the `launch` method of the `Launcher` component which gets added to the builder store as `program` by the `Distributor` component. So we are still testing the really mava launcher, we are just setting it to 'test mode'. There is a bit more of a discussion about it [on the launchpad repo](https://github.com/deepmind/launchpad/blob/master/docs/get_started.md).

RuanJohn: Another way, which would remove the need for doing this entirely, would be to add a `test` parameter to the actual `Launcher` component but I opted for not changing a component in order to suit the needs of the tests in this case.

tests/jax/systems/mappo_system_test.py
RuanJohn: Please see the discussion above. I have added a comment above `MockDistributor` to explain was is being done, where the original code can be found and that the code should be updated if the `Distributor` code changes.

tests/jax/systems/mappo_system_test.py
RuanJohn: Good catch, we are testing the system in a distibuted way so `self._multi_process` will be `True` and this while loop will never get executed.

tests/jax/systems/mappo_system_test.py
```diff
@@ -1,30 +1,167 @@
...
+                    print("Performed evaluator run.")
+
+                print(f"Episode {episode} completed.")
+                episode += 1
+                executor_steps += executor_stats["episode_length"]
+
+
+# The code for the distributor is available in
+# mava/components/jax/building/distributor.py the reason
+# for creating a mock distibuto here is so that the launch method of the Launcher
```
mmorris44, line 88: typo: distributor

mmorris44 (changes requested):
Nice changes, thanks! Found a new mistake and something else which slipped past initially.

Also, please don't forget to add comments in the corresponding `launcher` and `distributor` components noting that we should update the tests if they are changed.

tests/jax/systems/mappo_system_test.py
mmorris44: Happy with the existing solution now that comments have been added. As discussed: we don't think it's a good idea to change the launcher to make it more testable.

tests/jax/systems/mappo_system_test.py
mmorris44: Happy now, thanks!

tests/jax/systems/mappo_system_test.py
mmorris44: Cool cool: I'm happy to leave it here though, since it just comes from the copied code.

tests/jax/systems/mappo_system_test.py
```diff
@@ -1,30 +1,167 @@
...
+        multi_process: bool,
+        nodes_on_gpu: List = [],
+        sp_trainer_period: int = 10,
+        sp_evaluator_period: int = 10,
+        name: str = "System",
+        terminal: str = "current_terminal",
+    ) -> None:
+        """Create mock launcher component in order to be able to \
+            overwrite the launch method which enables the testing of \
+                a full distributed Mava system."""
```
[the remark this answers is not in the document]
RuanJohn: > Nice changes, thanks! Found a new mistake and something else which slipped past initially.
> 
> Also, please don't forget to add comments in the corresponding `launcher` and `distributor` components noting that we should update the tests if they are changed.

Please see the most recent chages. Being able to pass in a launch type as a parameter to the system avoids having to create the previous mock classes and avoids adding all the comments throughout the code.

mmorris44 (dismissed):
This looks incredible. I've decided I like this WAY more than the previous solution. In general: I also support setting the launch type in the config instead of manually in the launcher.

Approved, thanks @RuanJohn!

tests/jax/systems/mappo_system_test.py
```diff
@@ -15,16 +20,14 @@
...
 def test_full_system() -> System:
-    """Add description here."""
+    """Full mava system fixture for testing"""
     return mappo.MAPPOSystem()
 
 
-# TODO: fix test
-@pytest.mark.skip(reason="test is currently breaking ci pipeline")
-def test_except_trainer(
+def test_mappo(
```
AsadJeewa, line 27: I get that it is arbitrary which env/ network... we select but is the goal to have a single integration test for each system (and its variations) with the simplest network on the simplest env? If that is the case, then this name is fine (tf was `test_mappo_on_debugging_env`)

tests/jax/systems/mappo_system_test.py
```diff
@@ -1,8 +1,13 @@
 import functools
```
AsadJeewa, line 1: small thing but please add the license to the top of the file

tests/jax/systems/mappo_system_test.py
```diff
@@ -54,36 +63,36 @@ def test_except_trainer(
...
-        multi_process=False,
-        run_evaluator=True,
+        executor_parameter_update_period=1,
+        multi_process=True,
+        run_evaluator=False,
         num_executors=1,
         use_next_extras=False,
-        sample_batch_size=2,
+        sample_batch_size=5,
+        checkpoint=True,
```
AsadJeewa, line 83: What is the thinking behind these variable choices i.e. including checkpointing (Tf does not) and also excluding the evaluator? Links to my next comment

AsadJeewa (changes requested):
Thanks, @RuanJohn. Clean test that's easy to follow. Small change requested. 
Also, please look into this: the test passes and then aborts abruptly with Fatal Python error: Aborted. This does not happen on the TF integration tests.

tests/jax/systems/mappo_system_test.py
```diff
@@ -54,36 +63,36 @@ def test_except_trainer(
...
+    test_full_system.launch()
+    trainer_run = trainer_node.create_handle().dereference()
 
-    # Step the executor
-    executor.run_episode()
-
-    # Step the trainer
-    trainer.step()
+    for _ in range(2):
+        trainer_run.step()
```
AsadJeewa, line 98: I get that we matched the way we were doing so with the TF system tests but is it not necessary to test more than the training node/ train step? Like for example, if other nodes in the system are working as expected?

tests/jax/systems/mappo_system_test.py
AsadJeewa: Also, I get that we matched the way we were doing so with the TF system tests but is it not necessary to test more than the training node/ train step? Like for example, if other nodes in the system are working as expected?

tests/jax/systems/mappo_system_test.py
AsadJeewa: DIscussed in meetings: goal is to use as simple param and env as possible

tests/jax/systems/mappo_system_test.py
```diff
@@ -54,36 +63,36 @@ def test_except_trainer(
...
-    assert isinstance(executor, acme.core.Worker)
+    test_full_system.launch()
+    trainer_run = trainer_node.create_handle().dereference()
 
-    # Step the executor
-    executor.run_episode()
-
-    # Step the trainer
-    trainer.step()
+    for _ in range(2):
```
AsadJeewa, line 97: Also, I get that we matched the way we were doing so with the TF system tests but is it not necessary to test more than the training node/ train step? Like for example, if other nodes in the system are working as expected?

AsadJeewa: Discussed in a meeting. This smoke test check if all components work together by validating with the train step function. Other integration tests will test the other nodes

AsadJeewa (changes requested):
Added replies to my comments. Small changes requested

tests/jax/systems/mappo_system_test.py
AsadJeewa: Discussed in meeting and decided that we want to test as simple param as possible

tests/jax/systems/mappo_system_test.py
AsadJeewa: Added

tests/jax/systems/mappo_system_test.py
```diff
@@ -54,36 +78,37 @@ def test_except_trainer(
...
+        multi_process=True,
         run_evaluator=True,
         num_executors=1,
+        max_queue_size=500,
         use_next_extras=False,
-        sample_batch_size=2,
+        sample_batch_size=5,
+        checkpoint=True,
+        nodes_on_gpu=[],
+        lp_launch_type=lp.LaunchType.TEST_MULTI_THREADING,
```
AsadJeewa, line 101: based on the fixes in the other integration tests, is it not possible to have this as MULTI_PROCESSING @OmaymaMahjoub ?

AsadJeewa (commented):
Few small queries. The main thing is making sure the aborted core dumped is a local issue and concensus on the launch type. Keen to the second reviewer's comments, then happy to approve

tests/jax/systems/mappo_system_test.py
OmaymaMahjoub: I am currently investigating it

OmaymaMahjoub: Based on the chat with the other team members, I found out that it is challenging at the current moment to test the system using `lp_launch_type=lp.LaunchType.TEST_MULTI_PROCESSING` so we will just have the tests in Multi_threading case :sweat: at the current time

AsadJeewa: I see that in the other integration tests, the single process test has MULTI_PROCESSING launch type and multi_process has MULTI_THREADING so yes, I agree that it is fair to merge this: it is a big step to be able to run in multi_process mode now :smile:. Ideally, we should not have to copy the node_fn `(node_fn = copy_node_fn(node_fn))` as it means that multi_threading launch type is shared. Happy to close this issue but please first create an issue to investigate launch_type.

AsadJeewa:
> Thanks, @RuanJohn. Clean test that's easy to follow. Small change requested. Also, please look into this: the test passes and then aborts abruptly with Fatal Python error: Aborted. This does not happen on the TF integration tests. (This happens both when I run the test directly with pytest and also make run-integration-tests)

seems to be a local issue as this is happening on other integration tests locally

tests/jax/systems/mappo_system_test.py
DriesSmit: Agreed. We can maybe remove the checkpointer if we are not checking that the files are getting created.

tests/jax/systems/mappo_system_test.py
DriesSmit: Agreed :slightly_smiling_face: After we create an issue we could merge this.

tests/jax/systems/mappo_system_test.py
OmaymaMahjoub: Okay I will remove it now

tests/jax/systems/mappo_system_test.py
```diff
@@ -54,36 +78,37 @@ def test_except_trainer(
...
 
-    # Step the executor
-    executor.run_episode()
+    test_full_system.launch()
+    trainer_run = trainer_node.create_handle().dereference()
 
-    # Step the trainer
-    trainer.step()
+    for _ in range(5):
+        trainer_run.step()
```
DriesSmit, line 114: How is the trainer stepping if we haven't run the executor yet? Or is the executor running in its own thread when launch is called?

DriesSmit (commented):
Thanks, @RuanJohn and @OmaymaMahjoub :slightly_smiling_face: Just see my few comments.

tests/jax/systems/mappo_system_test.py
OmaymaMahjoub: Okay about that I talked with Ruan about it the step here is not the step function from the trainer it is an internal function in the node to run the process

OmaymaMahjoub: the executor is running while we add system.launch

tests/jax/systems/mappo_system_test.py
OmaymaMahjoub: Does the issue will be closed with this PR or We will create an issue for the future?

OmaymaMahjoub: I have created the issue you can find it here  #690

AsadJeewa (approved):
Thanks @OmaymaMahjoub

AsadJeewa:
I have approved but we need to get in this PR first https://github.com/instadeepai/Mava/pull/692

AsadJeewa (approved):
Thanks, @OmaymaMahjoub

tests/jax/systems/mappo_system_test.py
DriesSmit: Is the trainer not then also running when we call system.launch?

DriesSmit: Happy to merge :)

DriesSmit (approved):
Thanks for all the work :raised_hands:

PR #666, opened 2022-08-03
Fix: Single process launcher

### LINKED ISSUE #667: [BUG] Single process launcher
### Describe the bug
Fix a bug where every node in the single process launcher uses the same store. This caused internal variables to be overwritten.

### To Reproduce
1. Run a single process example with multi_process=False.

### Expected behavior
Each node should have access to its own store that does not influence other nodes when updated.

### Proposal for a fix
Use deepcopy with memo to copy the store for each node.

### Definition of done
Each node should have access to its own store that does not influence other nodes when updated.

### LINKED ISSUE #668: [BUG] Single process code freeze
### Describe the bug
In the single process case, the queue will eventually fill up if the executors generate slightly more experience than the trainer consumes. This is not a problem in the multi-process case as the executors just wait. However, waiting in the single process case causes the trainer to never be able to sample the experience and everything freezes as a result.

### To Reproduce
1. Run a single process example with multi_process=False.

### Expected behavior
A single process system can run without freezing when the queue fills up.

### Proposal for a fix
At each step, before running the executors, check if the Reverb server is almost full. If so then skip the executor step. Otherwise, continue as normal. Furthermore, we can also skip the trainer step if the queue has to few samples in it.

### Definition of done
The single process code does not freeze after the queue has filled up or is empty.

DriesSmit:
## What?
Add a small fix to the distributor. This fixes an issue where the evaluator was overwriting the executor's internal variables in the single process case.
## Why?
Bugfix.
## How?
A shallow copy is now performed on the store values. The copy is shallow to prevent the error where the Reverb server and program cannot be copied in full. Let me know if a shallow copy is not good enough and if something more elaborate is needed like e.g. working with the deepcopy memo argument.
## Extra
Resolves #667, resolves #668


mava/components/jax/building/datasets.py
```diff
@@ -98,7 +98,7 @@ def on_building_trainer_dataset(self, builder: SystemBuilder) -> None:
             postprocess=self.config.postprocess,
         )
 
-        builder.store.dataset_iterator = iter(dataset)
+        builder.store.dataset = iter(dataset)
```
DriesSmit, line 101: Not sure why this changed. Should I change it back to dataset_iterator?

mava/systems/jax/launcher.py
```diff
@@ -26,6 +27,10 @@ class NodeType:
     reverb = lp.ReverbNode
     courier = lp.CourierNode
 
+def copy_store(builder):
+    buildercopy = copy.copy(builder)
+    builder.store = copy.copy(builder.store)
```
OmaymaMahjoub, line 32: Did you meant here `builder.store = copy.copy(builder.store)`-> `builder.store = copy.copy(buildercopy.store)`

mava/components/jax/building/datasets.py
OmaymaMahjoub: I think we need to get it back as it was, else we can make changes on the unit test that mention the dataset as dataset_iterator and it will be fine!

OmaymaMahjoub: I think because the develop branch was not updated because we make this change with the latest commit with dataset unit test

mava/systems/jax/launcher.py
DriesSmit: Yes :) I copied from the wrong file. It should be fixed now :+1:

AsadJeewa (changes requested):
Thanks @DriesSmit this was a nice learning experience! I tested extensively. Just the small dataset_iterator name change and I am happy to approve

mava/components/jax/building/datasets.py
AsadJeewa: Yes, please

DriesSmit: Will do. I can't seem to see other differences with the develop branch? @OmaymaMahjoub maybe we can have a chat about that other file that might be missing?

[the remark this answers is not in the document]
DriesSmit: Will do :slightly_smiling_face:

DriesSmit: I see the test file showed up now. I fixed it :+1: Let me know if everything looks good now. @AsadJeewa did the system train in your experiments?

mava/systems/jax/launcher.py
```diff
@@ -14,18 +14,32 @@
...
 from mava.utils import lp_utils
 
 
 class NodeType:
     reverb = lp.ReverbNode
     courier = lp.CourierNode
 
+def copy_store(builder: SystemBuilder, multi_process: bool) -> SystemBuilder:
+    if multi_process:
+        # Note: It is unnecessary to copy when Launchpad is used.
```
RuanJohn, line 33: Just a minor question, but do we not always use Launchpad?

OmaymaMahjoub: Nope launchpad only in the case of multi-process. Check `mava/systems/jax/launcher.py `  
```
if multi_process:
      self._program = lp.Program(name=name)
```

AsadJeewa: we only use lp with multi_process
https://github.com/instadeepai/Mava/blob/develop/mava/systems/jax/launcher.py#L59

RuanJohn: Thanks for the replies. It makes sense, but just to clarify we are still using the launchpad nodes in single process?

AsadJeewa: As far as I see it, no (https://github.com/instadeepai/Mava/blob/develop/mava/systems/jax/launcher.py#L112) but it is misleading that we check the launchpad node type in the next line

mava/components/jax/building/datasets.py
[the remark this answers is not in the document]
AsadJeewa: Thanks

mava/systems/jax/launcher.py
DriesSmit: We are only using the Launchpad node type to determine if we should create a Reverb server or not in the single process case.

mava/systems/jax/launcher.py
```diff
@@ -27,6 +29,30 @@ class NodeType:
     courier = lp.CourierNode
 
 
+def copy_store(builder: SystemBuilder, multi_process: bool) -> SystemBuilder:
```
mmorris44, line 32: Please rename this to something like `copy_store_in_builder`, to make its function more clear.

mava/components/jax/building/distributor.py
```diff
@@ -54,26 +57,27 @@ def on_building_program_nodes(self, builder: SystemBuilder) -> None:
             nodes_on_gpu=self.config.nodes_on_gpu,
             name=self.config.distributor_name,
             terminal=self.config.terminal,
+            lp_launch_type=self.config.lp_launch_type,
         )
 
         # tables node
         data_server = builder.store.program.add(
-            builder.data_server,
+            copy_store(builder, self.config.multi_process).data_server,
```
mmorris44, line 65: I think we should rather do this (see likewise elsewhere) in the `add` method of `launcher.py`. There's already a check there for single/multi-processing.

This can be done by either passing through the builder as an argument as well, or perhaps by accessing `self` from the `node_fn`. Not sure if the latter is possible, but worth a try; if not, the former will still be quite neat.

The copy should happen just before the following is run:
```process = node_fn(*arguments)```

File: mava/components/jax/building/distributor.py [hunk 6]
```diff
@@ -17,9 +17,11 @@
 from dataclasses import dataclass
 from typing import Callable, List, Optional, Union
 
+import launchpad as lp
+
 from mava.components.jax import Component
 from mava.core_jax import SystemBuilder
-from mava.systems.jax.launcher import Launcher, NodeType
+from mava.systems.jax.launcher import Launcher, NodeType, copy_store
```
mmorris44, line 24: Should we rather have `copy_store` in a utils file somewhere? Might be useful elsewhere. Just feels a little weird to have it belong in the launcher.

I would even prefer it being with the builder.

mmorris44 (changes requested):
Thanks @DriesSmit, really nice solution to the deep copying!
I have a couple of thoughts on ways to perhaps make this code a little nicer and more intuitive. Keen to hear if you / others disagree and prefer the way it currently is.

mava/components/jax/building/distributor.py
DriesSmit: Agreed :slightly_smiling_face:  Option one is definitely possible and if it looks cleaner we should do it. I intially tried option two, but it seemed quite hard to drive the self, as you mentioned, from the function itself. Maybe it is possible?

mava/systems/jax/launcher.py
DriesSmit: Thanks @mmorris44 :slightly_smiling_face: It was a team effort :fire:  I changed the naming to `copy_builder`. As we are actually copying the entire builder except for store.program. Let me know if you agree.

mava/systems/jax/launcher.py
mmorris44: @AsadJeewa is correct. We do not use any `launchpad` nodes in the single process system

mava/components/jax/building/distributor.py
DriesSmit: Agreed :+1: I updated the code as suggested.

mava/components/jax/building/distributor.py
DriesSmit: Hmmm, not actually sure how to do option one. Once the function is created it is already too late. Any ideas @mmorris44 ?

mmorris44: Option 2 can be achieved with the following:
```
class Class:
    def __init__(self):
        self.param = 1
        
    def method(self):
        return self.param
    
obj = Class()
fn = obj.method
print(fn())
fn.__self__.param = 2
print(fn())
```

mava/systems/jax/launcher.py
mmorris44: Ahh, my mistake. Yup, agreed then! I like the new name and file location.

mava/components/jax/building/distributor.py
DriesSmit: But we can possibly use your method if we call the same function on the buildercopy and return that :+1:

DriesSmit: @OmaymaMahjoub figure it out :partying_face:

OmaymaMahjoub: Just need to fix the launcher unit test! And I think it will be done!

DriesSmit: Implemented small updates to get to option 1. Let me know if you are happy with it @OmaymaMahjoub :slightly_smiling_face:

OmaymaMahjoub: I will just run pre commit and check the tests to verify why the checks failed

mmorris44 (approved):
Happy with it now, thanks everyone!
Just one minor comment which can be ignored if you want.

examples/jax/debugging/simple_spread/feedforward/decentralised/run_mappo_single_process.py
```diff
@@ -0,0 +1,97 @@
...
+        num_executors=1,
+        use_next_extras=False,
+        max_queue_size=5000,
+        sample_batch_size=2,
+        max_in_flight_samples_per_worker=4,
+        num_workers_per_iterator=-1,
+        rate_limiter_timeout_ms=-1,
+        checkpoint=True,
+        nodes_on_gpu=[],
+        lp_launch_type=lp.LaunchType.LOCAL_MULTI_PROCESSING,
```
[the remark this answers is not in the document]
DriesSmit: Agreed. I removed it when trying to get the system to train.

mava/components/jax/building/datasets.py
AsadJeewa: It runs on docker but when I run on venv I get 
`jaxlib.xla_extension.XlaRuntimeError: INTERNAL: Attempting to perform BLAS operation using StreamExecutor without BLAS support`. This is a GPU memory issue so not sure if it a localised issue

mmorris44 (dismissed):

mava/systems/jax/launcher.py
```diff
@@ -159,24 +170,38 @@ def launch(self) -> None:
...
+            # getting the maximum queue size
+            queue_threshold = data_server.server_info()["trainer"].max_size
 
-                if episode % self._sp_trainer_period == 0:
+            while True:
+                # if the queue is too full we skip the executor to ensure that the
+                # executor won't hang when trying to push experience
+                if data_server.server_info()["trainer"].current_size < int(
+                    queue_threshold * 0.75
+                ):
```
KaleabTessera, line 195: Do we need this check? PPO uses a queue, so each element get's removed as it is used. The max `current_size` I could get when running the example was ~ 7, when queue's `.max_size` is 500. 

Basically, was there an issue this is solving or just a precaution? I think the reverb params like `max_times_sampled` should be handling these sort of things.

What do you think?

KaleabTessera (changes requested):
Thanks so much, @DriesSmit, this is an excellent issue to fix :rocket: ! Just some minor comments.

mava/systems/jax/launcher.py
```diff
@@ -122,6 +127,12 @@ def add(
                     + "Single process currently only supports one node per type."
                 )
 
+            copy_store_builder = copy_builder(builder=node_fn.__self__)
+
+            # Execute the function from the copied builder.
+            function_name = str(repr(node_fn).split(" ")[2].split(".")[-1])
+            node_fn = getattr(copy_store_builder, function_name)
```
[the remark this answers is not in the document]
mmorris44: If this works, I much prefer this as well. The string stuff is a bit dodgy.

mava/systems/jax/launcher.py
mmorris44: This is the `launcher` for all systems, so will this not be useful when we add other systems?

DriesSmit: This prevents Reverb from pausing the executor when the queue fills up. In this case the queue does not fill up yes. But it does fill up under other settings. The same thing with the queue is empty when the trainer samples. We can't let Reverb handle it internally as Reverb just waits which hangs the entire training process.

mava/systems/jax/launcher.py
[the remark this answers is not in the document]
DriesSmit: Agreed. If we can remove this string stuff that would be better. I am not completely understanding the `copy_fn` function. Where do we copy the store?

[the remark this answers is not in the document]
DriesSmit: It works! Thanks @KaleabTessera :fire:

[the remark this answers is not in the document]
DriesSmit: Unfortunately, it does not seem to be learning :/Maybe you can try the quick change and see if you can get it learning?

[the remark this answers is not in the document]
DriesSmit: Screenshot from 2022-08-10 13-47-26

The difference in performance between the old (red) and new (orange) methods.

mava/systems/jax/launcher.py
KaleabTessera: Does a `trainer.step` always reduce the size of the table? Even when are not using a queue?

DriesSmit: `trainer.step `hangs if there is less experience in the `data_server` than is requested by the trainer sample. This is not just for the queue, but also for other table types. So if the trainer samples more data than gets generated at some point it will hang. Does that make sense?

mava/systems/jax/launcher.py
```diff
@@ -35,12 +36,15 @@ def __init__(
         self,
         multi_process: bool,
         nodes_on_gpu: List = [],
-        sp_trainer_period: int = 10,
+        sp_trainer_period: int = 1,
```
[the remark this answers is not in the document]
DriesSmit: Updated :+1:

KaleabTessera (approved):
Thanks @DriesSmit !

OmaymaMahjoub (dismissed):
Thanks @DriesSmit!

KaleabTessera (dismissed):

KaleabTessera (approved):

OmaymaMahjoub (approved):

mava/systems/jax/launcher.py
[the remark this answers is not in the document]
KaleabTessera: This appears to work now.

PR #674, opened 2022-08-10
Test: integration test of the Data Server

### LINKED ISSUE #646: [TEST] Data Server Integration
### What do you want to test?
Full integration test for data server.

### Outline of test structure
Should use all components typically found in the data server and run without error, but not using launchpad. See the jax IPPO example for how the components are typically logically distributed between core nodes. If multiple tests are required to cover different distinct cases (e.g. value-based vs policy-gradient), then they should be included.

### Definition of done
Test(s) passing and including all typical components.

OmaymaMahjoub:
## What?
Dataserver integration test
## Why?
## Extra
Close #646 
The data_server tested only in the case of single process


tests/jax/integration/data_server_test.py
```diff
@@ -12,91 +12,134 @@
...
-            )
+    table_trainer = data_server.server_info()["trainer"]
+    assert table_trainer.num_episodes == 5  # 5 episodes
+    assert table_trainer.num_unique_samples != 0
 
-        # On policy params
-        assert_if_value_is_not_none(
-            system_config.get("max_queue_size"), table_info.max_size
-        )
+    # dataset added by the exevutor via the adders
```
siddarthsingh1, line 99: `exevutor` should be `executor`.

siddarthsingh1 (dismissed):
Looks good. Thanks @OmaymaMahjoub. There's just a small typo in one of the comments. I also wanted to ask, we only test the use of the `TrajectoryDataset` since we only have IPPO, would we need to add the `TransitionDataset` after DQN is available?

OmaymaMahjoub:
> Looks good. Thanks @OmaymaMahjoub. There's just a small typo in one of the comments. I also wanted to ask, we only test the use of the `TrajectoryDataset` since we only have IPPO, would we need to add the `TransitionDataset` after DQN is available?

Thanks @siddarthsingh1  for the review! We made the two types testing in the unit test, and I believe that the integration test needs to be on a system that is predefined  which is in our cas IPPO. However, if you feel the importance of making `TransitionDataset` integration test I will try to see how to manage and add it to a customized system

tests/jax/integration/data_server_test.py
OmaymaMahjoub: Done!

siddarthsingh1 (dismissed):
Thanks for the explanation @OmaymaMahjoub. The current test seems fine 👍

tests/jax/integration/data_server_test.py
```diff
@@ -12,91 +12,134 @@
...
+from tensorflow.python.data.ops import dataset_ops
+
+from mava.systems.jax import System
+from tests.jax.integration.mock_systems import mock_system_single_process
+
+
+@pytest.fixture
+def test_system() -> System:
+    """A built IPPO system"""
+    return mock_system_single_process()
```
DriesSmit, line 29: Thanks, @OmaymaMahjoub :slightly_smiling_face:  Looks good. Are we only testing multi_process for the integration tests?

OmaymaMahjoub: Thanks @DriesSmit  for reviewing! Unfortunately since data server a Reverb node it was tough to find out how to test it in the multi process so at least we tested it in the single process case

DriesSmit: That makes sense.

DriesSmit (approved):
Looks good, thanks @OmaymaMahjoub :fire:

tests/jax/integration/data_server_test_data.py
```diff
@@ -1,100 +0,0 @@
-# python3
```
AsadJeewa, line 1: happy that we have removed this

AsadJeewa (approved):
Extensive testing. Thanks for all your efforts @OmaymaMahjoub

tests/jax/integration/data_server_test_data.py
OmaymaMahjoub: It was before attached to an old test was made (draft)

PR #675, opened 2022-08-10
Test: trainer integration test and system integration test editing

### LINKED ISSUE #644: [TEST] Trainer Integration
### What do you want to test?
Full integration test for trainer.

### Outline of test structure
Should use all components typically found in the trainer and run without error, but not using launchpad. See the jax IPPO example for how the components are typically logically distributed between core nodes. If multiple tests are required to cover different distinct cases (e.g. value-based vs policy-gradient), then they should be included.

### Definition of done
Test(s) passing and including all typical components.

OmaymaMahjoub:
## What?
Trainer integration test
Edit the system integration test to work in the multi-process case #542 

## Why?

## Extra
Close #644


tests/jax/integration/trainer_test.py
```diff
@@ -0,0 +1,201 @@
+import functools
```
AsadJeewa, line 1: please add the license at the top of this file

OmaymaMahjoub: Done!

tests/jax/components/building/distributor_test.py
```diff
@@ -119,27 +119,34 @@ def test_on_building_program_nodes_multi_process(
         "evaluator",
         "trainer",
     ]
+
```
AsadJeewa, line 122: see above (line 50 and elsewhere). Think the docstring should change from "to test no multi_process" to "to test single-process"

tests/jax/components/building/distributor_test.py
```diff
@@ -162,23 +169,28 @@ def test_on_building_program_nodes_multi_process_no_evaluator(
...
-    assert (
-        mock_builder.store.program._program._groups["parameter_server"][-1]._constructor
-        == mock_builder.parameter_server
+        ]._constructor()
+        == "Parameter Server Test"
     )
     assert (
-        mock_builder.store.program._program._groups["executor"][-1]._constructor
-        == mock_builder.executor
+        mock_builder.store.program._program._groups["executor"][-1]._constructor(
```
AsadJeewa, line 184: Think a comment about why we need to index -1 would be useful

tests/jax/integration/trainer_test.py
```diff
@@ -0,0 +1,218 @@
...
+
+    (
+        data_server,
+        parameter_server,
+        executor,
+        evaluator,
+        trainer,
+    ) = test_system._builder.store.system_build
+
+    for i in range(0, 4):
```
AsadJeewa, line 113: 0 is reduendant so can be removed. Should just run for `for i in range(5)` like the mappo test

tests/jax/integration/trainer_test.py
```diff
@@ -0,0 +1,218 @@
...
+    )
+
+    # Disable the run of the trainer node
+    (trainer_node,) = test_system._builder.store.program._program._groups["trainer"]
+    trainer_node.disable_run()
+    test_address_builder.bind_addresses([trainer_node])
+
+    # launch the system
+    test_system.launch()
+    time.sleep(10)  # wait the executor to run at least one time
```
AsadJeewa, line 199: "wait till the executor has run"

tests/jax/integration/trainer_test.py
```diff
@@ -0,0 +1,218 @@
...
+        sample_batch_size=5,
+        checkpoint=True,
+        nodes_on_gpu=[],
+        lp_launch_type=lp.LaunchType.TEST_MULTI_THREADING,
+    )
+
+    # Disable the run of the trainer node
+    (trainer_node,) = test_system._builder.store.program._program._groups["trainer"]
+    trainer_node.disable_run()
+    test_address_builder.bind_addresses([trainer_node])
```
AsadJeewa, line 195: the test still seems to pass when I remove this line. Same for removing the sleep (though I get the need for sleep)

tests/jax/integration/trainer_test.py
```diff
@@ -0,0 +1,218 @@
...
+            assert jnp.all(categorical_value_head["w"] == 0)
+
+    # Step function
+    trainer.step()
+
+    # Check that the trainer update the network
+    for net_key in trainer.store.networks["networks"].keys():
+        mu = trainer.store.opt_states[net_key][1][0][-1]
+        for categorical_value_head in mu.values():
+            assert not jnp.all(categorical_value_head["b"] == 0)
```
AsadJeewa, line 217: really like this check :smile:

AsadJeewa (changes requested):
Thanks @OmaymaMahjoub 
Really nice work on this. Few small changes requested

tests/jax/integration/trainer_test.py
OmaymaMahjoub: I think we need to keep the time.sleep because it depends on the speed of the device that you have! For the bind it is needed else from my side it crushes  and it was mentioned in the tests in Launchpad

OmaymaMahjoub: You were right we don't need the `test_address_builder.bind_addresses([trainer_node])` for the test to pass

tests/jax/components/building/distributor_test.py
AsadJeewa: discussed offline

tests/jax/integration/trainer_test.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # After run step method
+    for net_key in trainer.store.networks["networks"].keys():
+        mu = trainer.store.opt_states[net_key][1][0][-1]
+        for categorical_value_head in mu.values():
+            assert not jnp.all(categorical_value_head["b"] == 0)
+            assert not jnp.all(categorical_value_head["w"] == 0)
+
+
+def test_trainer_multi_thread(test_system: System) -> None:
```
AsadJeewa, line 129: since our variable is multi_process, I think that this should be called `test_trainer_multi_process()` still

AsadJeewa (changes requested):
Small request (which is not essential), then happy to approve. Thanks @OmaymaMahjoub

tests/jax/integration/trainer_test.py
OmaymaMahjoub: But in this case aren't we having lp_launch_type=lp.LaunchType.TEST_MULTI_THREADING and probably in the future we will add the test for lp_launch_type=lp.LaunchType.TEST_MULTI_PROCESSING so I thought to keep it as that properly!
What do you think?

DriesSmit (dismissed):
This looks great, thanks @OmaymaMahjoub and @RuanJohn :fire: Really glad we can test multi_process now.

tests/jax/integration/trainer_test.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # Check that the trainer update the network
+    for net_key in trainer.store.networks["networks"].keys():
+        mu = trainer.store.opt_states[net_key][1][0][-1]
+        for categorical_value_head in mu.values():
+            assert not jnp.all(categorical_value_head["b"] == 0)
+            assert not jnp.all(categorical_value_head["w"] == 0)
+
+
+def test_trainer_multi_process(test_system_mp: System) -> None:
```
AsadJeewa, line 108: The only issue is that the test (passes) does not terminate

AsadJeewa (commented):
Lets discuss the termination of the test offline

tests/jax/integration/trainer_test.py
OmaymaMahjoub: It won't make any problem since it is a test that's what I understood, also the checks passes them so it won't bother
what do you think?

AsadJeewa: Ideally, we should use the terminators to fix this

OmaymaMahjoub: It won't work unfortunately we need to come up with another idea

AsadJeewa (approved):
Thanks @OmaymaMahjoub

AsadJeewa (approved):

DriesSmit (approved):
Thanks @OmaymaMahjoub :slightly_smiling_face:

tests/jax/integration/trainer_test.py
OmaymaMahjoub: We solved that for the time being using `os.kill()`, however I will create an issue to add a feature that give us the possibility of using `lp.stop()` within the Terminator components

PR #699, opened 2022-08-24
Feature: Remove multi-threading

### LINKED ISSUE #690: [INVESTIGATION] The multi-thread vs multi-process
### What do you want to investigate?
In MAVA currently we have three options **SIngle process**, **Multi process** and **Multi thread**.

To run the multi-thread case, we need to have `multi_process` attribute equal True, and to  fill the `lp_launch_type` attribute with `lp.LaunchType.LOCAL_MULTI_THREADING`. However, if in the future we are going to consider the case of multi thread we need to identify an attribute for the `multi_thread` independent of  themulti-procss case.

### Definition of done
- Are we only going to keep the single-process and multi-process cases, or will we also keep the code for the multi-thread case?
- If we want to keep the multi-thread case, we should change the launcher code.

DriesSmit:
## What?
Remove the multi-threading option from the distributor and launcher.
## Why?
The multi-threading option is unnecesary and complicates our testing process.
## How?
Remove the multi-threading option from the distributor and launcher code.

## Extra
Close #690

### DIFF
--- a/examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_single_process.py
+++ b/examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_single_process.py
@@ -18,8 +18,6 @@ import functools
 import os
 from datetime import datetime
 from typing import Any
-
-import launchpad as lp
 import optax
 from absl import app, flags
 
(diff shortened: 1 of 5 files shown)


AsadJeewa (dismissed):
An important step! Thanks @DriesSmit

OmaymaMahjoub (approved):
Looks good from my side thanks @DriesSmit! Just try to run the trainer test locally before merging the PR to check

AsadJeewa (approved):

PR #700, opened 2022-08-25
fix: fix the ippo test and add a test flag

OmaymaMahjoub:
## What?
Add a test flag helping to launch the system in mode TEST_MULTI_THREADING
FIx the ippo test
## Why?


mava/systems/jax/launcher.py
```diff
@@ -43,10 +43,11 @@ def __init__(
         single_process_max_episodes: Optional[int] = None,
         name: str = "System",
         terminal: str = "current_terminal",
+        is_test: Optional[bool] = False,
     ) -> None:
         """Initialise the launcher.
 
-        If multi-process, set up the launchpad program.
+        If multi-process,et up the launchpad program.
```
RuanJohn: smakk typo to fix please :)

OmaymaMahjoub: yeah I just fixed thanks!

RuanJohn (changes requested):
It looks good @OmaymaMahjoub! One other minor request I have is that we also now only have to copy the node function in the multi process case when we are testing. Could you please add an if statement there as well?

OmaymaMahjoub:
> It looks good @OmaymaMahjoub! One other minor request I have is that we also now only have to copy the node function in the multi process case when we are testing. Could you please add an if statement there as well?

Thanks @RuanJohn  for the review! I updated the PR

RuanJohn (dismissed):
Thank you @OmaymaMahjoub :fire: Ran the test locally with the new changes and it works!

OmaymaMahjoub:
> Thank you @OmaymaMahjoub fire Ran the test locally with the new changes and it works!

Thanks @RuanJohn  let's wait for @DriesSmit  and @AsadJeewa because they came up with an idea to improve the test!

mava/systems/jax/launcher.py
```diff
@@ -166,12 +169,21 @@ def launch(self) -> None:
                 nodes_on_gpu=self._nodes_on_gpu,
             )
 
-            lp.launch(
-                self._program,
-                launch_type=lp.LaunchType.LOCAL_MULTI_PROCESSING,
-                terminal=self._terminal,
-                local_resources=local_resources,
-            )
+            if self._is_test:
```
DriesSmit, line 172: Can we maybe change this to:
```
launch_type = lp.LaunchType.TEST_MULTI_THREADING if self._is_test else lp.LaunchType.LOCAL_MULTI_PROCESSING
```

DriesSmit: And just asign the launch_type.

DriesSmit (changes requested):
Thanks @OmaymaMahjoub :fire: Just see my one small comment.

mava/systems/jax/launcher.py
OmaymaMahjoub: Okay!

OmaymaMahjoub: Done

DriesSmit (approved):

AsadJeewa (approved):
Nice!

RuanJohn (approved):
Thank you @OmaymaMahjoub :muscle:

PR #709, opened 2022-09-01
feat: added evaluation interval to environment loop

### LINKED ISSUE #708: [FEATURE] Implement evaluation intervals
### Feature
Add an `evalution_interval` argument to Mava which allows for triggering evaluation after a fixed amount of executor steps or episodes. Additionally include a way to control for how many episodes should run when triggered. Logging will also be updated to log aggregated data from all evaluator runs. 

### Testing
Ensure that the executor and trainer are still running as expected and verify that existing tests pass. 

### Definition of done
Evaluation occurs as expected and does not hinder the trainer and executor processes. 

RuanJohn:
## What?
Update mava evaluation to run at fixed intervals and for a given number of episodes at a time. 
## Why?
This will allow for better evaluation in general and for comparing to results of other repos and papers more fairly. 
## How?
Include`evaluation_interval` and `evaluation_duration` parameters to the executor which can trigger the evaluation loop to run for a set number of episodes after a set amount of executor steps or episodes have been completed. When evaluation is run at fixed intervals for a fixed number of episodes, the mean of all metrics computed over the evaluation episodes are logged. The only difference currently is for Smac, where the `WinRate` metric is now logged as `num_battles_won/num_evaluation_episodes`. The executor and trainer are left untouched and will still log continually as was previously done. The evaluator will also default to continual evaluation if no interval is passed in to the system. 
## Extra
These updates have only been included in the `ParallelEnvironmentLoop` class for now, but the `SequentialEnvironmentLoop` class can be updated in similar way once needed and when TF has been deprecated. The current `ParallelEnvironmentLoop` class is also overly complicated due to having to support TF. Once TF has been deprecated, we can clean the environment loop.  
Please see here an example (with `evaluation_interval = {"executor_steps": 10000}` and `evaluation_duration=32`) of how the metric logging has been updated and notice how the new win rate is no longer the prervious cumulative win rate and how the evaluator logs significantly less than the executor:
Screenshot from 2022-09-09 15-18-09
We will include additional evaluation feature, like the computation of the absolute metric and logging evaluation data json files, in later PRs. 
Close #708


mava/components/jax/building/parameter_client.py
```diff
@@ -120,6 +120,13 @@ def on_building_executor_parameter_client(self, builder: SystemBuilder) -> None:
         else:
             set_keys = [x for x in set_keys if x.startswith("executor")]
 
+        # Set this so the evaluator is able to update continually
+        # with the latest parameters
+        if builder.store.is_evaluator is True:
+            update_period = 1
```
DriesSmit, line 126: Isn't 1 a bit too much? This means that we are pulling the networks at every environment step.

mava/components/jax/executing/base.py
```diff
@@ -46,7 +47,18 @@ def on_execution_init_start(self, executor: SystemExecutor) -> None:
         Returns:
             None.
         """
-        executor._interval = self.config.interval  # type: ignore
+        # TODO (Ruan): Double check that this is necessary
+        if self.config.evaluation_duration:
+            try:
```
DriesSmit, line 52: Change this to an if and raise ValueError exception, please :)

tests/jax/components/building/parameter_client_test.py
```diff
@@ -320,7 +320,8 @@ def test_executor_parameter_client_evaluator_with_parameter_client(
     assert mock_builder.store.executor_parameter_client._get_call_counter == 0
     assert mock_builder.store.executor_parameter_client._set_call_counter == 0
     assert mock_builder.store.executor_parameter_client._set_get_call_counter == 0
-    assert mock_builder.store.executor_parameter_client._update_period == 200
+    assert mock_builder.store.executor_parameter_client._update_period == 1
```
DriesSmit, line 323: Just see my above comment.

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
```diff
@@ -0,0 +1,105 @@
...
+    str(datetime.now()),
+    "Experiment identifier that can be used to continue experiments.",
+)
+flags.DEFINE_string("base_dir", "~/mava", "Base dir to store experiments.")
+
+
+def main(_: Any) -> None:
+    """Example running feedforward MADQN on SMAC environment."""
+
+    # WARNING (dries): This code has not been run yet. There might still
```
DriesSmit, line 47: Can you remove this warning throughout Mava, please :)

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
```diff
@@ -0,0 +1,117 @@
...
+
+import optax
+from absl import app, flags
+
+from mava.systems.jax import ippo
+from mava.utils.environments import debugging_utils
+from mava.utils.loggers import logger_utils
+
+# Without this flag, JAX uses the whole GPU from the beginning and our trainer crashes.
+os.environ["XLA_PYTHON_CLIENT_PREALLOCATE"] = "false"
```
DriesSmit, line 30: This flag is not needed anymore, right?

mava/components/jax/executing/base.py
```diff
@@ -46,7 +47,18 @@ def on_execution_init_start(self, executor: SystemExecutor) -> None:
         Returns:
             None.
         """
-        executor._interval = self.config.interval  # type: ignore
+        # TODO (Ruan): Double check that this is necessary
```
DriesSmit, line 50: Is this TODO addressed?

mava/environment_loop.py
```diff
@@ -575,24 +580,68 @@ def should_run_loop(eval_condtion: Tuple) -> bool:
...
+                        # Add consecutive evaluation run data
+                        result = self.run_episode()
+                        episode_count += 1
+                        step_count += result["episode_length"]
+                        results = jax.tree_map(lambda x, y: x + y, results, result)
+
+                    # compute the mean over all evaluation runs
+                    results = jax.tree_map(lambda x: x / evaluation_duration, results)
+
+                    # Compute true win rate for SMAC
```
DriesSmit, line 629: I feel we should not have custom SMAC logic in our environment loop, but somehow include the logic in the SMAC wrapper.  We then call a generic environment function here. Not sure if this is possible?

DriesSmit (commented):
Thanks @RuanJohn and @OmaymaMahjoub! :rocket: This PR looks awesome. Great work on this one! :slightly_smiling_face:  Just see my few comments.

mava/components/jax/building/parameter_client.py
RuanJohn: It is set to 1 so that the evaluator can continually get the latest executor steps value from the parameter server [here](https://github.com/instadeepai/Mava/blob/9b768f0f1b1955ab24cc30da613918479e423db0/mava/environment_loop.py#L651). Unless there is a better way to continually update only the steps instead of all the parameters?

mava/components/jax/executing/base.py
RuanJohn: Done :)

mava/components/jax/executing/base.py
RuanJohn: Adressed and removed

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
RuanJohn: Done :)

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
RuanJohn: You are correct, removed it

tests/jax/components/building/parameter_client_test.py
RuanJohn: Same applies as above

mava/components/jax/executing/base.py
DriesSmit: Thanks :)

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
DriesSmit: Thanks :)

mava/components/jax/executing/base.py
DriesSmit: Thanks :)

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
DriesSmit: Thanks :)

mava/components/jax/building/parameter_client.py
DriesSmit: We can maybe create a `force_update` function for the executor? When we call `force_update` it forces the `parameter_client` to update the executor. I am just worried that setting it to 1 would affect performance when the evaluators are running.

OmaymaMahjoub: Done, I added a flag named force_update that if is it true the update will happen immediately else it will be based on the update config

mava/environment_loop.py
OmaymaMahjoub: Done we have added a function called extra_ stats that will return the win rate from smac wrapper in the case of SMAC

examples/jax/smac/feedforward/decentralised/run_ippo_sep_networks.py
```diff
@@ -44,9 +44,6 @@
 def main(_: Any) -> None:
```
AsadJeewa, line 44: Please delete this file as it is no longer valid since we removed the sep networks. Also, please remove the incorrect docstring `single_network` (3 times) in this file: https://github.com/instadeepai/Mava/blob/develop/mava/systems/jax/ippo/networks.py#L201

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
```diff
@@ -0,0 +1,114 @@
+# python3
```
AsadJeewa, line 1: Please update the examples/README.md to add links to these these files

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
```diff
@@ -0,0 +1,114 @@
...
+    logger_factory = functools.partial(
+        logger_utils.make_logger,
+        directory=FLAGS.base_dir,
+        to_terminal=True,
+        to_tensorboard=True,
+        time_stamp=FLAGS.mava_id,
+        time_delta=log_every,
+    )
+
+    # Optimizer.
```
AsadJeewa, line 84: Please make this and other places optmiSer

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
```diff
@@ -0,0 +1,102 @@
...
+    logger_factory = functools.partial(
+        logger_utils.make_logger,
+        directory=FLAGS.base_dir,
+        to_terminal=True,
+        to_tensorboard=True,
+        time_stamp=FLAGS.mava_id,
+        time_delta=log_every,
+    )
+
+    # Optimizer.
```
AsadJeewa, line 73: Please make this and other places optmiSer

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
```diff
@@ -0,0 +1,102 @@
...
+    optimizer = optax.chain(
```
AsadJeewa, line 74: Example does not run: This needs to be separate policy_optmiser and critic_opmiser since yesterdays PR from Dries

mava/components/jax/executing/base.py
```diff
@@ -46,7 +47,18 @@ def on_execution_init_start(self, executor: SystemExecutor) -> None:
         Returns:
             None.
         """
-        executor._interval = self.config.interval  # type: ignore
+
+        if self.config.evaluation_duration != 1 and (
```
AsadJeewa, line 51: could we also raise an error if evaluator duration is  > evaluator interval (for episodes key)

tests/jax/components/executing/base_test.py
```diff
@@ -73,7 +73,23 @@ def test_on_execution_init_start(
...
+
+    Args:
+        mock_executor: Executor
+        dummy_config: ExecutorInitConfig
+    """
+    mock_executor.store.is_evaluator = True
+    executor_init = ExecutorInit(config=dummy_config)
+    executor_init.on_execution_init_start(executor=mock_executor)
+
+    assert mock_executor.store._evaluation_interval == dummy_config.evaluation_interval  # type: ignore # noqa: E501
```
AsadJeewa, line 92: please add an assertion to test duration being set also

mava/environment_loop.py
```diff
@@ -575,28 +580,61 @@ def should_run_loop(eval_condtion: Tuple) -> bool:
...
         episode_count, step_count = 0, 0
 
-        # Currently, we only use intervals for eval loops.
-        environment_loop_schedule = (
-            self._executor._evaluator and self._executor._interval
-        )
-        if environment_loop_schedule:
-            eval_condition = check_count_condition(self._executor._interval)
+        # TODO (Ruan): Checking whether we are using Jax or TF here
+        # this should be removed once we deprecate TF.
```
AsadJeewa, line 589: Removing TF will make the code much more readable

mava/components/jax/executing/base.py
```diff
@@ -46,7 +47,18 @@ def on_execution_init_start(self, executor: SystemExecutor) -> None:
...
+            self.config.evaluation_interval is None
+        ):
+            raise ValueError(
+                "Missing evaluation interval value: Evaluation duration"
+                + " value was provided without a value for the evaluation interval."
+            )
+
+        if executor.store.is_evaluator:
+            executor.store._evaluation_interval = self.config.evaluation_interval  # type: ignore # noqa: E501
```
AsadJeewa, line 60: why is this variable with an underscore?

AsadJeewa (changes requested):
Thanks, @OmaymaMahjoub and @RuanJohn 
A few changes requested

mava/components/jax/building/parameter_client.py
OmaymaMahjoub: After having discussion with @KaleabTessera  I find out that the hook should not include parameters so in order to keep the same design, we will add a new function called `force_update` in the executor.

mava/environment_loop.py
```diff
@@ -575,28 +580,61 @@ def should_run_loop(eval_condtion: Tuple) -> bool:
...
+        if hasattr(self._executor, "store"):
+            environment_loop_schedule = self._executor._evaluator and (
+                self._executor.store._evaluation_interval is not None
+            )
+
+            if environment_loop_schedule:
+                eval_condition = check_count_condition(
```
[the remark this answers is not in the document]
RuanJohn: Yes indeed, any of these should work, although we have only tested `executor_steps` for now.

examples/jax/smac/feedforward/decentralised/run_ippo_sep_networks.py
RuanJohn: Done. Thank you for catching that :call_me_hand:

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
RuanJohn: Added some more examples to the readme that are already in `develop`. I can add these examples in another PR once we merge this one so that the develop urls for the examples exists. What are your thoughts on that?

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
RuanJohn: Done

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
RuanJohn: Done

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
RuanJohn: Fixed.

mava/environment_loop.py
RuanJohn: Agreed! :muscle:

mava/components/jax/executing/base.py
RuanJohn: I am not sure whether we should? I think we run the risk of having to cater for every possible configuration that could lead to weird evaluation. The reason we raise this in particular is because passing in an `evaluation_duration` parameter without giving an `evaluation_interval` will lead to the duration being ignored by the system. But if you think we should I am happy to add another error in?

tests/jax/components/executing/base_test.py
RuanJohn: Done :)

mava/components/jax/executing/base.py
RuanJohn: We were following the convention that [was in dev](https://github.com/instadeepai/Mava/blob/9846ebf90722afba7888c448d139503be198e8e5/mava/components/jax/executing/base.py#L49)  where the `inteval` variable was set using an underscore.

mava/components/jax/building/parameter_client.py
DriesSmit: Great, thanks :)

tests/jax/components/building/parameter_client_test.py
DriesSmit: Addressed above, thanks.

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
AsadJeewa: Though once we merge, it will work at the links will be created so I think you could add them in now

mava/components/jax/executing/base.py
AsadJeewa: Makes sense. We discussed this offline. The main thing is that the code must not crash which it was explained, will not happen

mava/components/jax/executing/base.py
AsadJeewa: It is not a major thing but since the variable is being used externally, without accessor/ mutators needed, I do not think it should be marked private with an underscore

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_eval_intervals.py
RuanJohn: Sorted :)

mava/components/jax/executing/base.py
RuanJohn: Is it okay if we leave it as is then?

mava/environment_loop.py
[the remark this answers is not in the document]
AsadJeewa: We should add this and a general explanation of evaluation to the docs. In another PR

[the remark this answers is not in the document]
OmaymaMahjoub: Yeah we will, once we finish the full evaluation necessary elements (absolute metric, extracting json file...)

mava/components/jax/executing/base.py
OmaymaMahjoub: Done!

AsadJeewa: Thanks

mava/environment_loop.py
```diff
@@ -550,22 +554,23 @@ def should_terminate(episode_count: int, step_count: int) -> bool:
...
             """
             should_run_loop = False
             eval_interval_key, eval_interval_count = eval_condition
             counts = self.get_counts()
+
             if counts:
-                count = counts.get(eval_interval_key)
-                # We run eval loops around every eval_interval_count (not exactly every
-                # eval_interval_count due to latency in getting updated counts).
+                count = counts[eval_interval_key]
```
AsadJeewa, line 571: This is not a problem for this PR but the eval interval specified and actual is quite different which is not ideal for papers
For example, when interval is 2000, 
``[evaluator/0] Running eval loop at executor step: 119350
[evaluator/0] Running eval loop at executor step: 122150
[evaluator/0] Running eval loop at executor step: 125050``

mava/environment_loop.py
```diff
@@ -575,28 +580,61 @@ def should_run_loop(eval_condtion: Tuple) -> bool:
...
-                result = self.run_episode()
-                episode_count += 1
-                step_count += result["episode_length"]
-                # Log the given results.
-                self._logger.write(result)
+            if (not environment_loop_schedule) or (should_run_loop(eval_condition)):
+                # TODO (Ruan): Remove store check once TF is deprecated.
+                if environment_loop_schedule and hasattr(self._executor, "store"):
+                    # Get first result dictionary
+                    results = self.run_episode()
```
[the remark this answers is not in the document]
AsadJeewa: Discussed offline. Needed to compute average

mava/environment_loop.py
```diff
@@ -575,28 +580,61 @@ def should_run_loop(eval_condtion: Tuple) -> bool:
...
+                    episode_count += 1
```
AsadJeewa, line 613: As discussed, there is a bug where steps and episodes are not updating only on the node itself i.e. on the evaluator node, evaluator steps and evaluator episodes is always 0. Same for executor

mava/environment_loop.py
```diff
@@ -575,28 +580,61 @@ def should_run_loop(eval_condtion: Tuple) -> bool:
...
+                )
+                evaluation_duration = self._executor.store.evaluation_duration
+        else:
+            environment_loop_schedule = (
+                self._executor._evaluator and self._executor._interval
+            )
+            if environment_loop_schedule:
+                eval_condition = check_count_condition(self._executor._interval)
 
         while not should_terminate(episode_count, step_count):
```
AsadJeewa, line 607: As discussed, we should remove this, perhaps in another PR

mava/components/jax/executing/base.py
```diff
@@ -46,7 +47,18 @@ def on_execution_init_start(self, executor: SystemExecutor) -> None:
...
+            self.config.evaluation_interval is None
+        ):
+            raise ValueError(
+                "Missing evaluation interval value: Evaluation duration"
+                + " value was provided without a value for the evaluation interval."
+            )
+
+        if executor.store.is_evaluator:
+            executor.store.evaluation_interval = self.config.evaluation_interval  # type: ignore # noqa: E501
+            executor.store.evaluation_duration = self.config.evaluation_duration  # type: ignore # noqa: E501
```
AsadJeewa, line 61: think this variable should be explicitly called evaluation_episode_duration

mava/environment_loop.py
```diff
@@ -575,28 +580,61 @@ def should_run_loop(eval_condtion: Tuple) -> bool:
...
+                    # Get first result dictionary
+                    results = self.run_episode()
+                    episode_count += 1
+                    step_count += results["episode_length"]
+                    for _ in range(evaluation_duration - 1):
+                        # Add consecutive evaluation run data
+                        result = self.run_episode()
+                        episode_count += 1
+                        step_count += result["episode_length"]
+                        results = jax.tree_map(lambda x, y: x + y, results, result)
```
AsadJeewa, line 620: Maybe add a comment here saying compute the mean (like below). I do think that sum would make more sense but as discussed, this messes with speed calculations. Though we could maybe do something like dividing roughly by a factor of `duration` for speed comparisons

mava/wrappers/smac.py
```diff
@@ -281,7 +284,24 @@ def get_stats(self) -> Optional[Dict]:
...
+        return stats
+
+    def get_extra_stats(self) -> Optional[Dict]:
+        """Return extra stats to be logged.
+
+        Returns:
+           Win rate
+        """
+        extra_stats: Dict[str, Any] = {}
+        extra_stats["win_rate"] = (
```
AsadJeewa, line 299: Nice! :rocket:

AsadJeewa (changes requested):
Thanks @OmaymaMahjoub and @RuanJohn 
Small changes requested. Also the bug on the logging.

mava/wrappers/smac.py
```diff
@@ -281,7 +284,24 @@ def get_stats(self) -> Optional[Dict]:
         Returns:
             extra stats to be logged.
         """
-        return self._environment.get_stats()
+        stats = self._environment.get_stats()
+        stats["cumulative_win_rate"] = stats["win_rate"]
```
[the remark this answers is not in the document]
OmaymaMahjoub: I think we need to keep it because it is the predefined metric in the smac environment, what do you think?

mava/environment_loop.py
RuanJohn: Fixed. There was a bug in the way the get and set keys were getting defined in the parameter client.

mava/environment_loop.py
RuanJohn: Yes, I think perhaps a separate PR. Ideally we can do this once TF is removed.

mava/environment_loop.py
RuanJohn: Done

mava/components/jax/executing/base.py
RuanJohn: I think it is nice to leave it as is for now so that we have flexibility to change it to a dictionary as well in the future? Then we could evaluate over number of steps or even for a fixed amount of time. What are your thoughts?

mava/wrappers/smac.py
```diff
@@ -281,7 +284,24 @@ def get_stats(self) -> Optional[Dict]:
...
+        del stats["win_rate"]
+        return stats
+
+    def get_extra_stats(self) -> Optional[Dict]:
```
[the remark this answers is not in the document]
RuanJohn: I'm not sure this is necessary because not all environments will have these interval stats?

mava/environment_loop.py
RuanJohn: The most common seems to be to evaluate after a set amount of executor steps for a set amount of episodes. But we have the flexibility to trigger based on any count parameter.

mava/wrappers/smac.py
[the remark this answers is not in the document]
AsadJeewa: Fair point. We could add a method to return an empty dict in the superclass then we would not need the hasattr if statement butthis is fine too

mava/environment_loop.py
AsadJeewa: Agreed

mava/components/jax/executing/base.py
AsadJeewa: I guess that is fair but until we change it to a dict, it can be confusing for new people to the code

mava/wrappers/smac.py
```diff
@@ -281,7 +284,24 @@ def get_stats(self) -> Optional[Dict]:
...
+        """Return extra stats to be logged.
```
AsadJeewa, line 293: Also, please change this docstring as it has the same description as `get_stats` above

OmaymaMahjoub: Okay!

OmaymaMahjoub: Done!

DriesSmit (dismissed):
Thanks @RuanJohn and @OmaymaMahjoub :raised_hands: This is a really nice update to Mava :slightly_smiling_face:

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
```diff
@@ -74,12 +70,13 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
         time_delta=log_every,
     )
 
-    # Optimisers.
+    # Optimiser.
```
DriesSmit, line 73: Maybe change this back to Optimisers.

mava/wrappers/smac.py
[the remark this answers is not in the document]
AsadJeewa: Ah okay that is fine then

mava/components/jax/building/parameter_client.py
```diff
@@ -93,8 +93,11 @@ def on_building_executor_parameter_client(self, builder: SystemBuilder) -> None:
             builder: SystemBuilder.
         """
         # Create policy parameters
-        params = {}
-        get_keys = []
+        params: Dict[str, Any] = {}
+        # Executor does not explicitly set variables i.e. it adds to count variables
```
DriesSmit, line 97: Interesting that this pops up? I thought @AsadJeewa pushed this code to develop?

AsadJeewa: I added the fix to this branch, not develop

mava/systems/jax/parameter_client.py
```diff
@@ -129,6 +129,23 @@ def get_async(self) -> None:
...
+    def force_get_async(self) -> None:
+        """Force asynchronously updates the parameters with the latest copy from server.
+
+        Returns:
+            None.
+        """
+        if self._get_future is None:
+            # The update period has been reached and no request has been sent yet, so
+            # making an asynchronous request now.
+            self._get_future = self._async_request()
```
DriesSmit, line 141: Do we need an async case?

mava/components/jax/executing/base.py
DriesSmit: Or maybe `num_episodes_per_eval`? Otherwise, we can change it to a dict now. Making it more specific is probably a good idea. But I get what both of you are saying. Maybe we can create a dictionary of one entry in that case? @RuanJohn @OmaymaMahjoub what do you think?

mava/systems/jax/parameter_client.py
OmaymaMahjoub: I will investigate it and give it a try by making it an immediate request

mava/wrappers/smac.py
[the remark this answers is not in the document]
DriesSmit: I guess we can probably leave it as is. I see the super class only has three abstract methods at the moment. @AsadJeewa I can't find `SingleEnvWrapper` in the codebase :smiling_face_with_tear:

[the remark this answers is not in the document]
AsadJeewa: ah sorry I meant `SequentialEnvWraper`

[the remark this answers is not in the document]
AsadJeewa: Yeah we definitely need to discuss superclasses. For that matter, I think that the base class should have much more content like `get_stats` for example. But perhaps we can do so in another PR

examples/jax/smac/feedforward/decentralised/run_ippo_eval_intervals.py
RuanJohn: Done.

DriesSmit: Thanks.

mava/components/jax/executing/base.py
OmaymaMahjoub: Done here are [the changes ](https://github.com/instadeepai/Mava/pull/709/commits/1943dcb7569bbfa4bf8c27df3fc2362176d36244)!

mava/wrappers/smac.py
```diff
@@ -281,7 +284,29 @@ def get_stats(self) -> Optional[Dict]:
...
+            computed over a given number of evaluation episodes.
+
+        An example would be a win rate calculation where it is required to
+        keep track of the number of games won over a given set of evaluation
+        episodes.
+
+        Returns:
+           Win rate
+        """
+        extra_stats: Dict[str, Any] = {}
```
AsadJeewa, line 303: Think it would make more sense to call this dict interval_stats

AsadJeewa (changes requested):
small change requested
Thanks @DriesSmit @OmaymaMahjoub

mava/wrappers/smac.py
OmaymaMahjoub: probably evaluation_stats?

AsadJeewa: but the get_stats method also returns evaluation stats

OmaymaMahjoub: Yeah you are right the win rate is related to an interval stats I will do that

OmaymaMahjoub: Done!

mava/systems/jax/parameter_client.py
OmaymaMahjoub: Done!

DriesSmit: Thanks @OmaymaMahjoub :fire:

DriesSmit (approved):
Thanks, @RuanJohn and @OmaymaMahjoub :fire: I'm happy from my side :slightly_smiling_face:

AsadJeewa (approved):
Thank you @OmaymaMahjoub and @RuanJohn 
Important PR!!!

PR #712, opened 2022-09-02
Feature: Make separate networks ippo build from shared networks ippo.

### LINKED ISSUE #707: [FEATURE] Make ippo with separate and shared networks more similar
Get ippo with separated and shared networks to use more of the same components. 

Try to get both systems to use the same parameter client and parameter server components. Furthermore, maybe make one of the systems use the other system as a basis. E.g. the shared networks IPPO uses the independent network IPPO as a basis, but updates the trainer components.

DriesSmit:
## What?
Make the separate networks ippo system build from the shared networks ippo.
## Why?
This shows how components can be used to upgrade a system.
## How?
Changed the ippo system file.
## Extra
The network sizes are now also updated to be an exponent of 2.
Close #707

### DIFF
--- a/examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo.py
+++ b/examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo.py
@@ -64,7 +64,7 @@ def main(_: Any) -> None:
     # Networks.
     def network_factory(*args: Any, **kwargs: Any) -> Any:
         return ippo.make_default_networks(  # type: ignore
-            policy_layer_sizes=(254, 254, 254),
+            policy_layer_sizes=(256, 256, 256),
             critic_layer_sizes=(512, 512, 256),
             *args,
             **kwargs,
--- a/mava/components/jax/building/parameter_client.py
+++ b/mava/components/jax/building/parameter_client.py
@@ -311,21 +311,7 @@ class ExecutorParameterClientSeparateNetworks(ExecutorParameterClient):
             parameter_client.get_and_wait()
 
         builder.store.executor_parameter_client = parameter_client
-
-    @staticmethod
-    def name() -> str:
-        """Component type name, e.g. 'dataset' or 'executor'."""
-        return "executor_parameter_client"
-
-    @staticmethod
-    def config_class() -> Optional[Callable]:
-        """Config class used for component.
-
-        Returns:
-            config class/dataclass for component.
-        """
-        return ExecutorParameterClientConfig
-
+        
 
 class TrainerParameterClientSeparateNetworks(TrainerParameterClient):
     def __init__(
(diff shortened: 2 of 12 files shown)


RuanJohn (approved):
Thanks @DriesSmit, this look really good! :muscle:

mava/specs.py
```diff
@@ -170,3 +170,11 @@ def get(self) -> Dict[str, Any]:
             _description_
         """
         return self.__dict__
+
+    def set(self, name: str, component: Any) -> None:
+        """_summary_
```
RuanJohn, line 175: Can we please add a descriptive dosctring here?

RuanJohn (changes requested):
Thank you @DriesSmit! It looks :fire: Just one minor request from my side.

mava/components/jax/building/parameter_client.py
```diff
@@ -238,7 +238,7 @@ def name() -> str:
 ###################
 
 
-class ExecutorParameterClientSeparateNetworks(BaseParameterClient):
+class ExecutorParameterClientSeparateNetworks(ExecutorParameterClient):
```
AsadJeewa, line 241: I get why you did this but the code in this class is exactly the same as ExecutorParameterClient besides when we create the networks so if we extend this we should reuse code or extend base

AsadJeewa (commented):
Thanks for this @DriesSmit 
Small comment

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo.py
```diff
@@ -64,7 +64,7 @@ def main(_: Any) -> None:
     # Networks.
     def network_factory(*args: Any, **kwargs: Any) -> Any:
         return ippo.make_default_networks(  # type: ignore
-            policy_layer_sizes=(254, 254, 254),
```
AsadJeewa, line 67: Out of interest, would be good to know why this was set to54. Obviously we want a power of 2 but weird that it was set to this

DriesSmit: I think it was just an arbitrary choice :)

mava/specs.py
DriesSmit: Fixed :+1:

mava/components/jax/building/parameter_client.py
DriesSmit: Nice catch @AsadJeewa :smile: I updated the code. Does it look better now?

mava/specs.py
RuanJohn: Thank you :)

RuanJohn (dismissed):
Thank you @DriesSmit :muscle:

AsadJeewa (approved):
Looks good. Thanks, @DriesSmit

mava/components/jax/building/parameter_client.py
AsadJeewa: I am happy

RuanJohn (approved):
Thank you @DriesSmit :muscle:

PR #717, opened 2022-09-05
Fix update period

### LINKED ISSUE #697: [FEATURE] Updated update_period
Use a large value for the update period in the ParameterClient. The current setting might affect performance as the trainer and executors pull and push parameters too often.

Set the default update_period to e.g. 10. Or set it to be the same as used for the TensorFlow systems.

DriesSmit:
## What?
Fix the `update_period` flag for all the JAX and TF systems. The various systems inside JAX and TF have inconsistent values for both the executor and trainer.
## Why?
This is to make testing between the systems more fair.
## How?
Updated the `update_period` flags throughout Mava.
## Extra
Close #697


RuanJohn (approved):
Thank you @DriesSmit! This looks good to me :+1:

tests/jax/components/building/parameter_client_test.py
```diff
@@ -562,7 +562,7 @@ def test_trainer_parameter_client_separate_networks(
     assert mock_builder.store.trainer_parameter_client._get_call_counter == 0
     assert mock_builder.store.trainer_parameter_client._set_call_counter == 0
     assert mock_builder.store.trainer_parameter_client._set_get_call_counter == 0
-    assert mock_builder.store.trainer_parameter_client._update_period == 1
+    assert mock_builder.store.trainer_parameter_client._update_period == 5
```
AsadJeewa, line 565: This assertion would fail if we change the default value of update_period in the config which is not ideal. We should rather set the value in the test and assert that it is the same as the set value

AsadJeewa: Same above. we should just do what we do with executor update period in the test

tests/jax/components/building/parameter_client_test.py
```diff
@@ -287,7 +287,7 @@ def test_executor_parameter_client_evaluator_with_parameter_client(
     mock_builder = mock_builder_with_parameter_client
     mock_builder.store.is_evaluator = True
     exec_param_client = ExecutorParameterClient(
-        config=ExecutorParameterClientConfig(executor_parameter_update_period=100)
+        config=ExecutorParameterClientConfig(executor_parameter_update_period=200)
```
AsadJeewa, line 290: not imp but why was this increased?

AsadJeewa (changes requested):
Thanks, @DriesSmit This is an important PR for comparing to TF. Minor comment and request

tests/jax/components/building/parameter_client_test.py
DriesSmit: Agreed. I updated the tests to reflect this.

tests/jax/components/building/parameter_client_test.py
DriesSmit: This was increased to match the assertion test: `assert mock_builder.store.executor_parameter_client._update_period == 200`

AsadJeewa (approved):
Thanks @DriesSmit 
All good from my side :smile:

RuanJohn (approved):
Thank you @DriesSmit!

PR #719, opened 2022-09-05
Fix JAX quickstart notebook

DriesSmit:
## What?
Fix the JAX quickstart notebook by setting the GPU memory preallocation to false. Thanks @RuanJohn for the suggestion :slightly_smiling_face: 
## Why?
The trainer runs out of GPU memory and crashes.
## How?
Updated the JAX notebook file.

### DIFF
--- a/examples/jax/quickstart.ipynb
+++ b/examples/jax/quickstart.ipynb
@@ -34,7 +34,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 10,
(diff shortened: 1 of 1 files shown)


RuanJohn (approved):
Thank you @DriesSmit :muscle:

OmaymaMahjoub (approved):
Runned the file and it works properly, Thanks @DriesSmit!

PR #725, opened 2022-09-07
Fix: Provide unique rng keys to each process

### LINKED ISSUE #724: [BUG] Processes start with the same ring keys
### Describe the bug
Every process is starting with exactly the same ring key. This means that every executor is generating exactly (or almost exactly) the same episodes.

### Possible Solution
Split the initial builder key into a unique key per process and make sure each process uses its own key.

DriesSmit:
## What?
This PR splits the initial builder key into a unique key per process.

## Why?
Every process is starting with exactly the same rng key. This means that every executor is generating exactly (or almost exactly) the same episodes.

## How?
Updated the distributor component and builder class.

## Extra
Close #724


KaleabTessera (dismissed):
Thanks @DriesSmit , super useful!

RuanJohn (approved):
Looks good to me. Thank you @DriesSmit :muscle:

RuanJohn (approved):
Thank you @DriesSmit :muscle:

mava/systems/jax/builder.py
```diff
@@ -114,6 +120,15 @@ def executor(
         self.store.parameter_server_client = parameter_server_client
         self.store.is_evaluator = self.store.executor_id == "evaluator"
 
+        if self.store.is_evaluator:
+            # Set the rng key for the evaluator.
+            self.store.base_key = self.store.eval_key
+        else:
+            # Set the rng key for the executor.
+            self.store.base_key = self.store.executor_keys[
```
AsadJeewa, line 128: I see that you did not add the len check for multiple executors here like you do for the trainer since we name the executor with an _index always. I think we should do the same with the trainers in ditributor.py instead of just assigning the name trainer https://github.com/instadeepai/Mava/blob/develop/mava/components/jax/building/distributor.py#L109

mava/systems/jax/builder.py
```diff
@@ -168,6 +183,14 @@ def trainer(
             System trainer.
         """
 
+        # Set the rng key for the trainer.
+        if len(self.store.trainer_keys) > 1:
```
AsadJeewa, line 187: This check could then be moved based on above

tests/jax/components/building/distributor_test.py
```diff
@@ -110,6 +115,24 @@ def test_on_building_program_nodes_multi_process(
...
+            mock_builder.store.eval_key,
+        ]
+        + list(mock_builder.store.executor_keys)
+        + list(mock_builder.store.trainer_keys)
+    )
+    start_keys = [float(key[0]) for key in keys]
+    end_keys = [float(key[1]) for key in keys]
+
+    assert len(set(start_keys)) == len(start_keys)
+    assert len(set(end_keys)) == len(end_keys)
```
AsadJeewa, line 132: nice :smile:

AsadJeewa (changes requested):
Thanks for this! All looks good. Just a small request @DriesSmit

mava/systems/jax/builder.py
DriesSmit: Good idea, thanks @AsadJeewa :slightly_smiling_face: I updated the code.

mava/systems/jax/builder.py
DriesSmit: I updated the code :+1:

tests/jax/components/building/distributor_test.py
DriesSmit: Thanks :slightly_smiling_face:

RuanJohn (approved):

AsadJeewa (approved):
Thanks @DriesSmit

AsadJeewa (approved):
lol made things a little frustrating for you with all the test updates needed to make the trainer change. Thanks @DriesSmit

KaleabTessera (approved):

PR #736, opened 2022-09-12
Feature/remove config classes

DriesSmit:
## What?
Remove all `config_class` functions.
## Why?
This simplifies the code as now components do not require `config_class` functions. This also fixes a bug where some components accidentally did not have config classes. In this case, their variables were not updatable through the system config. 
## How?
Updated the [config.py](https://github.com/instadeepai/Mava/blob/develop/mava/systems/jax/config.py) and [system.py](https://github.com/instadeepai/Mava/blob/develop/mava/systems/jax/system.py) files to use `component.__init__.__annotations__["config"]` instead of `component.config_class()`.
## Extra
I also updated the [gotchas.md](https://github.com/instadeepai/Mava/blob/develop/docs/gotchas.md) docs to highlight the importance of typing when creating components.


mava/systems/jax/config.py
```diff
@@ -223,8 +226,9 @@ def get_local_config(self, component: Type[Component]) -> SimpleNamespace:
         global_config = self._built_config
         local_config: Dict[str, Any] = {}
 
-        config_class = component.config_class()
-        if not config_class:  # no config class for component
+        config_class = component.__init__.__annotations__["config"]
+
+        if config_class is SimpleNamespace:  # no config class for component
```
AsadJeewa, line 231: This if statement is redundant though I get that you want to return before the for loop

AsadJeewa (commented):
Looks good @DriesSmit 
Small comment

docs/gotchas.md
```diff
@@ -20,3 +20,23 @@ In order to have access to the underlying jax.numpy arrays that are in these var
 with jax.disable_jit():
     pdb.set_trace()
 ```
+
+## Components
+When creating components, it is important to include variable types for both the config class and component init function. This is necessary because Mava uses it to determine which config variables are available in the system. In the example below, we provide the types for the config variables and also the config class variable inside the `Distributor` component. For the config class variable we use the `DistributorConfig` as the type, which points the system to the config class that is used.
```
RuanJohn, line 25: When creating components, it is important to include variable types for both the config class and component init function => When creating components, it is important to include variable types for both the config class and component class `__init__` method.

mava/systems/jax/config.py
```diff
@@ -16,6 +16,7 @@
 """Config class for Mava systems"""
 
 from dataclasses import fields, is_dataclass
+from distutils.command.config import config
```
RuanJohn, line 19: This looks like an accidental import? If not, please ignore.

DriesSmit: Sure is. Great catch :raised_hands:

docs/gotchas.md
DriesSmit: Updated :+1:

mava/systems/jax/config.py
DriesSmit: The code is to return before the loop if necessary, yes. Is there a better way of writing this that you can think of, @AsadJeewa? I updated the comments and code layout.

AsadJeewa (dismissed):
This removes a lot of boilerplate! Thanks @DriesSmit

mava/systems/jax/config.py
AsadJeewa: The latest change is better

AsadJeewa (dismissed):

mava/components/jax/building/adders.py
```diff
@@ -115,22 +102,17 @@ def required_components() -> List[Type[Callback]]:
...
     def __init__(
         self,
-        config: AdderSignatureConfig = AdderSignatureConfig(),
+        config: SimpleNamespace = SimpleNamespace(),
     ):
         """Component creates an adder signature for reverb adders.
 
         Args:
-            config: AdderSignatureConfig.
+            config: SimpleNamespace.
```
AsadJeewa, line 115: I noticed inconsistencies in the past so glad about this change. Is it, make it SimpleNamespace if not config was defined instead of needing to add a class and say pass?

AsadJeewa (dismissed):
Beautiful PR! Thanks for this @DriesSmit 
I really like that we are consistent now with empty configs also

RuanJohn (approved):
Thank you @DriesSmit :muscle:

mava/components/jax/building/adders.py
DriesSmit: Yes :slightly_smiling_face:

DriesSmit: The empty configs was @RuanJohn 's idea :fire:

AsadJeewa (approved):
Thanks @DriesSmit

mava/components/jax/building/adders.py
AsadJeewa: Nice @RuanJohn !

Issue #722, opened 2022-09-06
[FEATURE] Add option to normalise critic values during training


ulricharmel:
As suggested in this [paper](https://arxiv.org/pdf/2103.01955.pdf), standardize the target values for the critic loss can improve results in certain MARL environments.

PR #741, opened 2022-09-13
Feature/remove shared networks

DriesSmit:
## What?
Remove the shared policy and critic network option from IPPO.
## Why?
To simplify the code, we will only be evaluating IPPO using separate networks for the JAX release. If necessary, we can add the shared networks code back later.
## How?
Remove all code associated with shared networks IPPO. Renamed the separate networks components by removing the shared networks tag.

PR #742, opened 2022-09-13
Implemented target value normalization during training in the IPPO sy…

ulricharmel:
## What?
Add components needed to normalise target values during training for jax IPPO system. 
## Why?
Normalising target values as stated in the Yu et al (2021) paper can significantly improve training for certain environments. 
## How?

- Added functions in jax_training_utils.py to compute running statistics (mean and variance) of the data, normalise and denormalise.
- Added a new GAE function which accounts for the normalisation in the adavantage_estimation.py
- Added new Training parameters to store the running statistic.
- Added two new step functions in step.py to perform the normalisation during training for the shared and separate networks. 
- Added two examples the the Mava/examples/jax/debugging/simple_spread/feedforward/decentralised. 
## Extra
Benchmarking results are available [here.](https://www.notion.so/instadeep/Experiments-52346a5de10d444583896c8d5b026b25?p=86e13746fa544f5194bd8a2a9fbedc5e&pm=s) 
Closed [#722 ](https://github.com/instadeepai/Mava/issues/722)


DriesSmit (changes requested):
Thanks @ulricharmel :fire: Great to see your first PR up :slightly_smiling_face:  Could you maybe add a test that specifically tests that your value normalisation component is working?

mava/components/jax/training/advantage_estimation.py
```diff
@@ -111,3 +112,57 @@ def required_components() -> List[Type[Callback]]:
             List of required component classes.
         """
         return []
```
DriesSmit, line 114: Maybe change this to a pass. Similar to the other components we have.

mava/components/jax/training/step.py
```diff
@@ -719,3 +951,283 @@ def config_class() -> Optional[Callable]:
...
+            }
+            critic_params = {
+                net_key: networks[net_key].critic_params for net_key in networks.keys()
+            }
+            policy_opt_states = trainer.store.policy_opt_states
+            critic_opt_states = trainer.store.critic_opt_states
+            random_key, _ = jax.random.split(trainer.store.key)
+            stats = trainer.store.stats
+
+            states = TrainingStateStatsSeparateNetworks(
```
DriesSmit, line 1175: Once we get this [PR](https://github.com/instadeepai/Mava/pull/741), you can remove support for shared networks. Therefore you can remove the SeperateNetworks tags.

mava/components/jax/training/advantage_estimation.py
```diff
@@ -27,6 +27,7 @@
 from mava.components.jax.training.base import Utility
 from mava.core_jax import SystemTrainer
 
+from mava.utils.jax_training_utils import normalize, denormalize
 
 @dataclass
 class GAEConfig:
```
DriesSmit, line 33: Looks good. We recently updated our code to use SimpleNamespaces instead. So you can maybe remove all the empty dataclasses and just use a SimpleNamespace.

DriesSmit (commented):
Can you just link the correct issue as well, please :slightly_smiling_face:

mava/components/jax/training/step.py
ulricharmel: I have removed the support of separate Networks.

DriesSmit: Thanks :)

mava/components/jax/training/advantage_estimation.py
ulricharmel: I followed exactly the same approach as in the GAE function in the current main branch.

mava/components/jax/training/step.py
```diff
@@ -471,3 +472,314 @@ def required_components() -> List[Type[Callback]]:
             mava.components.jax.training.model_updating.MinibatchUpdate,
             mava.components.jax.building.adders.ParallelSequenceAdder,
         ]
+
+
+class MAPGWithTrustRegionStepNorm(Step):
```
DriesSmit, line 477: We would probably need to have a chat about this in Mava standup. The issue is that the entire training step is copied, which is the same as we did in the past with shared networks. When we add new features to the trainer step function we now need to implement the features in both step functions. Hopefully, we can find a slightly more straightforward way to do this without needing to copy the step code.

ulricharmel: Yes you are right. We should chat about it. I also also thought about it. We need to avoid having to maintain to parallel codes.

mava/utils/jax_training_utils.py
```diff
@@ -15,9 +15,89 @@ def action_mask_categorical_policies(
...
+    m_a = var * count
+    m_b = batch_var * batch_count
+    M2 = m_a + m_b + jnp.square(delta) * count * batch_count / tot_count
+    new_var = M2 / tot_count
+    new_count = tot_count
+
+    return jnp.array([new_mean, new_var, new_count])
+
+
+def dummy_running_mean_var_count(stats: Array, batch: Array) -> Array:
```
RuanJohn, line 51: Since this is getting used in the tests can we please define this function in one of the testing folders or even in the test where it is getting used?

mava/components/jax/training/step.py
DriesSmit: Great @ulricharmel :raised_hands: Thanks for the effort.

mava/components/jax/training/advantage_estimation.py
DriesSmit: See [here](https://github.com/instadeepai/Mava/blob/develop/mava/components/jax/executing/action_selection.py#L39) for an example.

mava/utils/jax_training_utils.py
ulricharmel: It doesn't do anything specific but I will add a test for it for completeness.

ulricharmel: @RuanJohn I have pushed the test

mava/components/jax/training/model_updating.py
```diff
@@ -93,6 +98,19 @@ def on_training_utility_fns(self, trainer: SystemTrainer) -> None:
...
+        if self.config.normalize_target_values:
+            trainer.store.running_stats_fn = compute_running_mean_var_count
+            running_stats = jnp.array([0, 0, 1e-4])
+        else:
+            running_stats = jnp.array([0, 1, 1e-4])
+            trainer.store.running_stats_fn = dummy_running_mean_var_count
+
+        trainer.store.stats = {}
+        for net_key in trainer.store.trainer_agent_net_keys.keys():
+            trainer.store.stats[net_key] = running_stats
```
arnupretorius, line 112: Do we capture stats other than the running stats here? Because if it is just the running stats, we could perhaps create the empty stats dict inside the if statement, remove the else statement and not need a dummy stats function? Then we only call the function from the store if it exists using something like `if hasattr(trainer.store, 'stats'): ...` In fact, this might be a recurring thing that happens across Mava, i.e. the store might have something or not, based on a boolean variable in the config of a component. So perhaps we can create a function for the store to check, e.g. `store.has("name_of_attribute")` which just returns `hasattr(self, "name_of_attribute")` to make checks like these neater. Let me know what you think. Also @KaleabTessera and @DriesSmit, your opinions as well 🙂

arnupretorius (commented):
Thanks @ulricharmel! 👍 Left a few comments but looking good 🙂

mava/components/jax/training/model_updating.py
ulricharmel: Yes I think that's a good idea to have a function that checks store attributes. I was trying to keep the code clean with the dummy function to avoid if statements later in the sgd_step function. But if the dummy function is confusing we can remove that and follow your suggestion. The running stats are initialised differently. The dummy ones that never change have variance of 1. In the other case the initial value is 0 so we will still need an else statement to capture the difference.

ulricharmel: @arnupretorius I removed the unnecessary dummy function.

mava/components/jax/training/advantage_estimation.py
```diff
@@ -75,12 +79,15 @@ def gae_advantages(
             rewards = jnp.clip(rewards, -max_abs_reward, max_abs_reward)
 
             advantages = rlax.truncated_generalized_advantage_estimation(
-                rewards[:-1], discounts[:-1], self.config.gae_lambda, values
+                rewards[:-1],
+                discounts[:-1],
+                self.config.gae_lambda,
+                denormalize(stats, values),
```
DriesSmit, line 85: Minor thing. Instead of denormalising twice we can probably just denormalise once and pass it to both functions that need it. I am happy to benchmark it after this point :slightly_smiling_face:

mava/utils/jax_training_utils.py
```diff
@@ -15,9 +15,73 @@ def action_mask_categorical_policies(
...
+
+    mean, var, count = stats[0], stats[1], stats[2]
+
+    delta = batch_mean - mean
+    tot_count = count + batch_count
+
+    new_mean = mean + delta * batch_count / tot_count
+    m_a = var * count
+    m_b = batch_var * batch_count
+    M2 = m_a + m_b + jnp.square(delta) * count * batch_count / tot_count
```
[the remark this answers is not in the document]
ulricharmel: It's the implementation of [Welfdord's](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm) algorithm for computing running mean and variance.  I just re-adapted the implementation from the [PPO baselines repo](https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/common/running_mean_std.py).

mava/components/jax/training/advantage_estimation.py
ulricharmel: I fixed this.

mava/utils/jax_training_utils.py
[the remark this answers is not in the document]
DriesSmit: I see, thanks :slightly_smiling_face:

mava/components/training/advantage_estimation.py
```diff
@@ -57,7 +58,10 @@ def on_training_utility_fns(self, trainer: SystemTrainer) -> None:
         """
 
         def gae_advantages(
-            rewards: jnp.ndarray, discounts: jnp.ndarray, values: jnp.ndarray
+            rewards: jnp.ndarray,
+            discounts: jnp.ndarray,
+            values: jnp.ndarray,
+            stats: jnp.ndarray = jnp.array([0, 1, 1e-4]),
```
KaleabTessera, line 64: I think you added docs for this below, but can you please add docstrings for this function.

mava/components/training/model_updating.py
```diff
@@ -93,6 +95,18 @@ def on_training_utility_fns(self, trainer: SystemTrainer) -> None:
             None.
         """
 
+        # Initilaise running statisitics here
+
+        if self.config.normalize_target_values:
+            trainer.store.running_stats_fn = compute_running_mean_var_count
+            running_stats = jnp.array([0, 0, 1e-4])
```
KaleabTessera, line 102: Looks we should take `running_stats = jnp.array([0, 0, 1e-4])` out the if statement since it applies in both conditions.

mava/utils/jax_training_utils.py
```diff
@@ -15,9 +15,73 @@ def action_mask_categorical_policies(
...
+
     return tfd.Categorical(logits=masked_logits, dtype=distribution.dtype)
 
 
+def compute_running_mean_var_count(
+    stats: jnp.ndarray, batch: jnp.ndarray
+) -> jnp.ndarray:
+    """Updates the running mean, variance and data counts during training.
+
+    stats (array) -- mean, var, count.
```
KaleabTessera, line 27: What is the purpose of the count param in stats? Like to prevent division by 0? If so maybe it should just be constant? 
That way we could make mean and var separate variables since passing in an array isn't always very readable.

mava/utils/jax_training_utils.py
```diff
@@ -15,9 +15,73 @@ def action_mask_categorical_policies(
...
+    mean, var = stats[0], stats[1]
+    normalize_batch = (batch - mean) / (jnp.sqrt(jnp.clip(var, a_min=1e-2)))
+
+    return normalize_batch
+
+
+def denormalize(stats: jnp.ndarray, batch: jnp.ndarray) -> jnp.ndarray:
+    """Transform normalized data back into original distribution
+
+    stats (array) -- mean, var, count
```
KaleabTessera, line 72: We don't seem to use the count here?

tests/components/training/value_normalisation_test.py
```diff
@@ -0,0 +1,39 @@
...
+    ]:
+
+        stats = jnp.array([0, 0, 1e-4])
+
+        stats = compute_running_mean_var_count(stats, jnp.array(x1))
+        stats = compute_running_mean_var_count(stats, jnp.array(x2))
+        stats = compute_running_mean_var_count(stats, jnp.array(x3))
+
+        x = jnp.array(np.concatenate([x1, x2, x3], axis=0))
+        stats2 = jnp.array([jnp.mean(x), jnp.var(x), x.size + 1e-4])
```
KaleabTessera, line 26: Still uncertain of the purpose of `+ 1e-4`.

mava/components/training/model_updating.py
ulricharmel: In the other case there is a slight difference we have 1 instead of 0 for the variance. 
`if self.config.normalize_target_values:
            trainer.store.running_stats_fn = compute_running_mean_var_count
            running_stats = jnp.array([0, 0, 1e-4])
        else:
            running_stats = jnp.array([0, 1, 1e-4])`

mava/utils/jax_training_utils.py
ulricharmel: We need to it keep track of the total number of samples. So that every time when computing the running mean and variance we divide by the correct factor.

tests/components/training/value_normalisation_test.py
ulricharmel: It's actually not necessary. I will remove that.

mava/components/training/advantage_estimation.py
ulricharmel: They was already a docstring but I have added more comments.

mava/components/jax/training/model_updating.py
arnupretorius: Thanks @ulricharmel , will take a look :)

arnupretorius (dismissed):
Thanks @ulricharmel, the changes look good to me 👍

DriesSmit (approved):
Thanks @ulricharmel :partying_face:

arnupretorius (approved):
Thanks @ulricharmel 👍

PR #751, opened 2022-09-14
Bugfix: Convert actions to python integers before stepping MPE environment

### LINKED ISSUE #745: [BUG] PettingZoo MPE environments action not in action space
### Describe the bug
When running the `simple_spread_v2`, `simple_reference_v2` and `simple_speaker_listener_v3` PettingZoo `MPE`
environments the following error is raised:
```
AssertionError: action is not in action space
```
even though the actions that were taken are within the action space bounds given in the PettingZoo documentation [here](https://www.pettingzoo.ml/mpe/simple_spread), [here](https://www.pettingzoo.ml/mpe/simple_reference) and [here](https://www.pettingzoo.ml/mpe/simple_speaker_listener).  

### To Reproduce
1. Run any of the PettingZoo examples mentioned above

### Expected behavior
The action space error mentioned abobe will be raised. 

### Context (Environment)
 - A clean conda environment with Mava installed directly from the develop branch using `pip install -e .[jax,reverb,pz]`

### Possible Solution
It could be that Jax device arrays are not able to be passed to the environment?

RuanJohn:
## What?
Small change made to the PettingZoo wrapper where agent actions are converted to Python integers from Jax device arrays right before stepping the environment. 
## Why?
Jax device array actions cause an error where PettingZoo says that valid actions are not in the action space. 
## How?
Convert actions using the `tolist()` method on the Jax device arrays. 
## Extra
Added examples for running IPPO on the `simple_spread_v2`, `simple_reference_v2` and `simple_speaker_listener_v3` PettingZoo MPE environments
Close #745


mava/wrappers/pettingzoo.py
```diff
@@ -444,6 +445,11 @@ def step(self, actions: Dict[str, np.ndarray]) -> dm_env.TimeStep:
 
         # Get valid actions for active agents
         actions = {key: actions[key] for key in self.agents}
+
+        # Convert Jax device array actions to python integers
+        if not all(type(value) == int for value in actions.values()):  # type: ignore
```
KaleabTessera, line 450: Maybe we should consider doing this in the executor? I had this in a different branch - https://github.com/instadeepai/Mava/commit/19c19e27adb2e4ecfe3da144b469e31e23226888. What do you think? We might need to benchmark this.

AsadJeewa (approved):
Thanks, @RuanJohn 
Important PR! Runs on my side

mava/wrappers/pettingzoo.py
RuanJohn: I considered doing this, but the problem seems to only be affecting PettingZoo. By only changing right before stepping the environment we also still store device arrays for use during training later on? What are your thoughts?

KaleabTessera (approved):
Thanks @RuanJohn , this is a great fix! :fire:

PR #754, opened 2022-09-14
feat: add huber loss from rlax.

### LINKED ISSUE #729: [FEATURE] Huber loss
Implement huber loss for the JAX IPPO system.

siddarthsingh1:
## What?
Huber loss added to IPPO.
## Why?
Further optimisation of MAVA's IPPO algorithm
## How?
Add option to use a huber loss instead of MSE.
## Extra
Closed #729 
[Benchmarking results](https://www.notion.so/instadeep/Huber-loss-for-value-function-b3b0ae1d1c21423a8da8b3b589bc6065)


DriesSmit:
Thanks, @siddarthsingh1 :slightly_smiling_face: Can you just link the issue as well :+1:

mava/systems/jax/ippo/system.py
```diff
@@ -94,3 +94,71 @@ def design(self) -> Tuple[DesignSpec, Any]:
             component_dependency_guardrails=ComponentDependencyGuardrails,
         )
         return system, default_params
+
```
OmaymaMahjoub, line 97: Change the name of the file to systems or separate the two systems each in a py file since we have a full folder called systems

OmaymaMahjoub (changes requested):
Looks good thanks @siddarthsingh1, it will be interesting if you add test case for the huber loss in the unit tests of `lesses_test.py` (whether in this PR or you can make a new PR for it)

tests/jax/components/training/losses_test.py
```diff
@@ -239,7 +239,6 @@ def test_mapg_loss_separate_networks(
     low_loss_critic = low_agent_0_critic_loss["loss_critic"]
 
     assert jnp.isclose(loss_entropy, -0.47500002)
-    assert loss_critic == 4.5
```
DriesSmit, line 242: Can you maybe add a test for the Huber loss? Instead of needing to remove this assert?

mava/components/jax/training/losses.py
```diff
@@ -166,6 +168,252 @@ def policy_loss_fn(
                 )
             return policy_grads, loss_info_policy
 
+        def critic_loss_grad_fn(
```
DriesSmit, line 171: Just maybe chat to @KaleabTessera about this. I think he was thinking one could provide a function to the system instead of having to duplicate the loss code?

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_huber.py
```diff
@@ -0,0 +1,117 @@
...
+    policy_optimiser = optax.chain(
+        optax.clip_by_global_norm(40.0), optax.scale_by_adam(), optax.scale(-1e-4)
+    )
+
+    critic_optimiser = optax.chain(
+        optax.clip_by_global_norm(40.0), optax.scale_by_adam(), optax.scale(-1e-4)
+    )
+
+    # Create the system.
+    system = ippo.IPPOSystemHuber()
```
DriesSmit, line 94: Maybe instead of creating a new system we can just do something like [this](https://github.com/instadeepai/Mava/blob/feature/recurrent-actors/examples/jax/debugging/simple_spread/recurrent/decentralised/run_ippo.py#L95). Otherwise we might need a new system for every configuration of the components.

mava/systems/jax/ippo/system.py
```diff
@@ -94,3 +94,71 @@ def design(self) -> Tuple[DesignSpec, Any]:
...
+
+class IPPOSystemHuber(System):
```
DriesSmit, line 99: We can probably remove this system and just add the Huber component in the example.

DriesSmit (changes requested):
Thanks @siddarthsingh1 :fire: Nice work :slightly_smiling_face:

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_huber.py
siddarthsingh1: Updated it to add the huber loss through the update instead of a separate system.

mava/systems/jax/ippo/system.py
siddarthsingh1: Removed.

tests/jax/components/training/losses_test.py
siddarthsingh1: I've added tests for the separated functions and readded the assert.

mava/systems/jax/ippo/system.py
siddarthsingh1: The extra system has been removed

mava/components/jax/training/losses.py
siddarthsingh1: I'll ask about that. I might have misinterpreted what was being discussed :sweat_smile: .

tests/jax/components/training/losses_test.py
DriesSmit: Thanks :)

examples/jax/debugging/simple_spread/feedforward/decentralised/run_ippo_huber.py
DriesSmit: Thanks :slightly_smiling_face:

mava/components/jax/training/losses.py
DriesSmit: Thanks :)

DriesSmit (dismissed):
Thanks @OmaymaMahjoub :fire:

OmaymaMahjoub:
> Thanks @OmaymaMahjoub fire

Thanks @DriesSmit for helping with the redesign!

mava/systems/config.py
```diff
@@ -136,13 +137,13 @@ def update(self, **kwargs: Any) -> None:
                         self._config[name] = dataclass
                 else:
                     raise Exception(
-                        f"The given component config ({name}) is not part of the current \
```
DriesSmit: Why was this removed?

mava/systems/config.py
```diff
@@ -85,7 +85,9 @@ def add(self, **kwargs: Any) -> None:
                             self._param_to_component[new_param_name] = name
                     self._config[name] = dataclass
             elif isinstance(dataclass, SimpleNamespace):
-                self._config[name] = dataclass
+                # SimpleNamespace implies that
```
DriesSmit, line 88: Why was this changed?

mava/systems/system.py
```diff
@@ -58,22 +57,17 @@ def _make_config(self) -> None:
...
-            self.config.add(**input)
 
-    @abc.abstractmethod
-    def design(self) -> Tuple[DesignSpec, Dict]:
-        """System design specifying the list of components to use.
-
-        Returns:
-            Tuple[system callback components, system config].
-        """
+            if config_class is not SimpleNamespace:
```
DriesSmit, line 61: Why did we add an if statement here? We still need to config right?

DriesSmit, line 61: Why did we add an if statement here? We still need to config right?

mava/systems/config.py
OmaymaMahjoub: It was because merge conflict edit I will restore it

mava/systems/system.py
OmaymaMahjoub: I just passed the systempy from develop branch

OmaymaMahjoub: I will check how to fix it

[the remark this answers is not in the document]
OmaymaMahjoub: This related to the develop branch it is not a change made in this PR :/

mava/components/component_groups.py
```diff
@@ -1,7 +1,12 @@
...
 
 # The components that are needed for a recurrent policy.
 recurrent_policy_components = [
     executing.RecurrentExecutorSelectAction,
     executing.RecurrentExecutorObserve,
 ]
+
+# The components that a needed for IPPO with the Huber value loss.
+huber_value_loss_components = [
+    training.HuberValueLossFunction,
```
arnupretorius, line 11: If there is just one component for the loss, should we be adding it to a "groups" file? 

Also, I think the group components here are general (since they are located inside the general components folder), so we can remove the string saying huber is just for IPPO, since other systems might want to use it as well.

mava/components/training/losses.py
```diff
@@ -28,6 +29,91 @@
 from mava.core_jax import SystemTrainer
 
 
+class DefaultValueLossFunction(Component):
```
arnupretorius, line 32: Since this is setting up a SE loss, I think we should rename this to `SquaredErrorLoss`? We can also drop the "Function" in the class names to make them shorter.

mava/components/training/losses.py
```diff
@@ -28,6 +29,91 @@
...
+            training.BaseTrainerInit
+        ]  # import from training to avoid partial dependency
+
+
+@dataclass
+class HuberValueLossFunctionConfig:
+    huber_delta: float = 1.0
+
+
+class HuberValueLossFunction(DefaultValueLossFunction):
```
arnupretorius, line 86: Let's drop "Function" in class name. Also, not sure why we need to inherit from the default/SE loss? I think what we should do is create an abstract loss, as done below with the `Loss` class, but then be more specific with the difference between these classes. Might be good to have a quick offline chat about this? 🙂

arnupretorius (changes requested):
Thanks for this! 👍 Made a few suggestions 🙂

mava/components/training/losses.py
KaleabTessera: +1

tests/components/training/losses_test.py
```diff
@@ -275,3 +288,99 @@ def test_mapg_loss(
...
     assert loss_critic == 4.5
     assert low_loss_critic < loss_critic
+
+
+def test_mapg_huber_loss(
+    mock_trainer: Trainer,
+    mapg_loss: MAPGWithTrustRegionClippingLoss,  # noqa: E501
+) -> None:
+    """Test whether mapg huber loss output is as expected"""
+    huber_loss = HuberValueLossFunction()
```
KaleabTessera, line 298: Can we set `huber_delta` in the test? Incase the default ever changes (we still want this test to pass).

mava/components/training/losses.py
OmaymaMahjoub: I think this could be a great option too, @DriesSmit  what do you think?

tests/components/training/losses_test.py
OmaymaMahjoub: Yeah sure!

KaleabTessera (commented):
Thanks @siddarthsingh1 @DriesSmit and @OmaymaMahjoub ! This is very close to being ready from my perspective!

tests/components/training/losses_test.py
OmaymaMahjoub: I made a test to check that it update the value in the config: https://github.com/instadeepai/Mava/blob/451b5f896b07610eee46c7487dd743d5c81f25b0/tests/components/training/losses_test.py#L175

mava/components/training/losses.py
DriesSmit: Sounds good to me :slightly_smiling_face:

mava/components/component_groups.py
DriesSmit: Agreed. I removed the IPPO text. I added the Huber component to a component group to standardise our component setup. But I don't have a strong opinion on whether it should be there or not. I'm happy to remove it if you think that is better :+1:

tests/components/training/losses_test.py
KaleabTessera: Thanks!

KaleabTessera (dismissed):

mava/components/component_groups.py
DriesSmit: I removed the component_group as we discussed :)

mava/components/training/losses.py
```diff
@@ -29,12 +29,28 @@
...
+    def name() -> str:
+        """_summary_
+
+        Returns:
+            _description_
+        """
+        return "value_loss"
+
+
+class SquaredErrorLoss(ValueLoss):
```
arnupretorius, line 48: Sorry, I know I suggest the original name change, and this is nitpicky, but should we change it to `SquaredErrorValueLoss` to be more consistent?

mava/components/component_groups.py
```diff
@@ -6,7 +6,7 @@
     executing.RecurrentExecutorObserve,
 ]
 
-# The components that a needed for IPPO with the Huber value loss.
+# The components that a needed for the Huber value loss.
 huber_value_loss_components = [
```
arnupretorius: Spoke to @DriesSmit about this offline: that we should remove single components from the groups file. This is just a reminder to do that before we merge.

arnupretorius (commented):
Last few comments 😉

mava/components/component_groups.py
arnupretorius: Ok this is done :)

mava/components/training/losses.py
arnupretorius: Or change `HuberValueLoss` to `HuberLoss`

OmaymaMahjoub: Okay I will change the Square error to `SquaredErrorValueLoss`

OmaymaMahjoub: just we need to wait for the benchmark to end

OmaymaMahjoub: Done!

KaleabTessera (approved):
Thanks everyone!

arnupretorius (approved):
Thanks! 🔥

PR #768, opened 2022-09-27
fix: fix the failing checks

OmaymaMahjoub:
## What?
Minor fix in the environment loop.
## Why?
The mypy checks are failing in all the PR because of this error:
```
mava/environment_loop.py:575: error: Unsupported operand types for - ("Array" and "int")
mava/environment_loop.py:575: note: Left operand is of type "Union[Any, Array]"
mava/environment_loop.py:578: error: Argument 1 to "int" has incompatible type "Union[Any, Array]"; expected "Union[str, bytes, SupportsInt, SupportsIndex, SupportsTrunc]"
```
## How?
Minor type edit in the count value in environment loop file

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,7 +126,7 @@ setup(
         "record_episode": record_episode_requirements,
         "sc2": smac_requirements,
         "envs": pettingzoo_requirements + open_spiel_requirements + smac_requirements,
-        "jax==0.3.17": jax_requirements,
+        "jax": jax_requirements,
     },
     classifiers=[
         "Development Status :: 3 - Alpha",
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ jax_requirements = [
     "chex",
     # TODO Unpin JAX, once new rlax and acme version is released.
     # To fix cannot import name 'tree_multimap' from 'jax.tree_util'.
-    "jax==0.3.17",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -40,7 +40,7 @@
     "chex",
     # TODO Unpin JAX, once new rlax and acme version is released.
     # To fix cannot import name 'tree_multimap' from 'jax.tree_util'.
-    "jax<=0.3.15",
+    "jax==0.3.17",
```
siddarthsingh1, line 43: Should the setup.py file use `<=` instead of `==` to reduce requirement version issues for local installations?

siddarthsingh1 (commented):
Thanks for looking at this @OmaymaMahjoub. I just have one small question about the installation versions.

setup.py
OmaymaMahjoub: Yeah you are right I will bring it back to <=0.3.15

arnupretorius (approved):
Thanks @OmaymaMahjoub 🙂

AsadJeewa (approved):
Thanks @OmaymaMahjoub and @RuanJohn

KaleabTessera (approved):
Thanks @RuanJohn @OmaymaMahjoub !

PR #771, opened 2022-09-29
fix: pin the version of jaxlib (checks are failing)

OmaymaMahjoub:
## What?
Pin the version of jaxlib
## Why?
Checks are failing because of this error
`jaxlib version 0.3.20 is newer than and incompatible with jax version 0.3.15. Please update your jax and/or jaxlib packages.`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ jax_requirements = [
     # TODO Unpin JAX, once new rlax and acme version is released.
     # To fix cannot import name 'tree_multimap' from 'jax.tree_util'.
     "jax<=0.3.15",
-    "jaxlib",
+    "jaxlib<=0.3.19",
     "dm-haiku",
     "flax",
     "optax",
--- a/setup.py
+++ b/setup.py
@@ -38,10 +38,8 @@ tf_requirements = [
 
 jax_requirements = [
     "chex",
-    # TODO Unpin JAX, once new rlax and acme version is released.
-    # To fix cannot import name 'tree_multimap' from 'jax.tree_util'.
-    "jax<=0.3.15",
-    "jaxlib<=0.3.19",
+    "jax~=0.3.20",
+    "jaxlib~=0.3.20",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -41,7 +41,7 @@
     # TODO Unpin JAX, once new rlax and acme version is released.
     # To fix cannot import name 'tree_multimap' from 'jax.tree_util'.
     "jax<=0.3.15",
-    "jaxlib",
+    "jaxlib<=0.3.19",
```
KaleabTessera, line 44: Thanks @OmaymaMahjoub !  We should pin jax and jaxlib to the same versions.

KaleabTessera (changes requested):

setup.py
OmaymaMahjoub: Okay! Should we keep jax to 0.3.15 or change it to 0.3.17 since it works fine in that version too?

KaleabTessera: I updated it :) . Please let me know if you are happy?

OmaymaMahjoub: Thank you! I think the latest jax is 0.3.18 and it didn't work in our case bringing some errors so probably we can just stay with 0.3.17 so let's wait to see if the checks are fine else we can re edit it

OmaymaMahjoub: It's working fine 🙂 thank you 🙌

KaleabTessera (approved):
Thanks @OmaymaMahjoub ! 

@RuanJohn Can we please run an exp or two to sanity check upgrading jax doesn't have any weird side effects? :)

RuanJohn:
> Thanks @OmaymaMahjoub !
> 
> @RuanJohn Can we please run an exp or two to sanity check upgrading jax doesn't have any weird side effects? :)

On it :+1:

RuanJohn (approved):
Looks good and the benchmarks show no noticeable performance decreases.

PR #772, opened 2022-10-03
fix: Python 3.7 checker issue.

DriesSmit:
## What?
Set importlib packange version < 5.0.
## Why?
The latest version of importlib breaks the Python 3.7 checker.
## How?
Set importlib package version to be less than 5.0 in setup.py.


OmaymaMahjoub (approved):
Thanks @DriesSmit!

RuanJohn (approved):
Thank you @DriesSmit

PR #773, opened 2022-10-03
fix: Environment loop

### LINKED ISSUE #708: [FEATURE] Implement evaluation intervals

OmaymaMahjoub:
## What?
Remove the if statments that are related to tensorflow system
## Why?
Mava will only support jax systems
## How?
Edit `mava/environment_loop.py`
## Extra
Should be merged after the Chore: Remove TF get merged
Close #708


mava/environment_loop.py
```diff
@@ -631,14 +614,12 @@ def should_run_loop(eval_interval_condition: Tuple) -> bool:
...
                         results.update(self._environment.get_interval_stats())
-                    self._logger.write(results)
-
                 else:
                     result = self.run_episode()
                     episode_count += 1
                     step_count += result["episode_length"]
                     # Log the given results.
-                    self._logger.write(result)
+                self._logger.write(result)
```
DriesSmit, line 622: Will this not break in the case where `environment_loop_schedule` is not empty? I see that in that case the variable is called results. Maybe we should use the same variable naming?

DriesSmit (commented):
Thanks @OmaymaMahjoub :slightly_smiling_face:

mava/environment_loop.py
```diff
@@ -631,14 +614,12 @@ def should_run_loop(eval_interval_condition: Tuple) -> bool:
                     # Check for extra logs
                     if hasattr(self._environment, "get_interval_stats"):
                         results.update(self._environment.get_interval_stats())
-                    self._logger.write(results)
```
RuanJohn, line 634: Please keep this at this level of indentation.

mava/environment_loop.py
RuanJohn, line 622: Similarly here, please keep it at the same indentation level. This is so that we can fall back on the continual logging default when not evaluation interval details have been passed in :)

DriesSmit: Should we also move this outside the if else?
```
results = self.run_episode()
episode_count += 1
```

mava/environment_loop.py
```diff
@@ -585,36 +580,24 @@ def should_run_loop(eval_interval_condition: Tuple) -> bool:
 
         episode_count, step_count = 0, 0
```
RuanJohn, line 581: We can also try removing these now.

RuanJohn: And then we can make the while loop be a `while True` loop since we don't handle termination at the environment loop level anymore but in the termination component.

mava/environment_loop.py
OmaymaMahjoub: since I found this command `self._logger.write(result)` will be called in both if statement I thought it is fine to keep it outside the condition

OmaymaMahjoub: Agree I think it works if we move them out of the if statement

mava/environment_loop.py
OmaymaMahjoub: isn't working fine if we leave it out the if statement since it is called in line 641 too?

mava/environment_loop.py
DriesSmit: Yes I agree. The only issue is the naming. See results vs result.

OmaymaMahjoub: same as above

DriesSmit: Thanks @OmaymaMahjoub :slightly_smiling_face:

mava/environment_loop.py
OmaymaMahjoub: oh I see, I will check the single process first and we can confirm from there!

mava/environment_loop.py
OmaymaMahjoub: Oh now it makes sense hh

mava/environment_loop.py
OmaymaMahjoub: Oh I see I saw the difference

mava/__init__.py
```diff
@@ -13,7 +13,7 @@
 from mava.core import Executor, Trainer
 
 # Expose the environment loop.
-from mava.environment_loop import ParallelEnvironmentLoop, SequentialEnvironmentLoop
+from mava.environment_loop import ParallelEnvironmentLoop
```
arnupretorius, line 16: Are we still going to support Sequential env loops in the future? If not, or at least if not in the near to medium term future, I guess we can simply call this class `EnvironmentLoop` and drop the "Parallel"?

arnupretorius (commented):
Thanks @OmaymaMahjoub 😉

mava/__init__.py
OmaymaMahjoub: I will check with @DriesSmit, if we won't support it I will change the name!

arnupretorius: Given our discussion offline, let's keep it as is for now 👍

tests/jax/mocks.py
```diff
@@ -151,37 +149,20 @@ def make_fake_env(
...
-            )
-        elif env_type == EnvType.Sequential:
-            env = SequentialMADiscreteEnvironment(
-                num_actions=18,
-                num_observations=2,
-                obs_shape=(84, 84, 4),
-                obs_dtype=np.float32,
-                episode_length=10,
-            )
+        env = ParallelMADiscreteEnvironment(
```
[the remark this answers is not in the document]
DriesSmit: Good idea. I removed the MA.

KaleabTessera (changes requested):
Thanks for amazing work @OmaymaMahjoub :rocket: 

Just some extra things we can remove in the env loop:
- We can simplify `record_counts()`. We were trying to handle various kinds of systems back then. 
- We can remove `if hasattr(self._executor, "after_action_selection"):` and the whole condition inside completely (this was part of tf systems).

tests/jax/mocks.py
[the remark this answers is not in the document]
DriesSmit: I added the MA again. It seems that MA makes sense as we are adapting the normal ParallelDiscreteEnvironment to the multi-agent setting. Happy to have a chat about this.

DriesSmit:
Close this PR as it has an cla issue in it.

PR #781, opened 2022-10-06
Feature: Orthogonal Networks Initialisation. 

### LINKED ISSUE #731: [FEATURE] Orthogonal initialisation
Implement orthogonal initialisation for the JAX IPPO system.

### LINKED ISSUE #780: [INVESTIGATION] Effect of activating final layer of MLP in Mava networks.
### What do you want to investigate?
Should the final layer of the intermediate MLP in the Mava networks be activated? 
Conceptually, it makes sense that this shoudl be done since the final layer of the MLP is still a hidden layer in the full network. The reason for this that it is followed by a `CategoricalValueHead` in the policy and a `ValueHead` in the critic. 

### Definition of done
Benchmarks will case a for or against.

RuanJohn:
## What?
Change the way policy and critic network weights are initialised. 
## Why?
Following from recommendations made in the surprising effectiveness of PPO paper.  
## How?
Add flag to networks for orthogonal initialisation. 
## Extra
The paper is not specific about what scale what values to use for the intialisation of the weights. So I used the values as given in the [ICLR blogpost](https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/). 
[Benchmarking results. ](https://www.notion.so/instadeep/Orthogonal-Network-Initialisation-82b8f24b0c294b9dbaf29ae08d6dcb58)
close #731
close #780


mava/systems/ippo/networks.py
```diff
@@ -187,6 +188,8 @@ def make_discrete_networks(
     policy_layer_sizes: Sequence[int],
     critic_layer_sizes: Sequence[int],
     observation_network: Callable = utils.batch_concat,
+    activate_mlp_final: bool = False,
```
DriesSmit, line 191: Should we not always set it to True? Do we need this flag?

mava/systems/ippo/networks.py
```diff
@@ -259,6 +313,8 @@ def make_networks(
     ),
     critic_layer_sizes: Sequence[int] = (512, 512, 256),
     observation_network: Callable = utils.batch_concat,
+    activate_mlp_final: bool = False,
```
DriesSmit, line 316: See above comment.

mava/systems/ippo/networks.py
```diff
@@ -198,34 +201,85 @@ def make_discrete_networks(
...
+                        policy_layer_sizes,
+                        activation=jax.nn.relu,
+                        activate_final=activate_mlp_final,
+                        w_init=hk.initializers.Orthogonal(scale=jnp.sqrt(2)),
+                        b_init=hk.initializers.Constant(constant=0.0),
+                    ),
+                    networks_lib.CategoricalHead(
+                        num_values=num_actions,
+                        dtype=environment_spec.actions.dtype,
+                        w_init=hk.initializers.Orthogonal(scale=0.01),
```
DriesSmit, line 230: Does the heads don't need a `b_init`? The code is probably correct. I'm just checking :)

mava/systems/ippo/networks.py
```diff
@@ -198,34 +201,85 @@ def make_discrete_networks(
...
+
+        def policy_fn(inputs: jnp.ndarray) -> networks_lib.FeedForwardNetwork:
+            policy_network = hk.Sequential(
+                [
+                    observation_network,
+                    hk.nets.MLP(
+                        policy_layer_sizes,
+                        activation=jax.nn.relu,
+                        activate_final=activate_mlp_final,
+                        w_init=hk.initializers.Orthogonal(scale=jnp.sqrt(2)),
```
DriesSmit, line 224: We can rather create two lambda functions depending on `if orthogonal_initialisation`. And then just call them inside the policy_fn and critic_fn. Instead of duplicating the policy and critic code.

DriesSmit (changes requested):
Thanks @RuanJohn :slightly_smiling_face:  This is a really nice catch :raised_hands: :fire:  Just see my few comments.

mava/systems/ippo/networks.py
RuanJohn: Totally agree. I just made a flag for now to be able to benchmark this again. I'll remove this flag :)

mava/systems/ippo/networks.py
RuanJohn: This is a good question. From reading the blog post it seems that the heads only get weight initialisation. Also acme's `CategoricalHead` cannot be initialised with a `b-init` parameter which led me to believe it might not be necessary. Please see [here](https://github.com/deepmind/acme/blob/49c8e1b591490d8d36dba21d84b3453cfdb92132/acme/jax/networks/distributional.py#L33).

mava/systems/ippo/networks.py
RuanJohn: I like this a lot more! Will update :)

mava/systems/ippo/networks.py
RuanJohn: [These benchmarks](https://www.notion.so/instadeep/Experiments-52346a5de10d444583896c8d5b026b25?p=3f347755c06147d1922c4cbd235fca19&pm=s) show conclusively that we should activate these intermediate layers so I'll set it to be the default for moving forward and remove the flag.

mava/systems/ippo/networks.py
RuanJohn: Done. Please let me know your thoughts. :)

mava/systems/ippo/networks.py
```diff
@@ -170,10 +170,11 @@ class ValueHead(hk.Module):
     def __init__(
         self,
         name: Optional[str] = None,
+        w_init: Optional[hk.initializers.Initializer] = None,
```
OmaymaMahjoub, line 173: If you can add a docstring to define the attributes of the class:
w_init: Optional initializer for weights.
Also why only w_init can't we extend to have b_init too? or it is uncessary?

mava/systems/ippo/networks.py
```diff
@@ -187,6 +188,8 @@ def make_discrete_networks(
     policy_layer_sizes: Sequence[int],
     critic_layer_sizes: Sequence[int],
     observation_network: Callable = utils.batch_concat,
+    orthogonal_initialisation: bool = False,
+    activation_function: Callable[[jnp.ndarray], jnp.ndarray] = jax.nn.relu,
```
OmaymaMahjoub, line 192: Probably it will be better to mention that these attributes are Optional? 
`orthogonal_initialisation: Optional bool = False,` ...

OmaymaMahjoub (changes requested):
Thank you @RuanJohn for this PR!
If you can also make an example file using the orthogonal network in the example repo .

mava/systems/ippo/networks.py
RuanJohn: Updated the docstring :) 
Please see my comment [here](https://github.com/instadeepai/Mava/pull/781#discussion_r989696451) about only having `w_init` in the `CategoricalValueHead` and `ValueHead` classes.

mava/systems/ippo/networks.py
RuanJohn: At the moment `orthogonal_initialisation` is not an optional parameter but it rather getting set to `False` by default. So I am not sure whether adding `Optional` to the type hint will work?

mava/systems/ippo/networks.py
OmaymaMahjoub: I see thanks  for the explanation!

mava/systems/ippo/networks.py
```diff
@@ -174,10 +176,19 @@ class ValueHead(hk.Module):
...
+        w_init: Optional[hk.initializers.Initializer] = None,
     ):
-        """Initialize the class"""
+        """Initialize the value head.
+
+        Args:
+            name: An optional string name for the class. Must be a valid
+                Python identifier. If ``name`` is not provided then the class
+                name for the current instance is converted to
+                ``lower_snake_case`` and used instead.
```
arnupretorius, line 187: Lower_snake_case? 🤔

mava/systems/ippo/networks.py
```diff
@@ -203,9 +216,16 @@ def make_discrete_networks(
...
         policy_recurrent_layer_sizes: Optionally add recurrent layers to the policy
-        recurrent_architecture_fn: Architecture to use for the recurrent units, e.g. LSTM or GRU.
-        policy_layers_after_recurrent: sizes of hidden layers for the policy network after the recurrent layers.
-        This is only used if an recurrent architecture is used.
+        recurrent_architecture_fn: Architecture to use for the recurrent units,
+        e.g. LSTM or GRU.
+        policy_layers_after_recurrent: sizes of hidden layers for the
+            policy network after the recurrent layers. This is only used if a
+            recurrent architecture is used.
+        orthogonal_initialisation: Whether network weights should be
```
arnupretorius, line 224: Would it perhaps be nicer to allow any initialiser to be passed in, or a dict of initialisers with keys such as "w_init", "b_init" etc? I'm not sure we need this flexibility, but just thought I would ask.

mava/systems/ippo/networks.py
```diff
@@ -225,7 +245,21 @@ def policy_fn(inputs: jnp.ndarray) -> networks_lib.FeedForwardNetwork:
         """
         # Add the observation network and an MLP network.
         policy_network = [
-            hk.nets.MLP(policy_layer_sizes, activation=jax.nn.relu),
+            hk.nets.MLP(
+                policy_layer_sizes,
+                activation=activation_function,
+                w_init=(
+                    lambda x: hk.initializers.Orthogonal(scale=jnp.sqrt(2))
```
arnupretorius, line 252: Do you need the lambda function here? Can you not use something like `w_init=hk.initializers.Orthogonal(scale=jnp.sqrt(2)) if orthogonal_initialisation else None`? Similar Q for the other cases. Also, if you can't as an argument, you can always assign the initialiser outside the network and use it as an argument later when defining the MLP.

arnupretorius (commented):
Thanks @RuanJohn! 🙂  Left a few comments.

mava/systems/ippo/networks.py
arnupretorius: Or have lambda functions, but then reuse them where appropriate as @DriesSmit suggested.

mava/systems/ippo/networks.py
RuanJohn: I think it is something we could perhaps consider in the future? Since our current need is only to have orthogonal initialisation or not. Perhaps when we see a new for many different initialisers or want to experiment with it we can implement this. What are your thoughts?

mava/systems/ippo/networks.py
RuanJohn: I got this from Haiku's docstrings :sweat_smile: 
I think it refers to converting `SomeClassName` to `some_class_name` when no name parameter is given when the class is initialised. Happy to remove this or change it. :)

mava/systems/ippo/networks.py
RuanJohn: I created reusable `w_init_fn` and `b_init_fn` lambda functions. Please let me know your thoughts on this? :)

mava/systems/ippo/networks.py
arnupretorius: Yeah, makes sense, let's leave it for a future PR then, if we find we need the flexibility. 👍

mava/systems/ippo/networks.py
RuanJohn: Totally agree with this. I've updated the docstring accordingly :)

RuanJohn:
> Thank you @RuanJohn for this PR! If you can also make an example file using the orthogonal network in the example repo .

Good catch! Added one :)

arnupretorius (dismissed):
Thanks @RuanJohn 🙂

mava/systems/ippo/networks.py
KaleabTessera: I think we can maybe remove the bias init since the default in haiku for the bias is 0 anyway - [link](https://github.com/deepmind/dm-haiku/blob/307cf7dbda64d637ca423cacc9978f0ca19dc8a6/haiku/_src/basic.py#L156).

arnupretorius:
> Thanks @RuanJohn 🙂

I see there is a check failing regarding code coverage but happy with all the changes 👍 Is this still being benchmarked?

mava/systems/ippo/networks.py
```diff
@@ -225,7 +255,13 @@ def policy_fn(inputs: jnp.ndarray) -> networks_lib.FeedForwardNetwork:
         """
         # Add the observation network and an MLP network.
         policy_network = [
-            hk.nets.MLP(policy_layer_sizes, activation=jax.nn.relu),
+            hk.nets.MLP(
+                policy_layer_sizes,
+                activation=activation_function,
+                w_init=w_init_fn(orthogonal_initialisation, jnp.sqrt(2)),
+                b_init=b_init_fn(orthogonal_initialisation, 0.0),
```
KaleabTessera, line 262: I think we can maybe remove the bias init since the default in haiku for the bias is 0 anyway - [link](https://github.com/deepmind/dm-haiku/blob/307cf7dbda64d637ca423cacc9978f0ca19dc8a6/haiku/_src/basic.py#L156).

KaleabTessera (changes requested):
Thanks @RuanJohn , nice implementation. In the future, it might make sense to allow people to pass in their own init functions, but for now this is perfect. 

Just a minor comment about this bias. Also, should the recurrent layer (https://github.com/instadeepai/Mava/blob/465a1eb2fb592b6df579413318e225ca3a6b5ca9/mava/systems/ippo/networks.py#L273), have this?

RuanJohn:
> Thanks @RuanJohn , nice implementation. In the future, it might make sense to allow people to pass in their own init functions, but for now this is perfect.
> 
> Just a minor comment about this bias. Also, should the recurrent layer (
> 
> https://github.com/instadeepai/Mava/blob/465a1eb2fb592b6df579413318e225ca3a6b5ca9/mava/systems/ippo/networks.py#L273
> 
> ), have this?

Yes, it is quite interesting how the paper handles recurrent networks. The have two feedforward layers, followed by a recurrent layer followed by a feedforward layer. Please see the final paragraph in Appendix D.2 and Tables 7 & 8 [here](https://arxiv.org/pdf/2103.01955v2.pdf).

mava/systems/ippo/networks.py
DriesSmit: Great :partying_face:

mava/systems/ippo/networks.py
RuanJohn: Good catch! Removed it :)

mava/systems/ippo/networks.py
```diff
@@ -236,14 +266,17 @@ def policy_fn(inputs: jnp.ndarray) -> networks_lib.FeedForwardNetwork:
...
                 activate_final=True,
             ),
 
         # Add a categorical value head.
         policy_network.append(
             networks_lib.CategoricalHead(
-                num_values=num_actions, dtype=environment_spec.actions.dtype
+                num_values=num_actions,
+                dtype=environment_spec.actions.dtype,
+                w_init=w_init_fn(orthogonal_initialisation, 0.01),
```
ulricharmel, line 279: @RuanJohn  The network layer gain is exactly what you implemented here. I checked their [implementation](https://github.com/marlbenchmark/on-policy/blob/main/onpolicy/algorithms/utils/act.py#L32) and this is exactly what they are doing. So we already have this covered. What I may suggest is to make the gain value a parameter so that we can play around with the value. In the paper they have experiments with values of 0.01 and 1. We can link this [issue](https://github.com/instadeepai/Mava/issues/789) to this PR. Also it will be great if we had a wrapper for the initialisation so that we can easily try different ones. Not just have orthogonal_initialisation as the only choice along with the default.

ulricharmel (changes requested):
Let make the gain value for the weight initialisation of the last actor layer a parameter. Not hard code it directly to 0.01.

mava/systems/ippo/networks.py
RuanJohn: Thank you for looking into this @ulricharmel. I have added the policy gain as a parameter to the networks. I agree that we can increase the flexibility in the future but I think we can do this as required or expand functionality after the release since we already have an idea of what scaling values to use. Please let me know your thoughts?

ulricharmel (approved):
Thanks @RuanJohn

arnupretorius (approved):
Thanks @RuanJohn 🙂

KaleabTessera (approved):
Thanks @RuanJohn !

DriesSmit (approved):

OmaymaMahjoub (approved):
Thanks @RuanJohn!

PR #786, opened 2022-10-07
update benchmarking labels

arnupretorius:
## What?
Update to ci for docker to reflect new benchmarking labels. Added new labels to be more specific about benchmarking status:

* Benchmark required - needs benchmarking, but has not yet been set to run.
* Benchmark in progress - experiments have been kicked off and are still running.
* Benchmarking is complete - results are in.

### DIFF
--- a/.github/workflows/push_docker.yaml
+++ b/.github/workflows/push_docker.yaml
@@ -103,7 +103,7 @@ jobs:
 
   # Action pushes new docker image if we have a PR to dev and have the label benchmark
   build-push-branches:
-    if:  ${{ github.event.label.name == 'benchmark' }}
+    if:  ${{ github.event.label.name == 'benchmark required' }}
     runs-on: ubuntu-latest
     steps:
       -


AsadJeewa (approved):
Thanks, @arnupretorius 
This will make reviewing a lot more streamlined :smile:

.github/workflows/push_docker.yaml
```diff
@@ -103,7 +103,7 @@ jobs:
 
   # Action pushes new docker image if we have a PR to dev and have the label benchmark
   build-push-branches:
-    if:  ${{ github.event.label.name == 'benchmark' }}
+    if:  ${{ github.event.label.name == 'benchmark required' }}
```
AsadJeewa, line 106: One minor thing is that we could make it "benchmark complete" with a small 'b' so it is in line with other labels. Same for the maintenance label

arnupretorius: Nice catch, will do 👍

RuanJohn (approved):
Thank you @arnupretorius :)

PR #787, opened 2022-10-09
Feature/observations value normalisation

### LINKED ISSUE #722: [FEATURE] Add option to normalise critic values during training

### LINKED ISSUE #723: [FEATURE] Add option to normalise features during training
As suggested in this [paper](https://arxiv.org/pdf/2103.01955.pdf), normalising the observations and state inputs can improve results in certain MARL environments. This is similar to feature to that suggested in this [issue](https://github.com/instadeepai/Mava/issues/722).

ulricharmel:
## What?

- Add option to normalise observations during training.
- Also modified the code for the target value normalisation.

## Why?

- Normalising target values as stated in the Yu et al (2021) paper can significantly improve training for certain environments.
- Observation normalisation is one of the recommendations from the [PPO implementation details blog post](https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details).

## How?

- Added observation normalisation parameters to the trainer parameter client.
- Update  observation normalisation parameters using running stats during training.
- Normalise observations before agents select actions.

## Extra

- [Benchmarking results](https://www.notion.so/instadeep/Observation-normalisation-7814ceebafe1446f84c1d60c1490b5f5)
- Removed the need to denormalise target values in the GAE function.
- flake 8 complains MAPGWithTrustRegionStep.on_training_step_fn' is too complex 
- Closes #723 
- Closes #722


mava/utils/jax_training_utils.py
```diff
@@ -47,41 +50,84 @@ def compute_running_mean_var_count(
...
+    Returns:
+        denormalize batch (Dictionary)
+    """
+
+    # The type casting is required because we need to preserve
+    # the data type before the policy info is computed else we will get
+    # an error from the table about the type of policy being double instead of float.
+    dtype = observation.observation.dtype
+    obs = observation.observation
+    norm_obs = np.array(normalize(stats, obs), dtype=dtype)
```
DriesSmit, line 126: Is it the stats that are not of the correct dtype? Could we not then maybe just change the stats dtype the be the same as the dtype of the obs when we init it?

DriesSmit (commented):
Thanks @ulricharmel :slightly_smiling_face: This is a really nice implementation :partying_face: Just see my few comment.

mava/components/training/step.py
```diff
@@ -478,8 +495,27 @@ def step(sample: reverb.ReplaySample) -> Tuple[Dict[str, jnp.ndarray]]:
                     constants.OPT_STATE_DICT_KEY
                 ] = new_states.critic_opt_states[net_key][constants.OPT_STATE_DICT_KEY]
 
-            # update the running stats
-            trainer.store.stats = new_states.stats
+            # Update the observation normalization parameters
+            for agent in trainer.store.trainer_agent_net_keys.keys():
+                obs_shape = len(
+                    trainer.store.obs_norm_params[constants.OBS_NORM_STATE_DICT_KEY][
```
DriesSmit, line 501: Maybe create a variable for  `trainer.store.obs_norm_params[constants.OBS_NORM_STATE_DICT_KEY][agent]` to make it easier to read.

mava/components/training/model_updating.py
```diff
@@ -95,17 +99,31 @@ def on_training_utility_fns(self, trainer: SystemTrainer) -> None:
...
+            )
+
+        # Initilaise observations running statisitics here
+        # We only need to modify the variance since this is actually intiailised
+        # in the training/trainer on_building_init_end callback.
+        if self.config.normalize_observations:
+            trainer.store.norm_obs_running_stats_fn = update_and_normalize_observations
+            for agent in trainer.store.trainer_agent_net_keys.keys():
+                obs_shape = len(
+                    trainer.store.obs_norm_params[constants.OBS_NORM_STATE_DICT_KEY][
```
DriesSmit, line 119: Maybe just create a variable as mentioned in my other comment. If there is any other place where this is also applicable please update the code.

mava/components/training/model_updating.py
```diff
@@ -95,17 +99,31 @@ def on_training_utility_fns(self, trainer: SystemTrainer) -> None:
...
+                        agent
+                    ]["var"]
+                )
+                for x in range(obs_shape):
+                    trainer.store.obs_norm_params[constants.OBS_NORM_STATE_DICT_KEY][
+                        agent
+                    ]["var"][x] = 0
```
DriesSmit, line 126: Just to sanity check. Why is the variance set to zero initially? Should this not be done in the builder?

DriesSmit:
Can you also just make sure that the normalisation variables for the observation and value function are getting loaded from a checkpoint when there is a checkpoint available.

mava/components/training/model_updating.py
ulricharmel: So in the builder we set it to 1 so that initial observations are not normalised. In the trainer we set them to 0 and these are updated correctly during the training step. Eventually when the parameters are updated by the trainer, the executors will fetch the correct values. To have a good approximation the initial value should be 0. But we can not normalise with a value of 0, so initially we just keep the variance in the executor to be 1. This is the same approach I saw in acme implementation.

mava/utils/jax_training_utils.py
ulricharmel: Thanks @DriesSmit casting the stats fixed the issue.

mava/components/training/step.py
ulricharmel: Just made a PR to make this easier to read. I made a variable for constants.OBS_NORM_STATE_DICT_KEY.

mava/utils/jax_training_utils.py
DriesSmit: Great, thanks :slightly_smiling_face:

mava/components/training/step.py
DriesSmit: Thanks :slightly_smiling_face: I was wondering if you could not make the variable for the entire `trainer.store.obs_norm_params[constants.OBS_NORM_STATE_DICT_KEY][agent]` instead, inside the for loop. But its up to you.

mava/components/training/model_updating.py
DriesSmit: Ah okay, that sounds good.  One thing that I'm still not sure about is what happens when a checkpoint is loaded. The `on_training_utility_fns` gets executed after the parameter client gets the variable values from the checkpointer. Would this not mean that if we load from a checkpoint we would overwrite the variance with zero? Is this intended behaviour?

ulricharmel: No the values in the executors only change when we call [executor update](https://github.com/instadeepai/Mava/blob/develop/mava/environment_loop.py#L164). I believe this can only happen after we have done atleast a training step so at that time the parameters should be correct. I get the correct behaviour for the test I wrote in a single process.

ulricharmel:
@DriesSmit I add a new parameter to the stats. This solve the issue. No need to change the variance from 0 to 1 any more. So this should solve any potential check-pointer issue.

mava/components/training/model_updating.py
DriesSmit: Yes, but the problem is when restore from a checkpoint. In that case we don't want zero variance right? But this function is called after we load from the checkpointer and therefore it resets the variance to zero. Or does your single process test check the checkpointer as well?

ulricharmel: Check the new changes. We don't need this anymore.

DriesSmit: I just saw your latest message. So everything is loaded from the builder now? That should solve the problem.

ulricharmel: Yes, the solution was to add an std parameter to the stats. Initialise it to 0. Keep the variance to 0 and when the variance is updated, update the std as the sqrt of the variance.

DriesSmit: Great. Thanks @ulricharmel :slightly_smiling_face:

DriesSmit (approved):
Thanks @ulricharmel :fire:

mava/components/updating/parameter_server.py
```diff
@@ -132,6 +132,9 @@ def on_parameter_server_init_start(self, server: SystemParameterServer) -> None:
                 f"critic_opt_state-{agent_net_key}"
             ] = server.store.critic_opt_states[agent_net_key]
 
+        # Normalization parameters
+        server.store.parameters["norm_params"] = server.store.norm_params
```
sash-a, line 136: Maybe add this to constants?

mava/utils/jax_training_utils.py
```diff
@@ -46,42 +49,88 @@ def compute_running_mean_var_count(
     M2 = m_a + m_b + jnp.square(delta) * count * batch_count / tot_count
     new_var = M2 / tot_count
     new_count = tot_count
+    new_std = jnp.sqrt(new_var)
 
-    return jnp.array([new_mean, new_var, new_count])
+    return dict(mean=new_mean, var=new_var, std=new_std, count=new_count)
```
sash-a, line 54: Is there a reason you changed this to a dict? In general jax arrays will keep more of the code jittable, although sometimes less readable. Don't feel like you have to change this because it'll probably be a big change, but in general I think keeping jax arrays as much as possible is a good idea

mava/utils/jax_training_utils.py
```diff
@@ -46,42 +49,88 @@ def compute_running_mean_var_count(
...
 
 
-def normalize(stats: jnp.ndarray, batch: jnp.ndarray) -> jnp.ndarray:
+def normalize(stats: Any, batch: jnp.ndarray) -> jnp.ndarray:
```
sash-a, line 57: change the typing to `Dict[str, float]` unless it can be anything else?

Although if this was a `jnp.array` we could jit this function :wink:

mava/utils/jax_training_utils.py
```diff
@@ -46,42 +49,88 @@ def compute_running_mean_var_count(
...
-    mean, var = stats[0], stats[1]
-    normalize_batch = (batch - mean) / (jnp.sqrt(jnp.clip(var, a_min=1e-2)))
+    mean, std = stats["mean"], stats["std"]
+    normalize_batch = (batch - mean) / (std + 1e-8)
 
     return normalize_batch
 
 
-def denormalize(stats: jnp.ndarray, batch: jnp.ndarray) -> jnp.ndarray:
+def denormalize(stats: Any, batch: jnp.ndarray) -> jnp.ndarray:
```
sash-a, line 73: Same comment as above

mava/utils/jax_training_utils.py
```diff
@@ -46,42 +49,88 @@ def compute_running_mean_var_count(
...
+    stats (Dictionary)   -- array with running mean, var, count.
+    batch (OLT namespace)   -- current batch of data for a single agent.
+
+    Returns:
+        normalize batch (Dictionary)
+    """
+
+    obs_shape = observation.observation.shape
+    obs = jax.tree_util.tree_map(
+        lambda x: jnp.reshape(x, [-1] + list(x.shape[2:])), observation.observation
```
sash-a, line 101: haiku has a `merge_leading_dims` function which would be a bit cleaner

`from haiku._src.basic import merge_leading_dims`

tests/components/building/parameter_client_test.py
```diff
@@ -82,7 +82,26 @@ def name() -> str:
     "critic_opt_state-network_agent_2",
 }
 
-expected_keys = expected_count_keys.union(expected_network_keys)
+normalisation_keys = {"norm_params"}
```
sash-a, line 85: again would be nice to replace this with a constant

mava/components/training/base.py
```diff
@@ -48,7 +48,8 @@ class TrainingState(NamedTuple):
     policy_opt_states: Dict[str, optax.OptState]
     critic_opt_states: Dict[str, optax.OptState]
     random_key: Any
-    stats: Any
+    target_stats: Any
+    observation_stats: Any
```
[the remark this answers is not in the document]
sash-a: Ok having read through all the code I would say rename this to `value_stats` or `target_value_stats` as it wasn't immediately obvious to me what this was

sash-a (changes requested):
Looks really good and logic all seems to be great! Just some minor code style comments :smile:

mava/components/training/base.py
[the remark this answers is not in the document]
ulricharmel: But what we normalise are the target_values (expected return from the GAE function). So we assume the output values from the neural network already normalised.

mava/utils/jax_training_utils.py
ulricharmel: Because I wanted to use the same functions for observation normalisation and target value normalisation. The observations are multidimensional so we can just stack them in array like for the target values where we only have single value values for the mean, var and ...

mava/utils/jax_training_utils.py
ulricharmel: So the stats will be arrays for observations and single numbers for target values.

mava/utils/jax_training_utils.py
sash-a: I think you could use a 3x1 array then? But again this isn't a big issue so I think you can leave it as is :smile:

mava/utils/jax_training_utils.py
sash-a: Ok then can you do `Dict[str, Union[jnp.array, float]]`?

mava/components/training/base.py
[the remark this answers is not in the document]
sash-a: Ok then could you change it to `target_value_stats` just a bit more clear :smile:

mava/utils/jax_training_utils.py
ulricharmel: Okay I will do that.

mava/components/training/step.py
```diff
@@ -240,6 +241,17 @@ def sgd_step(
                 data.extras,
             )
 
+            # Perform observation normalization if neccesary before proceeding
+            observation_stats = states.observation_stats
+            if hasattr(trainer.store, "norm_obs_running_stats_fn"):
```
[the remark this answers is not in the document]
arnupretorius: I think this might be a recurring thing, i.e. that we need to check if something is in the store or not. Perhaps, we can think about adding an easy method to the store class to check this, e.g. `store.has("something")`, and for this `hasattr` should work, but then at least it looks much cleaner in other parts of the code and is shorter than checking the config.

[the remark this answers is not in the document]
ulricharmel: I just adopted @sash-a approach. Here we actually really don't need since we can just use` if trainer.store.global_config.normalize_observations or if trainer.store.global_config.normalize_target_values`

mava/components/training/trainer.py
```diff
@@ -20,6 +20,8 @@
 from types import SimpleNamespace
 from typing import Any, Dict, List, Type
 
+import numpy as np
```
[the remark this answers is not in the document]
ulricharmel: adopted jax.numpy

mava/utils/jax_training_utils.py
ulricharmel: I will keep like this for now.

tests/components/building/parameter_client_test.py
ulricharmel: This is the main name for the normalisation parameters dict. Then we have two constant keys for this dictionary. 
```
obs_norm_key = constants.OBS_NORM_STATE_DICT_KEY
values_norm_key = constants.VALUES_NORM_STATE_DICT_KEY
```

sash-a (approved):
Looks good, thanks @ulricharmel! :fire:

DriesSmit (approved):
Nice work @ulricharmel  :slightly_smiling_face:

PR #792, opened 2022-10-11
feat: sum return feature

### LINKED ISSUE #791: [FEATURE] MPE sum episode return
### Feature
Add additional running statistics to the environment which will be the sum_episode_return.

### Proposal
Some papers use MPE with a global return equivalent to **the sum** instead of **the mean** of all the agents reward, therefore, to make our results comparable to the existing results we need to implement this feature.

### Definition of done
Implement the sum of episode return in `mava/wrappers/environment_loop_wrappers.py`

OmaymaMahjoub:
## What?
Add the option to have the sum of the agents' return as a global one
## Why?
To compare our results in MPE with prior works
## How?
add the flag `sum_episode_return` to the config that if it is true we will have sum_episode return
## Extra
Close #791


mava/wrappers/environment_loop_wrappers.py
```diff
@@ -98,6 +98,9 @@ def __init__(
             for stat in self._summary_stats:
                 self._running_statistics[f"{stat}_{metric}"] = 0.0
 
+        if self._sum_episode_return:
```
KaleabTessera, line 101: Can we log this by default, as opposed to a flag?

OmaymaMahjoub: Yeah sure

OmaymaMahjoub: Should we remove the flag or keep it with True?

KaleabTessera (changes requested):
This is amazing :fire: 

Could we log this by default and remove the flag and example? :)

mava/wrappers/environment_loop_wrappers.py
```diff
@@ -254,6 +255,7 @@ def _compute_episode_statistics(
         self._episode_length_stats.push(episode_steps)
         self._episode_return_stats.push(mean_episode_return)
         self._steps_per_second_stats.push(steps_per_second)
+        self._running_statistics["sum_episode_return"] = sum_episode_return
```
RuanJohn, line 258: Should we potentially treat this in the same way we are current treating the episode return which is the mean over all agent returns? That way we would have `Raw Sum Episode Return`, `Mean Sum Episode Return` etc logged. If we don't want to do this, can we then rename this metric to `raw_sum_episode_return` to be consistent with the naming of the other metrics in the logs? 
Keen to hear your thoughts on this as well @KaleabTessera?

KaleabTessera: This are great points @RuanJohn!

I am honestly worried about the confusion between these params e.g. Mean Sum Episode Return might be confusing. I think `raw_sum_episode_return` is a good compromise. We should update the logging docs to reflect this.

OmaymaMahjoub: Okay I will just change the name to `raw_sum_episode_return`

mava/wrappers/environment_loop_wrappers.py
```diff
@@ -229,7 +230,7 @@ def _compute_episode_statistics(
         # Collect the results and combine with counts.
         steps_per_second = episode_steps / (time.time() - start_time)
         mean_episode_return = np.mean(np.array(list(episode_returns.values())))
-
+        raw_sum_episode_return = np.sum(np.array(list(episode_returns.values())))
```
arnupretorius, line 233: Would `sum(episode_returns.values())` work here? Or do you need the type to be a numpy float?

arnupretorius: Same with the mean calculation.

arnupretorius (commented):
Looks good, thanks! 🙂 Just had one comment.

mava/wrappers/environment_loop_wrappers.py
OmaymaMahjoub: I will give it a try and back to you soon :)

OmaymaMahjoub: for the sum it works without numpy but for the mean we can make the equation instead of numpy if you find it really necessary to do so?
Should we remove np and replace it with sum and mean equation?

arnupretorius: This is minor, so whichever way you think is best :)

OmaymaMahjoub: I think we need to keep `np.sum` because it is more coeherent with `np.mean`, but is up to you!

mava/wrappers/environment_loop_wrappers.py
RuanJohn: I agree with that. Thank you @OmaymaMahjoub and @KaleabTessera. I don't think we have explicit logger docs anymore after the readme got trimmed down. But definitely worth introducing in the docs again at some point.

arnupretorius (approved):
Thanks @OmaymaMahjoub! 🙂

KaleabTessera (approved):

PR #793, opened 2022-10-11
Feature: Observations stacking

### LINKED ISSUE #788: [FEATURE] Observation concatenation.
When using feedforward policies, partial observability can be handled by using multiple observations instead of only the current one. This is achieved by concatenating the previous `n-1` observations to observations to the current observation.  

An additional environment wrapper that keeps track if the previous `n-1` environment observations and is able to concatenate then to be used as the input for a feedforward policy. For example, if `n=1`, then only the current observation is used as input, and if `n=2`, then the current and previous observations are concatenated to form the input. 

Checklist:
- Wrapper is implemented and working 
- Tests are written for the wrapper 
- The effect on system performance is benchmarked and reported.

ulricharmel:
## What?

- Add an environment wrapper that stacks new observation to past observations for training.

## Why?

- This is one of the suggestion in  Yu et al (2021) for Atari environments. Remembering past observations can help the agent choose a better policy.

## How?

-  Added a new wrapper class that uses a deque list to store past observations. 
-  During reset, we just repeat the same observation for the required number of frames.
-  During a step, we delete the oldest obs (add the top of the queue) and add the most recent one to the end of the queue.

[Benchmarks](https://www.notion.so/instadeep/Observation-stacking-test-c876ea2c8690451cb132ac88acb03052)

## Extra

- Closes #788


mava/utils/environments/pettingzoo_utils.py
```diff
@@ -130,7 +138,12 @@ def make_environment(
     else:
         raise Exception("Pettingzoo is not installed.")
 
+    # This should not be used with petting zoo environments since
+    # stacking is already done using supersuit.
+    if stack_frames > 1:
```
KaleabTessera, line 143: Are we going to implement/import observation stacking from Supersuit in another PR? If so, maybe we should just throw a `NotImplementedException` here for now.

mava/wrappers/env_preprocess_wrappers.py
```diff
@@ -213,3 +218,137 @@ def __getattr__(self, name: str) -> Any:
...
+            agent: deque([], maxlen=num_frames)
+            for agent in self._environment.possible_agents
+        }
+        self.num_frames = num_frames
+
+    def reset(self) -> dm_env.TimeStep:
+        """Reset the environment by setting the same observation to all the frames"""
+
+        timestep = self._environment.reset()
+        if type(timestep) == tuple:
```
KaleabTessera, line 240: I think we should spend some time trying to design simple, composable wrappers similar to how supersuit does this. They have a general wrapper class and each subclass modifies it wants , e.g. [here](https://github.com/Farama-Foundation/SuperSuit/blob/314831a7d18e7254f455b181694c049908f95155/supersuit/generic_wrappers/frame_stack.py#L50) where frame stracking only changes the `modify_obs` function. That way different wrappers can extend this base class.

What do you think @ulricharmel ?

ulricharmel: @KaleabTessera that's good idea. I will check what they do in depth.

mava/utils/environments/pettingzoo_utils.py
ulricharmel: I added this comment because we have [[this](https://github.com/instadeepai/Mava/blob/feature/frame_stacking/mava/utils/environments/pettingzoo_utils.py) in the atari pre-processing that already does the stacking. But I will prefer we have single approach to implement these wrappers as you suggested similar to supersuit

mava/wrappers/env_preprocess_wrappers.py
```diff
@@ -273,3 +274,145 @@ def __getattr__(self, name: str) -> Any:
...
+            timestep, env_extras = timestep
+        else:
+            env_extras = {}
+
+        new_observations: Any = {}
+
+        agent_0 = self._environment.possible_agents[0]
+        if type(timestep.observation[agent_0]) == OLT:
```
OmaymaMahjoub, line 312: Can you add docs for this if statement

mava/wrappers/env_preprocess_wrappers.py
```diff
@@ -273,3 +274,145 @@ def __getattr__(self, name: str) -> Any:
             return self.__getattribute__(name)
         else:
             return getattr(self._environment, name)
+
+
+class StackObservations:
```
OmaymaMahjoub, line 279: if there is the possibility of either:
1. Make abstract class related to the observation that has the basics methods for example:
```
@abc.abstractmethod
 def step(self, actions: Dict) -> dm_env.TimeStep:
         """Step the environment."""
        pass
```
2. Make the StackObservations inherit from ConcatAgentIdToObservation in order to not make redundant methods like observation_spec.
But this is of course optional!

OmaymaMahjoub (changes requested):
Thanks @ulricharmel, some minor requests, also we can chat about the TODO that you have in the test and make it functional in this PR

mava/wrappers/env_preprocess_wrappers.py
ulricharmel: Thanks @OmaymaMahjoub. I doubt this is still necessary actually. The first type I wrote this code, we had environments whose observations were not OLT named tuples so I used this if statement to cover both cases. I will remove this and see the tests passes.

OmaymaMahjoub: I was wandering why we have this if statement in this class and not in the ConcatAgentIdToObservation and ConcatPrevActionToObservation too?

mava/wrappers/env_preprocess_wrappers.py
ulricharmel: This has been done

mava/wrappers/env_preprocess_wrappers.py
ulricharmel: I have removed it and everything works now. Initially this use to throw issues in the wrappers test. But everything was trimmed recently and now we don't need this. Also the test I commented in the TODO now works. Also it used to be a problem because of the different observation types

OmaymaMahjoub (dismissed):
Thank you @ulricharmel

mava/wrappers/env_preprocess_wrappers.py
OmaymaMahjoub: Okay!

sash-a (changes requested):
Stacking looks good! Just see some minor comments and questions :medal_sports:

mava/wrappers/env_preprocess_wrappers.py
```diff
@@ -260,16 +264,94 @@ def obs_normalisation_start_index(self) -> int:
...
+        super().__init__(environment)
+
+        # Check that this is the first wrapper that is called on the environment
+        if self._environment.obs_normalisation_start_index != 0:
+            raise ValueError(
+                "Observation stacking should be the first custom\
+                    wrapper we call in the environment factory."
+            )
+
+        self.frames: Any = {
```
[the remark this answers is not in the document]
ulricharmel: I have change that to Dict[str, deque]. I didn't specify any type of the deque and this may vary per environment and observations. Also it will be another overhead to pull it out in the initialisation.

mava/wrappers/env_preprocess_wrappers.py
```diff
@@ -260,16 +264,94 @@ def obs_normalisation_start_index(self) -> int:
...
+        """Initialise wrapper."""
+
+        super().__init__(environment)
+
+        # Check that this is the first wrapper that is called on the environment
+        if self._environment.obs_normalisation_start_index != 0:
+            raise ValueError(
+                "Observation stacking should be the first custom\
+                    wrapper we call in the environment factory."
+            )
```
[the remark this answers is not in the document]
ulricharmel: Done

mava/wrappers/env_preprocess_wrappers.py
```diff
@@ -260,16 +264,94 @@ def obs_normalisation_start_index(self) -> int:
...
+
+        new_observations: Any = {}
+
+        old_observations = timestep.observation
+        for agent in self._environment.possible_agents:
+            agent_olt = old_observations[agent]
+            agent_observation = agent_olt.observation
+
+            for _ in range(self.num_frames):
+                self.frames[agent].append(agent_observation)
```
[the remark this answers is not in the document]
ulricharmel: Used this approach.

mava/wrappers/env_preprocess_wrappers.py
```diff
@@ -36,7 +38,48 @@
...
+    def step(self, actions: Dict) -> dm_env.TimeStep:
+        """Step the environment."""
+
+    def observation_spec(self) -> Dict[str, OLT]:
+        """Observation spec.
+
+        Returns:
+            types.Observation: spec for environment.
+        """
+        timestep = self.reset()
```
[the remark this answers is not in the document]
ulricharmel: Dries fixed this in the SMAC environment.

sash-a (approved):
Thanks @ulricharmel :+1:

OmaymaMahjoub (approved):
Thanks @ulricharmel  :fire:

PR #794, opened 2022-10-12
fix: remove integrity check

arnupretorius:
## What?
Fix broken code coverage. Seems to be related to a certification issue which I suspect comes from the integrity check
## How?
Remove integrity checks. We can look at adding these in later again since the issue is likely on codecov's side.


RuanJohn (approved):
Thank you @arnupretorius :muscle:

PR #797, opened 2022-10-12
Feat: Re-add Obs Networks.

KaleabTessera:
## What?
- Re-add obs networks. We have separate observation networks for the actor and the critic, which get optimized by different optimizers. In future, we should explore different variations of this. 
- Adapt assert that breaks on nested observations.
## Why?
- Certain envs uses obs networks e.g. pong. 
## How?
-
## Extra
-


mava/components/training/model_updating.py
```diff
@@ -295,16 +295,11 @@ def model_update_epoch(
...
-            # assert ...
-            assert (
-                list(batch.observations.values())[0].observation.shape[0]
-                == trainer.store.full_batch_size
-            )
+            # Check that our obs have correct batch shape
+            sample_observation = jax.tree_util.tree_leaves(
+                list(batch.observations.values())[0].observation
+            )[0]
+            assert sample_observation.shape[0] == trainer.store.full_batch_size
```
DriesSmit, line 302: Should we maybe loop through and check every observation?

mava/systems/ippo/networks.py
```diff
@@ -322,6 +323,7 @@ def critic_fn(inputs: jnp.ndarray) -> networks_lib.FeedForwardNetwork:
         """
         critic_network = hk.Sequential(
             [
+                observation_network,
```
DriesSmit, line 326: Are we now assuming a shared observation network if it is provided? However, our trainers are set up for separate networks. Would this not cause issues when two optimiser optimise the same parameters?

DriesSmit (commented):
Great work @KaleabTessera :raised_hands: Thanks for updating the README.md and TROUBLESHOOTING.mb :rocket: Just see my few comments.

mava/components/training/model_updating.py
KaleabTessera: I think it is to do only the first for now since I believe these obs within the same batch, have the same shape.

mava/systems/ippo/networks.py
KaleabTessera: We use seperate observation networks since Haiku enforces passing in a lambda function for the obs network.

DriesSmit: Thanks for the update. Separate networks sound better for now, yes.

mava/systems/ippo/networks.py
```diff
@@ -256,24 +256,29 @@ def policy_fn(inputs: jnp.ndarray) -> networks_lib.FeedForwardNetwork:
...
-                activation=activation_function,
-                w_init=w_init_fn(orthogonal_initialisation, jnp.sqrt(2)),
-                activate_final=True,
-                layer_norm=layer_norm,
-            ),
-        ]
+        policy_layers = []
+        if observation_network:
+            policy_layers.extend([observation_network])
+        policy_layers.extend(
```
DriesSmit, line 262: Minor comment, but we can probably just append for both the above lines.

mava/systems/ippo/networks.py
```diff
@@ -326,7 +331,10 @@ def critic_fn(inputs: jnp.ndarray) -> networks_lib.FeedForwardNetwork:
         Returns:
             FeedForwardNetwork class
         """
-        critic_network = hk.Sequential(
+        critic_layers = []
+        if observation_network:
+            critic_layers.extend([observation_network])
```
DriesSmit, line 336: Same with this line.

mava/systems/ippo/networks.py
sash-a, line 262: I think keep this consistent with below and append the item instead of extending a list of the item

mava/systems/ippo/networks.py
```diff
@@ -326,7 +331,10 @@ def critic_fn(inputs: jnp.ndarray) -> networks_lib.FeedForwardNetwork:
...
+        critic_layers.extend(
```
sash-a, line 337: Same here

mava/systems/ippo/networks.py
```diff
@@ -453,6 +466,7 @@ def make_default_networks(
     orthogonal_initialisation: bool = False,
     activation_function: Callable[[jnp.ndarray], jnp.ndarray] = jax.nn.relu,
     policy_network_head_weight_gain: float = 0.01,
+    observation_network: Callable = None,
```
sash-a, line 469: `Optional[Callable]`

DriesSmit (commented):
Thanks, @KaleabTessera :slightly_smiling_face: Just see my minor comments.

mava/systems/ippo/networks.py
```diff
@@ -377,6 +386,7 @@ def make_networks(
     policy_recurrent_layer_sizes: Sequence[int],
     recurrent_architecture_fn: Any,
     policy_layers_after_recurrent: Sequence[int],
+    observation_network: Callable = None,
```
sash-a, line 389: `Optional[Callable]`

mava/components/training/model_updating.py
sash-a: Seems like an expensive check to do every update is there a reason it was commented out originally?

mava/systems/ippo/networks.py
```diff
@@ -202,11 +202,11 @@ def make_discrete_networks(
...
```
sash-a, line 205: `Optional[Callable]`

tests/integration/networks_test.py
```diff
@@ -99,3 +133,27 @@ def test_feedforward_network(feedforward_network: Callable) -> None:
     """Feedforward network smoke test"""
 
     assert feedforward_network()
+
+
+def test_feedforward_network_with_obs(
```
sash-a, line 138: This would probably be an excessive test, but would be a good check to see if the network can handle a 2D observation being passed through now. Just pass something 2D through and assert that no error is thrown?

Feel free to ignore this one as it's a bit excessive

sash-a (changes requested):
Looks good, just very minor changes. 

If you could also open an issue for investigating a shared obs net, that would be great

mava/systems/ippo/networks.py
RuanJohn: This is the only question I have as well.

tests/systems/ippo_system_evaluation_test.py
```diff
@@ -29,6 +29,9 @@ def test_ippo_system_mt() -> System:
     return ippo_system_multi_thread_eval()
 
 
+# TODO Re-add once issue is solved
+# https://github.com/instadeepai/Mava/issues/842
+@pytest.mark.skip
```
DriesSmit, line 34: Can we add this test back in if we merge Omayma's PR?

mava/components/training/model_updating.py
KaleabTessera: I am happy to remove this check completely. 

@DriesSmit you okay with this?

mava/systems/ippo/networks.py
KaleabTessera: Valid point. I changed it.

mava/systems/ippo/networks.py
KaleabTessera: So we need extend for line 337, since we have both `MLP_NORM` and `ValueHead`. You need extend to add multiple items.

mava/systems/ippo/networks.py
KaleabTessera: Done.

mava/systems/ippo/networks.py
KaleabTessera: Done.

tests/systems/ippo_system_evaluation_test.py
KaleabTessera: I believe so.

mava/components/training/model_updating.py
DriesSmit: I far as I understand `jit`, think this is fine. This function is jitted and therefore the assert is only computed once at compile time. See [here](https://github.com/instadeepai/Mava/pull/797/files). I think a loop might be even better as there is no hit to performance.

mava/systems/ippo/networks.py
RuanJohn: Can we split `MLP_NORM` and `ValueHead` and then append them separately instead of extending? 
This is quite minor, but seems nicer from a consistency standpoint in my opinion.

mava/systems/ippo/networks.py
KaleabTessera: I switched everything back to `extend` for consistency (better than having 3 appends and 1 extends).

mava/systems/ippo/networks.py
KaleabTessera: For consistency, I think it is better to have everything as an extends.

mava/systems/ippo/networks.py
KaleabTessera: For consistency, I think it is better to have everything as an extends.

mava/components/training/model_updating.py
KaleabTessera: Thanks @DriesSmit ! 

I think the main question here is whether this check gives us any useful information? If not, we should probably remove it.

RuanJohn (approved):
Thank you @KaleabTessera! :raised_hands: 
As discussed offline, I agree that it is better to have extends everywhere instead of an append/extend mix.

mava/components/training/model_updating.py
DriesSmit: I originally added the check to make sure that the shapes were correct when developing. But now that you mention it, that is what the tests are for. So I am happy to remove it :slightly_smiling_face:

mava/systems/ippo/networks.py
sash-a: Sure, either way. All extend or all append :smile:

mava/components/training/model_updating.py
sash-a: That's really cool, if we do have any other asserts in the training/executing loop we should really leave a comment with that link to explain that it's only executed once

KaleabTessera: So I removed this for :)

DriesSmit: Agreed @sash-a :+1:

mava/systems/ippo/networks.py
DriesSmit: I feel appends are cleaner for the first 3. But up to you :slightly_smiling_face:

mava/systems/ippo/networks.py
DriesSmit: See my above comment.

KaleabTessera:
> Looks good, just very minor changes.
> 
> If you could also open an issue for investigating a shared obs net, that would be great

I created an investigation issue for this. :+1:

sash-a (approved):
Nice thanks @KaleabTessera :fire:

RuanJohn (approved):
Thanks @KaleabTessera :call_me_hand:

PR #804, opened 2022-10-25
Feature/death masking normalisation

### LINKED ISSUE #810: [MAINTAIN]: prevent normalisation of certain features

- During observation normalisation, certain features such as agent's id and agent's previous actions are one hot encoded vectors and hence should not be normalise.
- Also if death masking is set to true we don't want to normalised observations that have been zeroed already by death masking.

ulricharmel:
## What?

- Prevent normalisation of concatenated attributes such agent ids and agent previous actions.

## Why?

- We don't to want loose one hot encoded vectors concatenated to observations.
- Also prevent normalisation of death masked agents as this will add change 0s to some values that can confuse the network in smac environments.

## How?

- Added a death_masked_agents abstract method to the environment wrappers to keep tracked of dead agents.
- Added an obs_normalisation_start_index abstract method to keep track of which features which shouldn't normalise. This assumes all the features we don't normalise are placed at the front of the array.

## Extra
[Previous benchmark results](https://www.notion.so/instadeep/Observation-normalisation-bugfix-30ed9fc2e89d465c8bec9a2722cf6df8)

[Latest benchmark results](https://www.notion.so/instadeep/Obs-normalisation-system-performance-b79ffbf19a7f46dc834eb5992909bd16)

[Latest PZ A/B test results with agent id fix](https://www.notion.so/instadeep/Obs-normalisation-with-agent-id-fix-on-PZ-f882f4f16a2d493ca881f3b0657d0c93)

[Latest smac A/B test results with agent id fix](https://www.notion.so/instadeep/Obs-normalisation-with-agent-id-fix-on-smac-448b5ed0abbf41aa8d87241224e52f61)

Closes #810


DriesSmit:
Thanks @ulricharmel :slightly_smiling_face: Can you split this PR into two PRs? One for the obs norm fix and one for the value norm fix.

mava/utils/environments/debugging_utils.py
```diff
@@ -36,9 +37,11 @@ def make_environment(
     return_state_info: bool = False,
     random_seed: Optional[int] = None,
     recurrent_test: bool = False,
+    concat_id: bool = False,
```
RuanJohn, line 40: Can we please change this variable name to `concat_agent_id` to match the naming in other environments? :)

ulricharmel: sure will do

mava/wrappers/env_preprocess_wrappers.py
```diff
@@ -71,12 +75,16 @@ def reset(self) -> dm_env.TimeStep:
...
                 timestep.step_type, timestep.reward, timestep.discount, new_observations
             ),
-            extras,
+            env_extras,
         )
 
     def step(self, actions: Dict) -> dm_env.TimeStep:
         """Step the environment and concat agent ID"""
-        timestep, extras = self._environment.step(actions)
+        timestep = self._environment.step(actions)
```
DriesSmit, line 83: What changed that resulted in the need to add this code?

DriesSmit (commented):
Thanks @ulricharmel :raised_hands: Great work :fire: It would be nice to benchmark this code :slightly_smiling_face: Just see my few comments.

mava/components/training/step.py
```diff
@@ -284,7 +284,7 @@ def get_behavior_values(
 
             # Denormalise the values here to keep the GAE function clean
             target_value_stats = states.target_value_stats
-            if hasattr(trainer.store, "target_running_stats_fn"):
+            if trainer.store.global_config.normalize_target_values:
```
RuanJohn, line 287: I like this a lot more than the hasattr :raised_hands:

mava/utils/jax_training_utils.py
```diff
@@ -58,13 +61,23 @@ def compute_running_mean_var_count(
...
-    new_var = M2 / tot_count
+    var = M2 / tot_count
+    std = jnp.sqrt(var)
     new_count = tot_count
-    new_std = jnp.sqrt(new_var)
+
+    # This assumes the all the features we don't want to
+    # normalise are all always at the front.
+    new_mean = jnp.zeros_like(mean)
+    new_var = jnp.zeros_like(var)
```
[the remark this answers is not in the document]
ulricharmel: - The variance is initially set to 0 to make sure we get the computation right at the beginning. The algorithm needs to track the variance. For the first batch we don't have any previous batch so the current variance is 0. But we can not use a variance of 0 for normalisation so the initial standard deviation is set to 1. Then after every computation the standard deviation will be modified to be the sqrt of the variance. 

- The initial count is set to a value close to 0 mainly to avoid any potential issue of having a division by zero.

mava/wrappers/env_preprocess_wrappers.py
ulricharmel: Certain environments mainly the debugging ones only return a timestep without any extras. So when I was testing on debugging environments, I realised we need to slight modify the code this way so that it handle such environments. Maybe if we delete all such environments we will not need this anymore

mava/components/training/step.py
ulricharmel: Cool

mava/utils/environments/debugging_utils.py
```diff
@@ -25,6 +25,7 @@
     SwitchGameWrapper,
     TwoStepWrapper,
```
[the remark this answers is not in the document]
ulricharmel: Since this is not directly related to this PR, I suggest we do it in a different PR to avoid maybe breaking something that actually has nothing to do with this PR.

mava/utils/jax_training_utils.py
```diff
@@ -121,9 +136,16 @@ def update_and_normalize_observations(
...
     )
-    upd_stats = compute_running_mean_var_count(stats, obs)
+    upd_stats = compute_running_mean_var_count(stats, obs, start_axes)
     norm_obs = normalize(upd_stats, obs)
 
+    # the following code makes sure we do not normalise
+    # death masked observations. This uses the assumption
+    # that all death masked agens have zeroed observations
+    sum_obs = jnp.sum(obs[:, start_axes:], axis=1)
+    mask = jnp.array(sum_obs != 0, dtype=obs.dtype)
```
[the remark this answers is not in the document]
ulricharmel: I have added an asset statement in the a parameter initialisation to feature if the observation space is 1D. So far I noticed we only had 1D features, so it is possible the features are 2D (like images). Then in that case, are we going to maybe use convolution neural networks?

mava/utils/jax_training_utils.py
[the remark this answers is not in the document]
DriesSmit: Sounds good :+1:

mava/utils/environments/debugging_utils.py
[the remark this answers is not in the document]
DriesSmit: Fair enough. I will create an issue for it.

mava/utils/jax_training_utils.py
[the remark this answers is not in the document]
DriesSmit: Great thanks. Yes if the features are 2D we might use conv nets. I am not sure if Kale-ab and them have tested this yet. But at least we are future-proofing the code.

mava/wrappers/env_preprocess_wrappers.py
DriesSmit: I see. Fair enough :)

mava/components/building/environments.py
```diff
@@ -68,6 +68,10 @@ def on_building_init_start(self, builder: SystemBuilder) -> None:
         )
         builder.store.extras_spec = {}
 
+        builder.store.obs_normalisation_start = (
+            self.config.environment_factory().obs_normalisation_start_index
+        )
+
```
sash-a, line 74: We call the env factory above, rather create the env once for both of these calls:

```
env = self.config.environment_factory()
...
MAEnvironmentSpec(env)
...
env.obs_normalisation_start_index
```

mava/systems/executor.py
```diff
@@ -113,16 +111,6 @@ def select_actions(
...
 
-        # Normalise the observations before selecting actions.
-        if self.store.global_config.normalize_observations:
-            observations_stats = self.store.norm_params[
-                constants.OBS_NORM_STATE_DICT_KEY
-            ]
-            for key in observations.keys():
-                observations[key] = normalize_observations(
-                    observations_stats[key], observations[key]
-                )
```
sash-a, line 124: Nice, thanks for moving this :+1:

mava/utils/jax_training_utils.py
```diff
@@ -58,13 +65,23 @@ def compute_running_mean_var_count(
...
+
+    # This assumes the all the features we don't want to
+    # normalise are all always at the front.
+    new_mean = jnp.zeros_like(mean)
+    new_var = jnp.zeros_like(var)
+    new_std = jnp.ones_like(std)
+
+    new_mean = new_mean.at[start_axes:].set(mean[start_axes:])
+    new_var = new_var.at[start_axes:].set(var[start_axes:])
+    new_std = new_std.at[start_axes:].set(std[start_axes:])
```
sash-a, line 84: Didn't we say we wanted this to be a slice and not a single number? What if the observations we want to normalize are between observations we don't want to normalize?

So instead of start axes I would have something like `normalization_indicies` where the user could pass in something like `slice(2,10)` or just a numpy array of the indices required to be normalised.

sash-a (changes requested):
Looks good! Main thing is would just change how we specify what axes to mask

mava/components/building/environments.py
ulricharmel: Cool will do that

mava/utils/jax_training_utils.py
ulricharmel: I checked and the features we don't want to normalise agent ids and agent previous actions are always added at the front of the the features array.  Given that the agent_ids and actions are always added to the front. We need to adopt a convention so that any other features we don't want to normalise are always at the front. Because we may have issues like we have features maybe from slice(10, 20) we don't want to normalise.  Then the user wraps this environment with concatenate agent ids the adds the agent one hot encoded vectors at the front. Let say we have 5 agents and that the initial number of features was 40. So with concatenate agent  ids now we have 45 features. Also we don't want to normalise from Slice (0, 5) for agent ids and now from Slice (15, 25). This will become very trick to maintain.

mava/systems/executor.py
ulricharmel: Cool

mava/components/building/environments.py
ulricharmel: Done

mava/utils/jax_training_utils.py
sash-a: I see what you are saying, but I think solution I proposed adds more options in the future.  
Ideally we will keep putting things we do not want to normalize at the front and as such getting the indicies will be simple, but it is possible at some in the future that we need to not normalize something in the middle of the observations and with the current method it would not be possible.  
With the proposed method you could pass in [5,6,...,14,15,25,26...] and you would be able to skip normalizing the beginning and the middle of the observations. While you are correct in saying that it would be hard to maintain an env that has something we don't want to normalize in the middle I would rather have the option to do this in the future if we need to, while keeping convention that non-normalized observations stay at the front. 

I think that keeps the maintainability and leaves room for customization if needed, what do you think?

Also @DriesSmit and @RuanJohn what are your thoughts?

mava/utils/environments/debugging_utils.py
```diff
@@ -63,6 +66,9 @@ def make_environment(
             env_module, return_state_info=return_state_info
         )
 
+    if concat_agent_id:
```
DriesSmit, line 69: This might not work for all environments. In some future environments, the features might not be a 1D vector. So we can maybe just keep it at the environment level?

mava/wrappers/env_preprocess_wrappers.py
DriesSmit: A followup question. Why is extras needed now when it was not in the past for this environment?

mava/utils/jax_training_utils.py
DriesSmit: I like @sash-a's idea :slightly_smiling_face: It keeps it more general. If we wrap the environment the values, that should be normalised, should automatically be different in the same way `obs_normalisation_start` would be different.

mava/utils/environments/debugging_utils.py
DriesSmit: Why is it needed that every environment has a `concat_agent_id`?

mava/utils/jax_training_utils.py
ulricharmel: @sash-a @DriesSmit let's use this approach. We should chat about it more at the standup.

RuanJohn: I like the future-proofing that your idea adds @sash-a. Along with the slices, can we set the default to be a slice that is inferred from the observations specs? That way, by default the slice could be something like `[-obs_dim:]` without a user having to explicitly pass in a slice. The user can just turn observation normalisation on. In the case of more complex use cases, like what we are future-proofing for, a user could then pass in a more complex slice.

mava/utils/jax_training_utils.py
```diff
@@ -106,7 +123,9 @@ def denormalize(
 
 
 def update_and_normalize_observations(
-    stats: Dict[str, Union[jnp.array, float]], observation: OLT
+    stats: Dict[str, Union[jnp.array, float]],
+    observation: OLT,
+    start_axes: int = 0,
```
RuanJohn, line 128: Please see the above comment about having a default slice instead of an integer.

RuanJohn: Or maybe it should be able to take in both an int and a list?

mava/wrappers/env_preprocess_wrappers.py
RuanJohn: Another way around this is to make the environment return an empty dictionary as its extras.

mava/utils/environments/debugging_utils.py
ulricharmel: Yes actually the make_environment functions are loose and we expect the user to be more informed about his environment. I only included this here to be general but the default value is False. I thought concatenating agent id and previous actions were tricks we wanted to make available to any environment via an environment wrapper. Then it's up to the user to use this wrapper or not, if the wrapper is actually compatible with his environment. Because I think in general the user has to provide this make_environment function with all the tools to wrap his environment.

mava/utils/jax_training_utils.py
ulricharmel: @DriesSmit @sash-a @RuanJohn I have made changes to use slices. We still keep the start_index mainly for concatenate_agent_id and previous_options but the use can pass a list for other axes he doesn't want to normalise. The implementation for now assumes 1D features spaces, so it will through error if we if the observation shape has more than 1 dimension.

mava/wrappers/env_preprocess_wrappers.py
ulricharmel: Actually I just saw that in different areas in Mava, this trick is applied. So maybe I think it wasn't include only in these wrappers. You can see the same [here](https://github.com/instadeepai/Mava/blob/develop/mava/environment_loop.py#L120). I prefer we always do this rather than tracking down every environment.

@DriesSmit regarding your question. The way the wrappers are implemented is that their step function calls the step function of main environemnt. Then modifies the output before returning it to the executor. The current implementation assumed every wrapper returns a timestep and an extras dict. I only modified this to make it capable of wrapping the two different kinds of environments currently supported in Mava.

DriesSmit: Yes I code it up long ago to handle both types of environments. Okay it seem fine then :slightly_smiling_face: Thanks @ulricharmel. @RuanJohn what do you think?

RuanJohn: This looks good to me in that case. Thank you @ulricharmel :)

mava/utils/jax_training_utils.py
```diff
@@ -37,13 +40,50 @@ def init_norm_params(stats_shape: Tuple) -> Dict[str, Union[jnp.array, float]]:
...
+        for x in axes_list:
+            if type(x) == tuple or type(x) == list:
+                element = slice(x[0] + start_axes, x[1] + start_axes)
+                indices.append(x[1] + start_axes)
+            else:
+                element = slice(x + start_axes, x + start_axes + 1)
+                indices.append(x + start_axes + 1)
+            return_list.append(element)
+
+        assert (
```
RuanJohn, line 70: Minor detail, but can we please raise an error here instead of the assert?

mava/utils/jax_training_utils.py
```diff
@@ -37,13 +40,50 @@ def init_norm_params(stats_shape: Tuple) -> Dict[str, Union[jnp.array, float]]:
...
+        return_list = []
+        indices = []
+        for x in axes_list:
+            if type(x) == tuple or type(x) == list:
+                element = slice(x[0] + start_axes, x[1] + start_axes)
+                indices.append(x[1] + start_axes)
+            else:
+                element = slice(x + start_axes, x + start_axes + 1)
+                indices.append(x + start_axes + 1)
+            return_list.append(element)
```
sash-a, line 68: Will discuss offline, but it is very hard to follow what is happening here

mava/utils/jax_training_utils.py
```diff
@@ -37,13 +40,50 @@ def init_norm_params(stats_shape: Tuple) -> Dict[str, Union[jnp.array, float]]:
...
+) -> Tuple[slice, ...]:
+    """Construt a list of Tuples containing the features on which to apply normalisation
+
+    start_axes (int) --- default axes from which to start
+    axes_list (List[int]) --- user specified
+    obs_shape (Tuple) --- observations shape
+
+    Returns:
+        axes_list: a tuple to be used with np.r_
+    """
```
sash-a, line 54: Please stick to how we do docstrings for other methods

mava/utils/jax_training_utils.py
```diff
@@ -58,13 +98,24 @@ def compute_running_mean_var_count(
...
     new_count = tot_count
-    new_std = jnp.sqrt(new_var)
+
+    # This assumes the all the features we don't want to
+    # normalise are all always at the front.
+    new_mean = jnp.zeros_like(mean)
+    new_var = jnp.zeros_like(var)
+    new_std = jnp.ones_like(std)
+
+    indices = np.r_[axes]
```
sash-a, line 115: I think if you just passed in axes as an `np.ndarray` you would be able to avoid this `np.r_` call and should simplify this whole process substantially

sash-a (changes requested):
I think the current changes have made it quite difficult to follow how you are doing the whole indicies thing, just need to add some comments and simplify a little bit :smile:

mava/utils/jax_training_utils.py
```diff
@@ -37,13 +40,50 @@ def init_norm_params(stats_shape: Tuple) -> Dict[str, Union[jnp.array, float]]:
...
+        ), "Choosen indices will lead to out of bounds index error"
+
+    return tuple(return_list)
+
+
 def compute_running_mean_var_count(
-    stats: Dict[str, Union[jnp.array, float]], batch: jnp.ndarray
+    stats: Dict[str, Union[jnp.array, float]],
+    batch: jnp.ndarray,
+    axes: Any = slice(0, 1),
```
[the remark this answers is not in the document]
ulricharmel: Yes the default is the first element only but the axes are always passed when dealing with observations. This mainly for the target values which are always arrays with only one element.

mava/utils/jax_training_utils.py
ulricharmel: Modified to raise a ValueError

mava/wrappers/env_preprocess_wrappers.py
RuanJohn: Is it possible to move this fix to its own PR? That way we can get it in and the PettingZoo agent ID fix will lead to IDs that can be concatentated.

mava/utils/jax_training_utils.py
```diff
@@ -44,34 +44,53 @@ def construct_norm_axes_list(
...
 
     Returns:
         axes_list: a tuple to be used with np.r_
+
+    The start_axes is 0 unless we use a wrapper like concat_agent_id
+    or concat_previous_actions which add one hot encorded vectors
+    at the start of the array.
+    axes_list corresponds to user specified axes we want to normalise
+    We aussume the user does not consider contenation when specifying axes_list.
+    If axes_list is empty then we nornmalise all the axes
```
sash-a, line 60: I think a slight function change here would be useful, if axes_list is `None` then normalize everything and if it is empty normalize nothing.

mava/utils/jax_training_utils.py
```diff
@@ -44,34 +44,53 @@ def construct_norm_axes_list(
     start_axes: int, axes_list: List[Any], obs_shape: Tuple
```
sash-a, line 44: Can you rename `axes_list` to `elements_to_norm` and make the type `List[Union[Int],Tuple[Int,Int]]` (depending on if you decide to keep or remove the extra case for tuples inside the list)

Also can you remove `start_axis` and make it the first element of `elements_to_norm`

mava/utils/jax_training_utils.py
```diff
@@ -44,34 +44,53 @@ def construct_norm_axes_list(
...
         return_list = []
-        indices = []
+        starts = []
+        ends = []
         for x in axes_list:
             if type(x) == tuple or type(x) == list:
                 element = slice(x[0] + start_axes, x[1] + start_axes)
-                indices.append(x[1] + start_axes)
+                starts.append(x[0] + start_axes)
+                ends.append(x[1] + start_axes)
```
sash-a, line 81: As discussed offline, you can keep or remove this, up to you

mava/utils/jax_training_utils.py
```diff
@@ -44,34 +44,53 @@ def construct_norm_axes_list(
     start_axes: int, axes_list: List[Any], obs_shape: Tuple
 ) -> Tuple[slice, ...]:
     """Construt a list of Tuples containing the features on which to apply normalisation
-
-    start_axes (int) --- default axes from which to start
-    axes_list (List[int]) --- user specified
-    obs_shape (Tuple) --- observations shape
+    Args:
+        start_axes (int) --- default axes from which to start
+        axes_list (List[int]) --- user specified
```
sash-a, line 49: Make the docstring more informative here:

List of elements to normalize, can be a list of ints for specifying each value to normalize or can contain a tuple for a range of values to normalize.

sash-a (changes requested):
Very minor comments, otherwise looks good :+1:

mava/utils/jax_training_utils.py
ulricharmel: I have decided to keep it. It just few lines of code and we never know if we may need it.

mava/utils/jax_training_utils.py
ulricharmel: we can't remove start_axis. start_axis is an information that comes from the environment. This is not provide by the user. Also I tried modify the type to be more explicit but this just leads to so many cryptic error messages from mypy.

mava/utils/jax_training_utils.py
ulricharmel: The code has been modified accordingly to follow this behaviour. Nice catch actually. Thanks @sash-a

sash-a (dismissed):
Looks good! Thanks Ulrich :fire:

mava/utils/debugging/environments/switch_game.py
```diff
@@ -0,0 +1,154 @@
...
+DIAL switch game implementation
+Adapted from https://arxiv.org/pdf/1605.06676.pdf.
+
+Actions:
+0 - Nothing
+1 - Tell
+"""
+
+
+class MultiAgentSwitchGame(gym.Env):
```
KaleabTessera, line 33: Was this re-added on purpose?

examples/debugging/simple_spread/feedforward/decentralised/run_ippo_with_value_norm.py
```diff
@@ -109,6 +110,12 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
         clip_value=True,
         normalize_target_values=True,
         normalize_observations=True,
+        normalize_axes=[
```
KaleabTessera, line 113: TBH, I am a bit worried using the `normalize_axes` is too complicated. Can we consider passing dict keys to select this? Maybe we can chat about this in standup?

mava/utils/debugging/environments/switch_game.py
DriesSmit: Good catch @KaleabTessera.

ulricharmel: Let me update the branch. This shouldn't be there.

ulricharmel: Thanks @KaleabTessera. This has been removed. I'm not sure how that got back in.

KaleabTessera (dismissed):
Thanks @ulricharmel ! This is excellent!

DriesSmit (approved):
Thanks @ulricharmel :raised_hands:

RuanJohn (approved):
Thank you @ulricharmel :raised_hands:

PR #805, opened 2022-10-26
fixed bug with target value normalisation when value clip is true

### LINKED ISSUE #811: [BUG]: Critic loss explodes with target value normalisation and value clipping both set to True
### Describe the bug

- Critic loss explodes when both target value normalisation and value clipping are set to True

### To Reproduce
1. Run any environment with Target value normalisation set to True. Then set the clip value to False and then to True.
2. Observe the behaviour of the critic loss in both cases.

ulricharmel:
## What?
Fixed bug with target value normalisation when clip_value is True
## Why?
The behaviour values are de-normalised in order to computed the advantages. To implement clip value, the new crtic values need to be compared with the old behaviour values but these were de-normalised.
## How?
Normalised the de-normalised behaviour values before sending it to the critic loss function.
## Extra
Closes #811 
[Benchmarking results. ](https://www.notion.so/Test-value-normalisation-bugfix-393d5f74562c4e43a36eb5dd70cd3a89)

### DIFF
--- a/examples/debugging/simple_spread/feedforward/decentralised/run_ippo_with_value_norm.py
+++ b/examples/debugging/simple_spread/feedforward/decentralised/run_ippo_with_value_norm.py
@@ -106,7 +106,7 @@ def main(_: Any) -> None:
         num_epochs=15,
         num_executors=1,
         multi_process=True,
-        clip_value=False,
+        clip_value=True,
         normalize_target_values=True,
         normalize_observations=True,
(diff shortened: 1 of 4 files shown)


DriesSmit (approved):
Looks good, thanks @ulricharmel :slightly_smiling_face:

RuanJohn (approved):
Thank you @ulrich :)

PR #806, opened 2022-10-26
Feature/new logo

arnupretorius:
## What?
New logo for mava
## Extra
Would be nice if we can add the "M" logo to github supported emoji list to use in the readme. Might look into this later.


sash-a (approved):
Where are the tests for this change?? :joy:

PR #808, opened 2022-10-26
Maintain: evaluation set based on executor or evaluator.

### LINKED ISSUE #807: [MAINTAIN] Set environment factory function based on executor or evaluator.
Set `evaluation` variable in environment factory based on if an executor or evaluator is in use.

### Describe the outcome you'd like
Set `evaluation` variable in environment factory based on `store.executor_environment`. 

### How do we know when maintenance is complete?
`evaluation` variable in environment factory will be set based on if an executor or evaluator is us use.

siddarthsingh1:
## What?
`evaluation` set using store instead of always set to `True`.
## Why?
Allows for changes in environments during evaluation vs execution time.
## How?
`evaluation` set using `builder.store.is_evaluator` in `environments.py`.
## Extra
Closes #807


DriesSmit (approved):
Thanks @siddarthsingh1 :raised_hands: Looks good to me.

RuanJohn (approved):
Thank you @siddarthsingh1 :)

PR #814, opened 2022-10-27
Feature: best checkpointer

### LINKED ISSUE #744: [FEATURE] Checkpoint best params only
Checkpointer currently checkpoints latest parameters at fixed intervals

Checkpointer that adds eval params to the param server to check if the reward/ score has improved and then only save the new parameters

Checklist:
- [ ] Checkpointer component updated
- [ ] Checkpointer unit test updated
- [ ] PR merged into main

OmaymaMahjoub:
## What?
Checkpointer that adds eval params to the param server to check if the reward/ score has improved and then only save the new parameters
## Why?
Checkpointer currently checkpoints latest parameters at fixed intervals instead of checkpointing the best params/ network weights. This could help in future features such as the absolute metric
## How?
add new checkpointers that checkpoint the best parameters in different subdirectory
## Extra
Close #744 
## Performance
 **Restore the best network corresponding to the best win rate** 
image
**Restore the latest network**
image


mava/components/building/distributor.py
```diff
@@ -37,6 +37,7 @@ class DistributorConfig:
     single_process_max_episodes: Optional[int] = None
     is_test: Optional[bool] = False
     wait: Optional[bool] = False
+    metrics_checkpoint: Tuple = ("mean_episode_return",)
```
DriesSmit, line 40: Why do we save mean_episode_return and not just the best parameters? Is it use in the evaluation somehow?

mava/components/updating/checkpointer.py
```diff
@@ -31,6 +32,7 @@
 @dataclass
 class CheckpointerConfig:
     checkpoint_minute_interval: float = 5
+    restore_best_net: str = "None"
```
DriesSmit, line 35: Would it make sense to have a boolean here? True to restore and False to not restore from best checkpoints.

mava/components/updating/parameter_server.py
```diff
@@ -144,6 +147,24 @@ def on_parameter_server_init_start(self, server: SystemParameterServer) -> None:
         # Interrupt the system in case all the executors failed
         server.store.parameters["num_executor_failed"] = 0
 
+        # Initiate best performance network values
+        server.store.parameters["best_checkpoint"] = {}
+        for metric in server.store.metrics_checkpoint:
+            server.store.parameters["best_checkpoint"][metric] = {}
+            for agent_net_key in networks.keys():
+                server.store.parameters["best_checkpoint"][metric][
```
DriesSmit, line 155: Can we add a loop here?

DriesSmit: Furthermore, did you check that the two network values can be different? It seems that this is assigned by reference which would allow the best_checkpoint values to be updated if the network values are updated. Or does this not happen?

mava/components/updating/parameter_server.py
```diff
@@ -194,6 +215,10 @@ def on_parameter_server_set_parameters(self, server: SystemParameterServer) -> N
                 # # Loop through tuple
                 # for var_i in range(len(server.store.parameters[var_key])):
                 #     server.store.parameters[var_key][var_i].assign(params[var_key][var_i])
+            elif var_key == "best_checkpoint":
```
DriesSmit, line 218: Could we possibly check for a type instead of looking at the var_key to make this more general. Maybe other methods can then also use this.

mava/environment_loop.py
```diff
@@ -260,6 +264,20 @@ def step_executor() -> None:
                     if hasattr(self._environment, "get_interval_stats"):
                         results.update(self._environment.get_interval_stats())
                     self._logger.write(results)
+                    # Best_performance_update
+                    for metric in self._executor.store.metrics_checkpoint:
+                        assert (
+                            metric in results.keys()
+                        ), f"The metric chosen to checkpoint it best performance doesn't exist.\
```
DriesSmit, line 271: This is really cool. I did not know you can write comments in asserts :)

mava/utils/checkpointing_utils.py
```diff
@@ -0,0 +1,72 @@
...
+        params[metric][f"critic_network-{agent_net_key}"] = executor.store.networks[
+            agent_net_key
+        ].critic_params
+        params[metric][
+            f"policy_opt_state-{agent_net_key}"
+        ] = executor.store.policy_opt_states[agent_net_key]
+        params[metric][
+            f"critic_opt_state-{agent_net_key}"
+        ] = executor.store.critic_opt_states[agent_net_key]
+        executor.store.executor_parameter_client.set_and_wait(
```
DriesSmit, line 42: Can you maybe chance this to async?

mava/utils/checkpointing_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    ), "Can't find the restored best network checkpointed"
+    assert (
+        metric in server.store.parameters["best_checkpoint"].keys()
+    ), f"The metric chosen does not exist in the checkpointed networks.\
+        The best checkpointed network only available for the following\
+            metrics {server.store.parameters['best_checkpoint'].keys()}"
+    network = server.store.parameters["best_checkpoint"][metric]
+    # Update network
+    for agent_net_key in server.store.agents_net_keys:
+        server.store.parameters[f"policy_network-{agent_net_key}"] = network[
```
DriesSmit, line 61: Can you add a loop here?

DriesSmit (commented):
Thanks @OmaymaMahjoub :fire: This looks really nice :slightly_smiling_face: Just see my minor comments. Could you possibly create an example that shows how a system trains and gets reloaded for evaluation?

mava/components/building/distributor.py
OmaymaMahjoub: This because we can checkpoint the best performance depending on multiple metrics (e.g. best network for win rate it can be not the best network for the return)

mava/components/updating/checkpointer.py
OmaymaMahjoub: Same as above, we have multiple best checkpointed network each belong to a specific metric so we need config to mention if someone want to have win rate best net or reward best net..

mava/components/updating/parameter_server.py
OmaymaMahjoub: I will check that is a deep copy or not because from the unit tests it seems to works fine but I will re check it

mava/components/updating/parameter_server.py
OmaymaMahjoub: I was thinking of type dict but I ended up making that case happen for the normal network too since it is also a nested and multu dim dict

mava/environment_loop.py
OmaymaMahjoub: Hh me too, I founded out through looking at step function when I had to make unit test the one who implemented the step fn used the assert with msg so I learned from it.

mava/utils/checkpointing_utils.py
OmaymaMahjoub: Okay I will give it a try

mava/components/updating/parameter_server.py
OmaymaMahjoub: You were right it was reference

OmaymaMahjoub: Solved!

DriesSmit: Great :slightly_smiling_face:

ulricharmel (commented):
Thanks @OmaymaMahjoub. This looks great. I can't pick any obvious issue.

mava/components/updating/parameter_server.py
DriesSmit: Maybe we can have a chat about this? :)

OmaymaMahjoub: Already updated the function, where we don't need this if statement anymore

DriesSmit: Great.

mava/utils/checkpointing_utils.py
OmaymaMahjoub: It's hard to make it in a same loop but if you have any idea I will be thankful to apply it

mava/components/building/distributor.py
```diff
@@ -37,6 +37,8 @@ class DistributorConfig:
...
```
sash-a, line 40: I would say maybe add a comment here explaining what this variable does, as it is not obvious. Something like `# saves networks based on this metric`. 

For the same reason I'd rename the variable to `checkpointing_metric`, but this is debatable, you can leave it as is if you feel that is better.

mava/components/building/parameter_client.py
```diff
@@ -116,6 +117,32 @@ def on_building_executor_parameter_client(self, builder: SystemBuilder) -> None:
         params["norm_params"] = builder.store.norm_params
         get_keys.append("norm_params")
 
+        # Create best performance network params in case of evaluator
```
sash-a, line 120: I think this method is getting quite long and would benefit from being broken up into smaller methods. This would also help when implementing other algorithms (such as value based ones) as those don't have a policy and critic network, so we could just override the smaller methods and leave most of the functionality untouched. 

^Just a comment for next time, don't make these changes here.

mava/components/updating/checkpointer.py
```diff
@@ -51,12 +53,20 @@ def on_parameter_server_init(self, server: SystemParameterServer) -> None:
...
             directory=server.store.experiment_path,
             add_uid=False,
             time_delta_minutes=0,
         )
+
+        # Check if the checkpointer restore the network
+        if old_trainer_steps != server.store.parameters["trainer_steps"]:
+            # Check if the user want network with best performance
+            if not self.config.restore_best_net == "None":
+                update_to_best_net(server, self.config.restore_best_net)
```
sash-a, line 68: Make this one `if`.

`if old_trainer_steps != server.store.parameters["trainer_steps"] and not self.config.restore_best_net == "None":`

mava/components/updating/checkpointer.py
sash-a, line 35: I would just make this `None` and not the string "None"

mava/environment_loop.py
```diff
@@ -291,6 +292,26 @@ def step_executor() -> None:
...
+                    if self._executor.store.checkpoint_best_perf:
+                        # Best_performance_update
+                        for (
+                            metric,
+                            best_performance,
+                        ) in self._executor.store.metrics_checkpoint.items():
+                            assert (
+                                metric in results.keys()
+                            ), f"The metric chosen to checkpoint it best performance doesn't exist.\
+                                This experiment has only the following metrics {results.keys()}"
```
sash-a, line 304: Grammar: The metric, {metric}, chosen for checkpointing doesn't exist. This experiment has only the following metrics {results.keys()}

sash-a (changes requested):
Very minor code style/grammatical errors, otherwise looks great! :fire:

mava/components/updating/checkpointer.py
OmaymaMahjoub: This is because of a mypy issue but I will change it and write `#type: ignore`

mava/environment_loop.py
OmaymaMahjoub: Thank you! Will make the change!

mava/components/building/parameter_client.py
OmaymaMahjoub: We can handle it in this PR, it's fine I will make the changes!

mava/components/updating/checkpointer.py
sash-a: I think rather make it a `Union[str, None]` type

mava/environment_loop.py
OmaymaMahjoub: Done!

mava/components/updating/checkpointer.py
OmaymaMahjoub: Nice idea! Done

mava/components/building/parameter_client.py
OmaymaMahjoub: I created an issue for this request https://github.com/instadeepai/Mava/issues/831

mava/utils/checkpointing_utils.py
DriesSmit: No problem then.

mava/components/building/distributor.py
```diff
@@ -75,6 +78,14 @@ def on_building_program_nodes(self, builder: SystemBuilder) -> None:
         # Save number of the executors
         builder.store.num_executors = self.config.num_executors
 
+        # Save list of metrics attached with their best performance
+        builder.store.checkpointing_metric: Dict[str, Any] = {}  # type: ignore
```
DriesSmit, line 82: Should this not be in a checkpointing component?

mava/components/updating/checkpointer.py
```diff
@@ -51,12 +53,21 @@ def on_parameter_server_init(self, server: SystemParameterServer) -> None:
...
+        old_trainer_steps = server.store.parameters["trainer_steps"].copy()
         server.store.system_checkpointer = acme_savers.Checkpointer(
             object_to_save=saveable_parameters,  # must be type saveable
             directory=server.store.experiment_path,
             add_uid=False,
             time_delta_minutes=0,
         )
+
+        # Check if the checkpointer restore the network
+        # and if the user want network with best performance.
```
DriesSmit, line 65: Check if the checkpointer restored the network parameters and if the user wants the network with the best performance.

DriesSmit (commented):
Thanks @OmaymaMahjoub :raised_hands: This looks really good!

mava/components/building/distributor.py
OmaymaMahjoub: The issue here is that we want to share the metrics names between the executor in the env loop, the checkpointer and the parameter server so the best place to have it for all the three is the distributor

sash-a (dismissed):
Thanks for making the changes, all looks good! :+1:

mava/components/building/distributor.py
DriesSmit: I don't completely understand. Maybe we can have a chat :sweat_smile:

OmaymaMahjoub: Done!

DriesSmit (approved):
Thanks @OmaymaMahjoub :raised_hands:  Really great work :fire:

sash-a (approved):

PR #815, opened 2022-10-28
Bugfix/pettingzoo agent id concatenation

RuanJohn:
## What?
Agent IDs were not being concatenated to observations for PettingZoo environments since the environment with agent IDs added was called `env` but the `make_environment` function was returning `environment`
## Why?
To fix the bug. 
## How?
Changed the variable name. 
## Extra
Raise a `NotImplementedError` in the `ConcatAgentIdToObservation` when trying to concatenate agent IDs to a non-vector observation. 
Has the same error now in the `ConcatAgentIdToObservation` as mentioned [here](https://github.com/instadeepai/Mava/pull/804#discussion_r1006432340) since certain PettingZoo environments do not return extras. So this depends on that PR.


DriesSmit (approved):
Great catch! :fire: Tanks @RuanJohn :slightly_smiling_face:

mava/utils/environments/pettingzoo_utils.py
```diff
@@ -131,6 +131,6 @@ def make_environment(
         raise Exception("Pettingzoo is not installed.")
 
     if concat_agent_id:
-        env = ConcatAgentIdToObservation(env)
+        environment = ConcatAgentIdToObservation(environment)  # type: ignore
```
ulricharmel, line 134: Thanks @RuanJohn . Okay cool I didn't know about the # type ignore flag. I actually added a fix to this in the frame staking PR but I had to define env as type Any to bypass the pre-commit type error.

ulricharmel (approved):
Thanks @RuanJohn

mava/utils/environments/pettingzoo_utils.py
RuanJohn: Thank you @ulricharmel, I try to use it very sparingly :sweat_smile: but in this case all my ideas for getting around the type checking made the code much less clean to read. The Any idea is also good.

PR #817, opened 2022-10-31
fix: Agent id observation mismatch

DriesSmit:
## What?

- This PR updates the environment loop print statements.
- It also updates the code to avoid shuffling of agent observations in the trainer.

## Why?

- The new catch statement in the environment loop prevents Launchpad from printing when using Docker.
- While the agent observation issue was not specifically a bug, it does lead to confusion when using a single trainer. This has now been fixed.

## How?

- This code adds a sort statement in the base adder.
- It also changes the prints in the environment loop to tf.print.

## Extra
.

### DIFF
--- a/mava/components/updating/terminators.py
+++ b/mava/components/updating/terminators.py
@@ -27,6 +27,8 @@ from mava.core_jax import SystemParameterServer
 from mava.utils.lp_utils import termination_fn
 from mava.utils.training_utils import check_count_condition
 
+import tensorflow as tf
+
 
 class Terminator(Component):
     @abc.abstractmethod
@@ -101,7 +103,7 @@ class CountConditionTerminator(Terminator):
             and parameter_server.store.parameters[self.termination_key]
             > self.termination_value
         ):
-            print(
+            tf.print(
                 f"Max {self.termination_key} of {self.termination_value}"
                 " reached, terminating."
             )
@@ -161,7 +163,7 @@ class TimeTerminator(Terminator):
             None.
         """
         if time.time() - self._start_time > self.config.run_seconds:
-            print(
+            tf.print(
                 f"Run time of {self.config.run_seconds} seconds reached, terminating."
             )
             self.config.termination_function(parameter_server)
--- a/mava/components/updating/terminators.py
+++ b/mava/components/updating/terminators.py
@@ -18,9 +18,9 @@ import abc
 import time
 from typing import Any, Callable, Dict, List, Optional, Type
 
-import tensorflow as tf
 from chex import dataclass
 
+import logging
 from mava.callbacks import Callback
 from mava.components.component import Component
 from mava.components.updating.parameter_server import ParameterServer
@@ -102,7 +102,7 @@ class CountConditionTerminator(Terminator):
             and parameter_server.store.parameters[self.termination_key]
             > self.termination_value
         ):
-            tf.print(
+            logging.exception(
                 f"Max {self.termination_key} of {self.termination_value}"
                 " reached, terminating."
             )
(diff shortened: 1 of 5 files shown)


mava/environment_loop.py
```diff
@@ -278,15 +279,15 @@ def step_executor() -> None:
 
             except Exception as e:
                 if self._executor._evaluator:
-                    print(
+                    tf.print(
```
OmaymaMahjoub, line 282: Nice idea @DriesSmit, if you can add it to the `mava/systems/trainer.py` too and remove time.sleep(60) since we don't need it anymore!

sash-a, line 282: Can we not use `jax.debug.print`?

OmaymaMahjoub (changes requested):
Thanks @DriesSmit :fire:

sash-a (approved):
Looks good, I just think ideally we'd like to avoid using `tf` as much as possible, but if this `tf` print is something we can't do another way then I wouldn't worry about it

mava/environment_loop.py
DriesSmit: Yip I agree. I initially tried using `jax.debug.print`, but it did not seem to work. So I am not sure if there is another way to print that does not use TensorFlow. Can you test if `tf.print` shows on your side when introducing a bug? You use a virtual environment right?

DriesSmit: Good point :) We don't need `time.sleep` anymore. I updated the code. Just make sure it works on your side as expected.

OmaymaMahjoub (changes requested):
@DriesSmit one more request about tf.print if you can also add it in the terminator side: `mava/components/updating/terminators.py` because the system interrupt without sending the prints in the terminator functions

OmaymaMahjoub (dismissed):
Thank you @DriesSmit  :raised_hands:

mava/components/updating/terminators.py
```diff
@@ -101,7 +102,7 @@ def on_parameter_server_run_loop_termination(
             and parameter_server.store.parameters[self.termination_key]
             > self.termination_value
         ):
-            print(
+            tf.print(
```
KaleabTessera, line 105: Can we replace this with `logging.exception(` or `logging.error` . They seemed to print when I tested locally.

KaleabTessera (changes requested):

mava/components/updating/terminators.py
DriesSmit: Thanks @KaleabTessera :raised_hands: Great find.

KaleabTessera (approved):
Thanks @DriesSmit! This is a great fix!

mava/components/updating/terminators.py
```diff
@@ -101,7 +102,7 @@ def on_parameter_server_run_loop_termination(
             and parameter_server.store.parameters[self.termination_key]
             > self.termination_value
         ):
-            print(
+            logging.exception(
```
OmaymaMahjoub, line 105: I tried to use `logging.exception`, it shows the message but also it shows another msg `NoneType: None` for example:
```
[evaluator/0] E1108 11:08:43.800992 140064315311936 environment_loop.py:299] Terminate the system
[evaluator/0] NoneType: None
```
And if I raise an error it shows only the error without the msg
```
[evaluator/0] Traceback (most recent call last):
[evaluator/0]   File "/home/omayma/Desktop/Mava/mava/environment_loop.py", line 281, in run
[evaluator/0]     step_executor()
[evaluator/0]   File "/home/omayma/Desktop/Mava/mava/environment_loop.py", line 265, in step_executor
[evaluator/0]     raise ValueError
[evaluator/0] ValueError
```

OmaymaMahjoub: Only the tf.print but we can keep this one for now and make an issue or more investigation in the future if we can find jax debug print or something similar

OmaymaMahjoub (approved):
Thanks @DriesSmit :fire:

ulricharmel (approved):
Thanks @DriesSmit

mava/components/updating/terminators.py
DriesSmit: Sounds good.

PR #825, opened 2022-11-09
feature: made a general extras spec component

sash-a:
## What?
Made a general extras spec base componenent as opposed to the one that lived inside the IPPO folder, as it will be needed for most if not all systems

## Why?
To increase code re-use

## How?
Moved the already existing class into the general components folder


ulricharmel (approved):
Thanks @sash-a. Really nice PR and I love the way you comment everything nicely.

KaleabTessera (approved):
Thanks @sash-a ! Looks really clean :rocket:

PR #826, opened 2022-11-09
fix: remove 1D observation check from concat agent id wrapper

RuanJohn:
## What?
There was a bug leading to Maximum attainable SMAC episode returns being lower than what they should be. 
## Why?
It seems like due to jitting the if statement in the `ConcatAgentIdToObservation` class checking whether observations were 1 dimensional was causing an issue. 
## How?
Removed the check. 
## Extra
This check was added in in this [PR](https://github.com/instadeepai/Mava/pull/815) and was not originally included, so removing it should not be an issue.

### DIFF
--- a/mava/wrappers/env_preprocess_wrappers.py
+++ b/mava/wrappers/env_preprocess_wrappers.py
@@ -49,18 +49,6 @@ class ConcatAgentIdToObservation:
         self._environment = environment
         self._num_agents = len(environment.possible_agents)
 
-        # Check that observation of first agent is a vector
-        if (
-            len(
-                list(self._environment.observation_spec().values())[0].observation.shape
-            )
-            > 1
-        ):
-            raise NotImplementedError(
-                "Agent ID concatenation is only implemented for vector\
-                    based observations."
-            )
-
     def reset(self) -> dm_env.TimeStep:
         """Reset environment and concat agent ID."""
         timestep = self._environment.reset()


DriesSmit (commented):
Great find, thanks @RuanJohn :raised_hands: :rocket:   Just one question. How is this if statement causing a bug?

mava/wrappers/smac.py
```diff
@@ -46,10 +46,11 @@ def __init__(
             return_state_info: return extra state info
         """
         self._environment = environment
+
         self._return_state_info = return_state_info
         self._agents = [f"agent_{n}" for n in range(self._environment.n_agents)]
 
-        self._reset_next_step = True
+        self._is_resetted = False
```
sash-a, line 53: Minor grammatical thing, just rename to `_is_reset`, maybe even `_is_first_step` is a more descriptive, although not so sure about that one. 

I would also put a comment there explaining why we now do this, something like "because smac returns weird values if you reset twice in a row".

mava/wrappers/smac.py
```diff
@@ -64,10 +65,11 @@ def reset(self) -> dm_env.TimeStep:
             dm_env.TimeStep: dm timestep.
         """
         # Reset the environment
-        self._environment.reset()
+        if not self._is_resetted:
```
sash-a, line 68: Also would be nice to be explicit in the docstring, "Resets the env, if it was not reset on the previous step"

sash-a (changes requested):
Great work on finding the bug! Very minor comments, otherwise all looks good!

mava/wrappers/smac.py
DriesSmit: Agreed and fixed.

sash-a (approved):
Nice thanks @DriesSmit for the deep dive and @RuanJohn for finding the bug! :fire:

ulricharmel (approved):
Thanks @DriesSmit, I am happy with the changes.

PR #830, opened 2022-11-14
fix: bug optional MLP layer after recurrent layer is never called

ulricharmel:
## What?

- Optional MLP layer on hidden features after recurrent layers is never called.

## Why?

- The optional mlp layer was created but never appended to the policy network list.

## How?

- Appended the mlp layer to the network list.
- Add an if statement to check if the len of the recurrent layers is > 0.


DriesSmit (approved):
Great find, thanks @ulricharmel :raised_hands:

RuanJohn (approved):
Great find, thank you @ulricharmel :raised_hands:

PR #834, opened 2022-11-17
fix: pin haiku version

RuanJohn:
## What?
Pin the haiku version to fix the issue of all integration tests currently failing.


ulricharmel (approved):
Thanks @RuanJohn.

KaleabTessera (approved):
Thanks @RuanJohn !

PR #832, opened 2022-11-15
Feature/layer normalisation

ulricharmel:
## What?

- Added an option to apply layer normalisation to the hidden features of MLP networks.

## Why?

- To follow the implementation style of the [onpolicy repo](https://github.com/marlbenchmark/on-policy/blob/79f6591882088a0f583f7a4bcba44041141f25f5/onpolicy/algorithms/utils/mlp.py#L19)

## How?

- Clone haiku.nets.MLP and modified it to include a layer normalisation after every linear layer.
- Added a new network option to apply layer norm or not. Default is False.
- Modified every call to haiku.nets.MLP with the customise implementation. 

## Extra

- Closes #828

PR #835, opened 2022-11-21
Feature: obs normalisation moved to networks

ulricharmel:
## What?

- Add an option to do apply layer norm directly on the input network features.

## Why?

- This is similar to what [onpolicy repo](https://github.com/marlbenchmark/on-policy/blob/79f6591882088a0f583f7a4bcba44041141f25f5/onpolicy/algorithms/utils/mlp.py#L52) does. If this performs well then we'll let layer norm directly handle observations normalisation and removed all the code we added to implement our custom normalisation. 

## How?

- Added a normalise feature option to our custorm MLP norm. Default value is False. This applies a layer norm to the features before the MLP layers.

## Extra
Related to #832


DriesSmit:
Closing this PR as we found using the custom normalisation code works better.

PR #841, opened 2022-11-23
Feature: absolute metric

### LINKED ISSUE #763: [FEATURE] Absolute metric computation
Implement a component that can compute and log the absolute metric for a given experiment run. For a discussion of the absolute metric please see paragraph 1 on page 9 [here](https://arxiv.org/pdf/2209.10485.pdf). 

After an experiment has reached its termination condition, the best model parameters are loaded are the policy is rolled out for a fixed number of episodes. The raw episode returns and any other relevant metrics are logged to a json file. An important detail is that this computation should take place before the experiment is terminated. So the termination components must be updated. 

### Describe alternatives you've considered
An idea could be that the termination component is updated to only terminate once a `computed_absolute_metric`  boolean flag is set to true. The absolute metric can be computed on the evaluator process by updating the evaluator weights to the best model weights found during training and by passing in a custom number of evaluation episodes to complete. 

The absolute metric is succesfully computed and logged.

OmaymaMahjoub:
## What?
Implement a component that can compute and log the absolute metric for a given experiment run. For a discussion of the absolute metric please see paragraph 1 on page 9 [here](https://arxiv.org/pdf/2209.10485.pdf).
## Why?
This will a nice feature for releable results and will enable for the usage of the EMARL repo tools library since it demands that recorded absolute metric results exist.
## How?
Create `run_evaluation` method inside the `environment loop`
## Extra
Close #763 
To test the feature try to run `examples/smac/feedforward/decentralised/calculate_absolute_metric.py`


examples/smac/feedforward/decentralised/calculate_absolute_metric.py
```diff
@@ -0,0 +1,128 @@
...
+        evaluation_duration={"evaluator_episodes": 32},
+        logger_factory=logger_factory,
+        logger_config=logger_config,
+        # Flag to activate the calculation of the absolute metric
+        absolute_metric=True,
+        # How many episodes the evaluator will run for
+        absolute_metric_duration=32,
+        # When the system will terminate and the absolute metric
+        # will be calculated
+        absolute_metric_interval=50000,
```
RuanJohn, line 118: Can we not use the `termination_condition` dictionary here?

mava/environment_loop.py
```diff
@@ -292,12 +377,11 @@ def step_executor() -> None:
                     results.update(eval_result)
                     self._logger.write(results)
 
-                    global_config = self._executor.store.global_config
                     # ideally this would be executor.has(BestCheckpointer),
                     # but that causes circular import
-                    if (
-                        hasattr(global_config, "checkpoint_best_perf")  # yuck
-                        and global_config.checkpoint_best_perf
+                    if hasattr(global_config, "checkpoint_best_perf") and (  # yuck
```
RuanJohn, line 382: Maybe remove the `# yuck` comment.

examples/smac/feedforward/decentralised/calculate_absolute_metric.py
OmaymaMahjoub: I will give it a try

mava/environment_loop.py
OmaymaMahjoub: Done

mava/components/building/best_checkpointer.py
```diff
@@ -6,12 +7,17 @@
...
 
 
 @dataclass
 class BestCheckpointerConfig:
     checkpointing_metric: Tuple = ("mean_episode_return",)
     checkpoint_best_perf: bool = False
+    # Flag to calculate the absolute metric
+    absolute_metric: bool = False
+    # How many episodes to run evaluation for
+    absolute_metric_duration: Optional[int] = 320
```
RuanJohn, line 21: Is there any way we could have this be something like `evaluation_duration * 10` by default?

mava/components/building/best_checkpointer.py
```diff
@@ -30,19 +36,34 @@ def on_building_init(self, builder: SystemBuilder) -> None:
...
+            if (
+                not (server.store.global_config.termination_condition is None)
+                and "executor_steps"
+                in server.store.global_config.termination_condition.keys()
+            ):
+                server.calculate_absolute_metric = True  # type: ignore
+            else:
+                logging.exception(
+                    f"{ValueError}: To calculate the absolute metric, you need to define\
+                    a termination condition related to the executor_steps."
```
RuanJohn, line 64: Won't it work with just any termination condition?

mava/components/building/best_checkpointer.py
OmaymaMahjoub: We can set it as default in case someone didn't define the absolute metric duration :raised_hands:

mava/components/building/best_checkpointer.py
OmaymaMahjoub: I think only with the executor step because we need the number of steps to run the absolute metric, should we change it to be for any condition?

RuanJohn: In that case it is okay, since terminating by steps makes the most sense.

mava/components/building/best_checkpointer.py
RuanJohn: Thank you :raised_hands:

mava/environment_loop.py
```diff
@@ -245,6 +248,72 @@ def should_run_loop(eval_interval_condition: Tuple) -> bool:
...
+        def run_evaluation(results: Any) -> None:
+            """Calculate the absolute metric"""
+
+            if self._executor.has(BaseNormalisation) and (
+                self._executor.store.global_config.normalise_observations
+                or self._executor.store.global_config.normalise_target_values
+            ):
+                warnings.warn(
+                    """The calculation of the absolute metric in
+                the case of normalisation is not supported"""
```
RuanJohn, line 260: Does the absolute metric not work when normalising for the same reason as the checkpointer? ie. We cannot effectively keep track of the normalisation params?

OmaymaMahjoub: The issue it will take time to investigate how to track the normalized params since they are not the same format as the usual one

OmaymaMahjoub: Created an [issue](https://github.com/instadeepai/Mava/issues/859) for it.

mava/components/building/best_checkpointer.py
```diff
@@ -21,6 +21,11 @@ class BestCheckpointerConfig:
 
 
 class BestCheckpointer(Component):
+    """Best checkpointer is a component that help in storing the best network
+    params for two options:
```
RuanJohn, line 25: Minor typo: 
`Best checkpointer is a component that help in storing the best network params for two options` -> `A component that stores the best network paramaters for two options`

OmaymaMahjoub: Done!

mava/components/building/best_checkpointer.py
```diff
@@ -6,15 +7,26 @@
...
 
 
 @dataclass
 class BestCheckpointerConfig:
     checkpointing_metric: Tuple = ("mean_episode_return",)
     checkpoint_best_perf: bool = False
+    # Flag to calculate the absolute metric
+    absolute_metric: bool = False
+    # How many episodes to run evaluation for
+    absolute_metric_duration: Optional[Any] = None
```
sash-a, line 20: Nice addition with the comments! Maybe add them for the other two config options also

mava/components/building/best_checkpointer.py
```diff
@@ -6,15 +7,26 @@
...
 
 
 class BestCheckpointer(Component):
+    """A component that stores the best network parameters for two options:"""
+
+    """ 1- Checkpointing the best network for a specific metric.
+    2- Calculating the absolute metric.
+    """
+
```
sash-a, line 29: Can you make this 1 docstring?

mava/components/building/best_checkpointer.py
```diff
@@ -30,19 +42,42 @@ def on_building_init(self, builder: SystemBuilder) -> None:
...
+            builder.store.is_evaluator
+            and (self.config.checkpoint_best_perf or self.config.absolute_metric)
+        ):
             return
 
+        if self.config.absolute_metric_duration is None:
+            self.config.absolute_metric_duration = (
+                10
+                * builder.store.global_config.evaluation_duration["evaluator_episodes"]
+            )
```
sash-a, line 55: to make this a bit easier to read on the line above just do
`episodes = builder.store.global_config.evaluation_duration["evaluator_episodes"]`
and then use the episode variable.

mava/components/updating/checkpointer.py
```diff
@@ -68,7 +68,10 @@ def on_parameter_server_init(self, server: SystemParameterServer) -> None:
         if (old_trainer_steps != server.store.parameters["trainer_steps"]) and (
             self.config.restore_best_net is not None
         ):
-            if server.has(BaseNormalisation):
+            if server.has(BaseNormalisation) and (
+                server.store.global_config.normalise_observations
+                or server.store.global_config.normalise_target_values
+            ):
```
sash-a, line 74: Nice catch!

There is a failure case here which is having only 1 of obs norm or value norm components and trying to access both of the config variables, but I think this is such a corner case that it probably doesn't matter.

sash-a (changes requested):
Very minor comments, looks good otherwise! :fire:

mava/components/building/best_checkpointer.py
OmaymaMahjoub: I tried but received an error related to mypy as it needs only one line for docstring

OmaymaMahjoub: I will try to find solution for it :raised_hands:

mava/components/building/best_checkpointer.py
OmaymaMahjoub: Will do thar!

mava/components/building/best_checkpointer.py
OmaymaMahjoub: Done!

mava/components/building/best_checkpointer.py
OmaymaMahjoub: Done!

mava/components/updating/checkpointer.py
OmaymaMahjoub: Done!

mava/components/building/best_checkpointer.py
OmaymaMahjoub: Done!

sash-a (changes requested):
Final tiny change

mava/components/updating/checkpointer.py
```diff
@@ -68,10 +68,18 @@ def on_parameter_server_init(self, server: SystemParameterServer) -> None:
...
+            normalisation = server.has(BaseNormalisation) and (
+                (
+                    hasattr(server.store.global_config, "normalise_observations")
+                    and server.store.global_config.normalise_observations
+                )
+                or (
+                    hasattr(server.store.global_config, "normalise_target_values")
+                    and server.store.global_config.normalise_target_values
+                )
+            )
```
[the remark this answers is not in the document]
OmaymaMahjoub: Do you think we won't have the error of the not existance of the normalise_observations in the global config?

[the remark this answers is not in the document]
OmaymaMahjoub: I removed this `server.has(BaseNormalisation)` but I keeped `hasattr` since we can't use `has`

mava/components/updating/terminators.py
```diff
@@ -97,6 +97,12 @@ def on_parameter_server_run_loop_termination(
         Returns:
             None.
         """
+        if (
+            hasattr(parameter_server, "calculate_absolute_metric")
```
ulricharmel, line 101: Did you try using parameter_server.has("calculate_absolute_metric")? So that we keep the consistency with using that custom "has" function Sasha implemented.

mava/environment_loop.py
```diff
@@ -245,14 +247,102 @@ def should_run_loop(eval_interval_condition: Tuple) -> bool:
...
+            normalisation = (
+                self._executor.has(ObservationNormalisation)
+                and self._executor.store.global_config.normalise_observations
+            ) or (
+                self._executor.has(ValueNormalisation)
+                and self._executor.store.global_config.normalise_target_values
+            )
+
+            if normalisation:
+                warnings.warn(
```
ulricharmel, line 262: Why does do we need a different computation when normalisation is applied?

sash-a (approved):
Looks great, thanks @OmaymaMahjoub :fire:

ulricharmel (commented):
Thanks @OmaymaMahjoub. I just left minor questions. When you reply I will approve the PR

mava/components/updating/terminators.py
OmaymaMahjoub: The issue that `has` works only in the case of instance of a class however `calculate_absolute_metric` is an attribute so we can't use has in this case.

mava/environment_loop.py
OmaymaMahjoub: Yes unfortunately, from what we found that storing the params of the network related to norm is not the same as non normalized ones therefore we can't calculate the absolute metric or store  the best checkpointed params using the same util functions (that's why we create an issue for that https://github.com/instadeepai/Mava/issues/859 )

OmaymaMahjoub: I may create a PR for it before the end of the week!

ulricharmel (approved):
Thanks @OmaymaMahjoub. Nice work!

PR #844, opened 2022-11-23
Fix:  failing checks fix

### LINKED ISSUE #842: [BUG] Running tests using Pytest XDist and checkpoint_best_perf=True
### Describe the bug
Occasionally, the `tests/systems/ippo_system_evaluation_test.py::test_ippo` fails on github actions. E.g. https://github.com/instadeepai/Mava/actions/runs/3531385679/jobs/5924493286 . Error `worker 'gw0' crashed while running 'tests/systems/ippo_system_evaluation_test.py::test_ippo'`.

Pytest XDist runs tests in parallel and for some reason it is having an issue with the test above. I think it is because `ippo_system_multi_thread_eval` has these extra params:
```
     evaluation_interval={"executor_steps": 10},
     evaluation_duration={"evaluator_episodes": 2},
     checkpoint_best_perf=True,
```

I think these params are the problem, because `ippo_system_multi_thread` works correctly and is close to `ippo_system_multi_thread_eval`, except for the mentioned params. 

### To Reproduce
1. Create PR. 

### Expected behavior
Consistently run.  

### Context (Environment)
-

### Additional context
-

### Possible Solution
-

OmaymaMahjoub:
## What?
Fixing the integration test of ippo when we have checkpointer and evaluation interval
## Why?
Occasionally, the tests/systems/ippo_system_evaluation_test.py::test_ippo fails on github actions. E.g. https://github.com/instadeepai/Mava/actions/runs/3531385679/jobs/5924493286 . Error worker 'gw0' crashed while running 'tests/systems/ippo_system_evaluation_test.py::test_ippo'.
## How?
Disable all the nodes before running the system manually.
## Extra
Close #842


tests/systems/systems_test_data.py
```diff
@@ -308,11 +308,12 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
         use_next_extras=False,
         sample_batch_size=5,
         nodes_on_gpu=[],
-        is_test=True,
         checkpoint_minute_interval=3 / 60,
         trainer_parameter_update_period=1,
         evaluation_interval={"executor_steps": 10},
         evaluation_duration={"evaluator_episodes": 2},
         checkpoint_best_perf=True,
+        termination_condition={"executor_steps": 100},
```
RuanJohn, line 316: Perhaps we can extends this a bit to something like 30000? 100 seems quite little especially since we were previously doing 5 trainer steps, but I am not 100% how many executor steps would match that.

OmaymaMahjoub: I will give it a try rhe issue is that the coverage bot is complaining

OmaymaMahjoub: I changed the steps to be 5000 because if I change it to 30000 it may take very long time while it is just integration test

sash-a (changes requested):
I think we need to keep the trainer_run.step() method in as otherwise aren't really doing a proper integration test. Are you sure you can't keep that in, but also keep the more infrequent trainer intervals?

tests/systems/ippo_system_evaluation_test.py
```diff
@@ -33,14 +33,5 @@ def test_ippo(
...
-        "trainer"
-    ]
-
-    trainer_node.disable_run()
 
     test_ippo_system_mt.launch()
-    trainer_run = trainer_node.create_handle().dereference()
-
-    for _ in range(5):
-        trainer_run.step()
```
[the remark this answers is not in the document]
OmaymaMahjoub: the step method here is for running the node and not the step function fromthe trainer but we can say it is quite similar however, the issue is that this causing the checks to fail because we didn't disable the other nodes (executor, evaluator...)

RuanJohn (dismissed):
Ran the test locally and it passes for me. Thank you @OmaymaMahjoub :raised_hands:

RuanJohn (approved):
Thank you :raised_hands:

sash-a (approved):
Thanks @OmaymaMahjoub :fire:

PR #846, opened 2022-11-24
Bugfix: Data shuffling

DriesSmit:
## What?
Fix a bug where we shuffle over both the batch and sequence dimensions in the trainer. This breaks the bootstrapped value estimation calculation and also limits the recurrent policy. 
## Why?
We should only shuffle over the batch dimension.
## How?
Update the trainer code.
## Extra
.


mava/components/training/model_updating.py
```diff
@@ -295,24 +294,23 @@ def model_update_epoch(
...
-            )
+            permutation = jax.random.permutation(shuffle_key, batch_shape)
 
             shuffled_batch = jax.tree_util.tree_map(
                 lambda x: jnp.take(x, permutation, axis=0), batch
             )
+
+            # Flatten the batch.
+            shuffled_batch = jax.tree_util.tree_map(
+                lambda x: x.reshape((-1,) + x.shape[2:]), shuffled_batch
```
sash-a, line 311: Haiku have a great method for this: `merge_leading_dims`

It's in haiku._src I think

mava/components/training/model_updating.py
```diff
@@ -295,24 +294,23 @@ def model_update_epoch(
...
-            # assert ...
-            assert (
-                list(batch.observations.values())[0].observation.shape[0]
-                == trainer.store.full_batch_size
-            )
+            # Note (dries): This computation is only performed once at compilation tome
+            # and not at every epoch step.
+            batch_shape = jax.tree_util.tree_leaves(
+                list(batch.observations.values())[0].observation
+            )[0].shape[0]
```
sash-a, line 301: Can leave it here, but might be cleaner to store the batch size in the store as we used to do with full_batch_size

sash-a (changes requested):
Great find Dries, just minor code style things! :fire:

mava/components/training/model_updating.py
DriesSmit: Awesome thanks :raised_hands: I was searching for that method.

mava/components/training/model_updating.py
DriesSmit: Agreed. I removed it as it is the same size as sample_batch_size :+1:

mava/components/training/losses.py
```diff
@@ -247,8 +248,9 @@ def policy_loss_fn(
                         )
 
                         # Flatten the distribution_params
+
                         distribution_params = jax.tree_util.tree_map(
-                            lambda x: x.reshape((-1,) + x.shape[2:]),
+                            lambda x: merge_leading_dims(x, 2),
```
RuanJohn, line 253: I really like this!

DriesSmit: I was @sash-a 's idea :raised_hands:

sash-a (approved):
Looks great! Hopefully the benchmarks are good! :fire:

RuanJohn (approved):
Thank you @DriesSmit! And a very good catch! 
Approved, just pending the benchmark results.

PR #847, opened 2022-11-24
fix: parameter client segmentation fault

sash-a:
## What?
Switched to using launchpad threads instead of python threads
## Why?
Python threads were causing segmentation faults when the trainer updated too fast. Not entirely sure why, hypothesis is that it was creating too many threads.
## How?
Removed the `ThreadPoolExecutor` inside the parameter client and called the methods with launchpad futures.

So`self._client.futures.some_method` instead of `executor.submit(self._client.some_method)`


mava/systems/parameter_client.py
```diff
@@ -110,6 +102,22 @@ def _adjust_and_request(self) -> None:
...
+        """
+        self._client.futures.set_parameters(  # type: ignore
+            {key: self._parameters[key] for key in self._set_keys},
+        )
+        # not sure how these futures will interact
+        # will it fetch the parameters set above?
+        get_future = self._client.futures.get_parameters(self._get_keys)  # type: ignore
+        get_future.add_done_callback(lambda ctx: self._copy(ctx.result()))
+
+        return get_future
```
sash-a, line 119: Should I return the get and set future?

mava/systems/parameter_client.py
```diff
@@ -110,6 +102,22 @@ def _adjust_and_request(self) -> None:
...
+        Returns:
+            A future for the parameter get
+        """
+        self._client.futures.set_parameters(  # type: ignore
+            {key: self._parameters[key] for key in self._set_keys},
+        )
+        # not sure how these futures will interact
+        # will it fetch the parameters set above?
+        get_future = self._client.futures.get_parameters(self._get_keys)  # type: ignore
+        get_future.add_done_callback(lambda ctx: self._copy(ctx.result()))
```
sash-a, line 117: This could cause a race condition so I'm not sure what to do here? 

A possibility is to only copy the stuff in get keys, because if we are setting we should have the most recent params relating to set keys right?

Any suggestions welcome :smile:

mava/systems/parameter_client.py
```diff
@@ -110,12 +111,37 @@ def _adjust_and_request(self) -> None:
...
+
     def get_async(self) -> None:
         """Asynchronously updates the parameters with the latest copy from server.
 
         Returns:
             None.
         """
+        # Do not do parallel calls if not multiprocess
+        if not self._client.store.global_config.multi_process:
+            return self.get_and_wait()
```
sash-a, line 144: Thoughts on this and the similar ones below? Previously we were using a thread in single process mode, I think this is more correct, but also we can't use a thread in single process mode anymore because the parameter server (`_client`) isn't a launchpad node

DriesSmit: We can maybe just set the `_async_request=get_and_wait` if not self._client.store.global_config.multi_process? This still allows us to not update at every call as is done with `get_and_wait`. What do you think?

DriesSmit: It seems that you are already do something like this?

mava/systems/parameter_client.py
sash-a: Fixed potential race condition by only getting parameters not in set_keys

mava/systems/parameter_client.py
sash-a: As discussed offline added an `DoneFuture` and now this is what is done

mava/systems/parameter_client.py
sash-a: Changed this to return both and now check that both are complete

mava/utils/done_future.py
```diff
@@ -0,0 +1,41 @@
+from concurrent import futures
+from typing import Any, Callable, Generic, Optional, TypeVar
+
+T = TypeVar("T")  # type of result
+
+
+class DoneFuture(futures.Future, Generic[T]):
```
DriesSmit, line 7: This looks really nice :raised_hands: Thanks @sash-a :slightly_smiling_face:

DriesSmit (approved):
Thanks, @sash-a :raised_hands: This is a really important fix! :slightly_smiling_face:  And PPO is still training on debug with the lastest updates?

mava/utils/done_future.py
sash-a: Thanks Dries :smile:

ulricharmel (approved):
Thanks @sash-a. Nice implementation. Nice idea renaming client to server.

PR #850, opened 2022-11-25
Bugfix: Policy state mismatch

DriesSmit:
## What?
- Fix a bug in Mava's IPPO where the `network_keys` and `policy_states` are not aligned with the `observations`.
- Updated the adders to allow for both `extras` and `next_extras` to be passed at the same time.
## Why?
The `policy_states` are not aligned with the `observations` which could impact performance when using a recurrent policy.
## How?
Updated our PPO implementation and our adders.
## Extra
Requires the [data shuffling PR](https://github.com/instadeepai/Mava/pull/846) to be merged first.


DriesSmit:
I am still not sure what the best naming convention for `extras` and `next_extras` is. We can even call them `obs_aligned_extras` and `action_aligned_extras` to make it more user-friendly. Let me know what you think?

OmaymaMahjoub:
On hold, until we invesitgate the adder unit tests

PR #852, opened 2022-11-28
feat: observation and value norm components

sash-a:
## What?
Separated observation and value norm into their own components
## Why?
So that they are not tied to default components and can excluded from certain systems
## How?
Moved initialization code for them into separate components
## Extra:
Requires this PR https://github.com/instadeepai/Mava/pull/847 to be merged in first


ulricharmel (commented):
Thanks @sash-a. This really looks great. And now the normalisation is done in the components users can actually try their own normalisation approach. The PR looks good to me.

mava/components/executing/action_selection.py
```diff
@@ -96,7 +97,10 @@ def on_execution_select_actions(self, executor: SystemExecutor) -> None:
 
         observations = executor.store.observations
         # Normalize the observations before selecting actions.
-        if executor.store.global_config.normalize_observations:
+        if (
+            executor.has(ObservationNormalisation)
```
DriesSmit, line 101: Should we maybe create a BaseObservationNormalsiation component that we can extend from? Therefore if there is new normalisation implementations this if statement would still pass?

mava/components/training/step.py
```diff
@@ -284,7 +289,10 @@ def get_behavior_values(
 
             # Denormalise the values here to keep the GAE function clean
             target_value_stats = states.target_value_stats
-            if trainer.store.global_config.normalize_target_values:
+            if (
+                trainer.has(ValueNormalisation)
```
DriesSmit, line 293: Same with ValueNormalisation?

DriesSmit (commented):
This PR looks really good :fire: Thanks @sash-a :slightly_smiling_face: It is really nice to see that you are making IPPO more composable. Just see my minor comments.

ulricharmel (approved):
Thanks @sash-a! Nice work.

DriesSmit (approved):
Thanks @sash-a :raised_hands:

Issue #845, opened 2022-11-24
[FEATURE] ability check if a system has a component


sash-a:
### Feature
Allow components to check if the system has a component, this will avoid most, if not all of the nasty `hasattr` calls. 

A method like: `builder.has_component(ObservationNormalisation)` would allow one to avoid checking `hasattr(builder.store, "norm_params")`

### Why
* `hasattr` is very _un-pythonic_
* If you change a variable name you have to find all the `hasattr`s checking for that variable and change those too!

### Proposal
Give components the or just system components (e.g builder, executor etc) the method `has_component` to check if the component is in the system.

### Definition of done
Core components have the `has_component` method

### Mandatory checklist before making a PR
* [ ] Core components have the `has_component` method

PR #853, opened 2022-11-29
feat: added `core_component.has` method

sash-a:
## What?
Added `core_component.has` method
## Why?
Useful as a better alternative to `hasattr`
## How?
Made base class for core component and implemented it there by checking if the given component type is in the callbacks list
## Extra
Closes: #845 
Required for: #852

### DIFF
--- a/mava/core_jax.py
+++ b/mava/core_jax.py
@@ -65,7 +65,7 @@ class BaseSystemComponent(abc.ABC):
 
     def __init__(self) -> None:
         """Core component init"""
-        self.store: SimpleNamespace
+        self.store: SimpleNamespace = SimpleNamespace()
         self.callbacks: List
 
     def has(self, component: Type, subtypes: bool = False) -> bool:
--- a/mava/core_jax.py
+++ b/mava/core_jax.py
@@ -68,25 +68,19 @@ class BaseSystemComponent(abc.ABC):
         self.store: SimpleNamespace = SimpleNamespace()
         self.callbacks: List
 
-    def has(self, component: Type, subtypes: bool = False) -> bool:
+    def has(self, component_type: Type) -> bool:
         """Checks if a component of the given type is in the system.
 
         Args:
             component: type of component for which to check.
-            subtypes: whether or not to check for subtypes of the given component.
 
         Returns:
             True if the system has a component of this type.
         """
-        if subtypes:  # checks for type of component and its subtypes
-            return any(
-                isinstance(other_component, component)
-                for other_component in self.callbacks
-            )
-        else:  # checks for exactly type of component
-            return any(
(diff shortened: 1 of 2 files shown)


mava/core_jax.py
```diff
@@ -60,18 +60,41 @@ def launch(self) -> None:
...
+        Args:
+            component: type of component for which to check.
+            subtypes: whether or not to check for subtypes of the given component.
 
-        # Simple namespace for assigning system builder attributes dynamically
-        self.store = SimpleNamespace()
+        Returns:
+            True if the system has a component of this type.
+        """
+        if subtypes:  # checks for type of component and its subtypes
```
ulricharmel, line 81: Are there instances where we must absolutely specify which test to use? If we don't have such instances then maybe we can try both approaches and remove the argument `subtypes`.

ulricharmel (commented):
Thanks @sash-a. Looks great. I left a minor comment.

mava/core_jax.py
sash-a: Was thinking the same as it's not the cleanest, but I think with some of the DQN stuff we want to be a bit stricter as there's quite a bit of inheritance there when we have DQN > recurrent DQN > VDN > QMIX. So we might want to be sure a specific component is in and no one of it's child classes.

So not sure if it's definitely necessary, but I think it might be in the future.

ulricharmel: Okay cool. I am happy with it then.

ulricharmel (approved):
Thanks @sash-a

DriesSmit (commented):
Thanks, @sash-a :raised_hands: This looks like a handy tool. Did you see any place in the IPPO code where we could use this function? Or is it only used in the DQN/VDN/QMIX implementations at the moment?

Update: I see it is used in the other PRs. I am happy to merge it :slightly_smiling_face:

DriesSmit (approved):

mava/core_jax.py
```diff
@@ -60,18 +60,41 @@ def launch(self) -> None:
...
+            return any(
```
DriesSmit, line 82: Feel free to ignore this. But can we use something like `check_fn = instance if subtypes else lambda x, y: type(x)==y` to simplify the code a bit?

DriesSmit (commented):
Just see my one comment.

mava/core_jax.py
sash-a: Thought about it a bit more and I think @ulricharmel is right, if we just have `isinstance` then we should be fine, you just need to pass the most specific component possible

DriesSmit (approved):
Thanks @sash-a :raised_hands:

ulricharmel (approved):
Thanks @sash-a

Issue #831, opened 2022-11-15
[MAINTAIN] on_building_executor_parameter_client too long!


OmaymaMahjoub:
### Please describe what needs to be maintained?
We need to break `on_building_executor_parameter_client` method (from `mava/components/building/parameter_client.py`) into smaller methods. This would also help when implementing other algorithms (such as value based ones) as those don't have a policy and critic network, so we could just override the smaller methods and leave most of the functionality untouched.

### How do we know when maintenance is complete?
`on_building_executor_parameter_client` method broken into different small methods

PR #854, opened 2022-11-29
Feature/best checkpoint component

sash-a:
## What?
Moved best checkpointer into it's own component, similar to what was done with value and observation normalisation.
## Why?
It was getting in the way of implementing new systems
## How?
Moved initialisation code into new components and put checks around code in env loop

Closes: #831


DriesSmit (dismissed):
Awesome work @sash-a :raised_hands:

sash-a:
Changing to WIP as @OmaymaMahjoub mentioned some changes could be made to make it work better with #841

mava/components/building/best_checkpointer.py
```diff
@@ -0,0 +1,74 @@
...
+            return
+
+        builder.store.best_checkpoint = self.init_checkpointing_params(builder)
+
+    def on_parameter_server_init(self, server: SystemParameterServer) -> None:
+        """Adding checkpointing parameters to parameter server"""
+        if not self.config.checkpoint_best_perf:
+            return
+
+        server.store.parameters.update(self.init_checkpointing_params(server))
```
OmaymaMahjoub, line 43: We need to make separate store variable for the best checkpoint else it will update the same vars related to the current training network

ulricharmel (approved):
Thanks @sash-a. Looks great to me.

mava/environment_loop.py
```diff
@@ -292,7 +292,7 @@ def step_executor() -> None:
                     results.update(eval_result)
                     self._logger.write(results)
 
-                    if self._executor.store.checkpoint_best_perf:
+                    if self._executor.store.global_config.checkpoint_best_perf:
```
ulricharmel, line 295: So if I understand the issue was that the previous if statement here with hasattr was never passing. Is there a default best check pointer component in the system? or Do we already have a check to make sure that whenever the best check pointer is set to true, a best check pointer component exist on the system.

ulricharmel (commented):
Thanks @sash-a and @OmaymaMahjoub. See my comment.

mava/environment_loop.py
OmaymaMahjoub: Oh nice catch @ulricharmel we missed the add of the other statement that check if the best checkpoint component exist or not.
In this if statement we need to check:
1. That the best checkpoint component exist
2. Check that even if the best checkpoint exist if the flag of checkpointing the best performance set to true

mava/components/building/best_checkpointer.py
```diff
@@ -0,0 +1,76 @@
...
+            return
+
+        server.store.parameters.update(
+            {"best_checkpoint": self.init_checkpointing_params(server)}
+        )
+
+    def init_checkpointing_params(
+        self, system: Union[SystemParameterServer, SystemBuilder]
+    ) -> Dict[str, Any]:
+        """Initialises the parameters used to checkpointing the best models"""
```
RuanJohn, line 50: Minor typo: ... used for checkpointing ...

mava/environment_loop.py
OmaymaMahjoub: Fixed by @sash-a

ulricharmel (approved):
Thanks @sash-a. Thanks @OmaymaMahjoub. Nice work

RuanJohn (approved):
Thank @sash-a and @OmaymaMahjoub :raised_hands:

PR #860, opened 2022-12-15
Recurrent IPPO critic

RuanJohn:
## What?
Added support for recurrent critics in IPPO. The system is working, learning and leads to performance increases. 
The trainer works by using the initial RNN hidden state for training instead of the network hidden state used during interacting with the environment. 

## Why?
Improve the improve IPPO system performance. 

## Extra
For now the batch size has to be passed in to `mava/systems/ippo/networks.py` for initialising the critic hidden states. 
The system has the most optimal performance when value clipping, orthogonal network initialisation and all normalisation except observation normalisation are turned off. Additionally an MSE loss should be used for the critic network for optimal performance.


DriesSmit:

PR #861, opened 2022-12-23
feat: make best checkpoint support norm params

### LINKED ISSUE #859: [MAINTAIN] Making the best checkpoint and the absolute metric support norm params
Currently the best checkpointer and the absolute metric features are working fine in the default case, however, they both don't support the case of the normalization of the params such as normalize observation and normalize the target value.

### Describe the outcome you'd like
Create the option of storing norm_params in the best checkpointer component

OmaymaMahjoub:
## What?
Make the bestcheckpoint component support the case of normalization
## Why?
Currently the best checkpointer and the absolute metric features are working fine in the default case, however, they both don't support the case of the normalization of the params such as normalize observation and normalize the target value.
## How?
Edit the stored network in the best_checkpoint params
## Extra
Close #859 
### To test this feature:
1. Run `examples/smac/feedforward/decentralised/calculate_absolute_metric.py` to check the logged json file
2. Run `examples/smac/feedforward/decentralised/best_checkpointed_net.py` to check that it restore the best network


sash-a (approved):
Great work @OmaymaMahjoub! This was a much needed fix! :fire:

KaleabTessera (approved):
Thansks for this @OmaymaMahjoub ! :raised_hands: 

Could you also please remove `normalize_target_values` and `normalize_observations` from all the examples and tests. I think it is still in a couple of places (you can do this in another PR if it is easer).

Issue #864, opened 2023-01-05
[BUG] Quickstart example fails in Colab


callumtilbury:
### Describe the bug
Problems when trying to run the `quickstart.ipynb` notebook

### To Reproduce
1. Visit <https://colab.research.google.com/github/instadeepai/Mava/blob/develop/examples/quickstart.ipynb>
2. Run the "Install required packages" cell
3. Hit error in installing `box-2d` (note: this install fails quietly, because the output is `%%capture`d)
4. As a result, `id-mava` isn't installed and the later cells won't run

### Expected behavior
The install should work without hiccup for any user trying the Colab notebook.

### Context (Environment)
 - OS: Google Colab – Release 2022/12/6

### Additional context
n/a

### Possible Solution
Common problem with box-2d: <https://stackoverflow.com/questions/54252800/python-cant-install-box2d-swig-exe-failed-with-error-code-1>, need to manually install `swig` first.

PR #865, opened 2023-01-06
fix: import of required packages in quickstart to run on Colab.

callumtilbury:
## What?
Updated the *Installing Packages* code block due to minor issues with Colab.

## Why?
Trying to run the Quickstart example on Colab faced two issues:
1. Sometimes when trying to install `box-2d`, a package called `swig` must first be installed—it seems to be a fairly common problem (for [example](https://stackoverflow.com/questions/54252800/python-cant-install-box2d-swig-exe-failed-with-error-code-1))
2. Intermittently, Colab also vaguely complained about numpy: 
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
RuntimeError: module compiled against API version 0xf but this version of numpy is 0xe
```

## How?
- `!apt-get install swig` installs swig before trying to install Mava, which allows `box-2d` to install successfully
- Seemingly by installing numpy again, the vague complaint disappears (note: the install command doesn't even reinstall numpy: `Requirement already satisfied: numpy in /usr/local/lib/python3.8/dist-packages (1.21.6)`, but it fixes the issue—could this be a Colab bug then?)

## Extra
I replaced `!pip` with the [best practice](https://discourse.jupyter.org/t/python-in-terminal-finds-module-jupyter-notebook-does-not/2262/9) magic command `%pip`. I also grabbed Mava from the repo in a single line, rather than what was previously done: deleting the extant Mava folder, cloning, etc. Not sure if there was a reason for that approach.

(Issue: #864)

### DIFF
--- a/examples/quickstart.ipynb
+++ b/examples/quickstart.ipynb
@@ -42,15 +42,20 @@
       "source": [
         "%%capture\n",
         "#@title Install required packages. (Run Cell)\n",
-        "! rm -r ./Mava\n",
-        "! git clone https://github.com/instadeepai/Mava.git\n",
-        "!pip install ./Mava[reverb,jax,launchpad,envs]\n",
+        "# Fix common bug in installing 'box-2d'\n",
+        "!apt-get install swig\n",
         "\n",
-        "# Installs for agent visualisation.\n",
-        "!pip install ./Mava[record_episode]\n",
-        "! apt-get update -y &&  apt-get install -y xvfb &&  apt-get install -y python-opengl && apt-get install ffmpeg && pip install pyvirtualdisplay \n",
-        "# Google colab has an old version of cloudpickle - issue with lp.  \n",
-        "!pip install cloudpickle -U"
+        "# Get Mava\n",
+        "%pip install \"id-mava[reverb,jax,envs] @ git+https://github.com/instadeepai/mava.git\"\n",
+        "\n",
+        "# Visualisation stuff\n",
+        "!apt-get update -y &&  apt-get install -y xvfb &&  apt-get install -y python-opengl && apt-get install ffmpeg && pip install pyvirtualdisplay \n",
+        "\n",
+        "# Colab has an old version of cloudpickle\n",
+        "%pip install cloudpickle -U\n",
+        "\n",
+        "# Colab has an old/conflicting numpy version\n",
+        "%pip install numpy"
       ]
     },
     {
--- a/examples/quickstart.ipynb
+++ b/examples/quickstart.ipynb
@@ -52,10 +52,15 @@
         "!apt-get update -y &&  apt-get install -y xvfb &&  apt-get install -y python-opengl && apt-get install ffmpeg && pip install pyvirtualdisplay \n",
         "\n",
         "# Colab has an old version of cloudpickle\n",
-        "%pip install cloudpickle -U\n",
-        "\n",
-        "# Colab has an old/conflicting numpy version\n",
-        "%pip install numpy"
+        "%pip install cloudpickle -U"
+      ]
+    },
+    {
+      "attachments": {},
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "At this point, it is necessary to restart the runtime. On the top menu, click `Runtime -> Restart runtime`. Alternatively, use the shortcut, `⌘/Ctrl + M .`"
       ]
     },
     {
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,7 @@ setup(
         "absl-py",
         "dm_env",
         "dm-tree",
-        "numpy",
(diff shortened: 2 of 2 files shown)


OmaymaMahjoub (dismissed):
Thanks @callumtilbury, it is working perfectly :fire:

examples/quickstart.ipynb
```diff
@@ -42,15 +42,20 @@
...
+        "%pip install \"id-mava[reverb,jax,envs] @ git+https://github.com/instadeepai/mava.git\"\n",
+        "\n",
+        "# Visualisation stuff\n",
+        "!apt-get update -y &&  apt-get install -y xvfb &&  apt-get install -y python-opengl && apt-get install ffmpeg && pip install pyvirtualdisplay \n",
+        "\n",
+        "# Colab has an old version of cloudpickle\n",
+        "%pip install cloudpickle -U\n",
+        "\n",
+        "# Colab has an old/conflicting numpy version\n",
+        "%pip install numpy"
```
KaleabTessera, line 58: So I did some digging on the `module compiled against API version 0xf but this version of numpy is 0xe` error and it seems like our jaxlib version is compiled for another version of numpy. (our jaxlib version is compiled for 1.22.x (https://github.com/numpy/numpy/blob/f8e87f8d5892dd5fe11ddd05470ae83a0104d417/numpy/core/setup_common.py#L46), while colab has np 1.21.x . 

Can you try changing the setup.py to set our numpy version (`numpy==1.22.4`) and then run this cell with `pip install numpy` removed?

We shouldn't ever see the `module compiled...` error.

KaleabTessera (changes requested):
Well done on your first PR @callumtilbury ! 

Just a comment on the np issue.

examples/quickstart.ipynb
callumtilbury: Thanks, @KaleabTessera!

Firstly—apologies for making the commit (https://github.com/instadeepai/Mava/pull/865/commits/9b75e1731c25b153489a00c33bdfd694b88861bd) in order to _test_ whether it works. This is bad practice, but I am not sure how else one can debug with changes _within_ Colab? I can't test locally and then push, because the problem seems to be with Colab itself. (I suppose I could try manually uploading the repo to Colab's workspace 🤔)

Anyway, I tried setting the numpy version as suggested, but a different version is loaded beforehand:
```
WARNING: The following packages were previously imported in this runtime:
  [dataclasses,enum,google,numpy]
You must restart the runtime in order to use newly installed versions.
```

Ignoring the warning, the old `numpy` is used—so we see the same `module compiled...` error. 

Of course, following the advice to restart the runtime fixes the issue, but its frustrating: forcing the user to restart their runtime to continue with the notebook. Doing it in code also feels hacky.

Thoughts?

OmaymaMahjoub:
Thanks @callumtilbury for the changes, there is a minor issue with choosing numpy to 1.22.0 that it doesn't work in 3.7 as they make this version support 3.8-3.10, Python 3.7 has been dropped. https://numpy.org/devdocs/release/1.22.0-notes.html 
@KaleabTessera shoud we change numpy version or drop the test related to 3.7?

examples/quickstart.ipynb
KaleabTessera: No worries about the extra commit! It is tricky debugging the quickstart. 

Agreed, doing it in code feels hacky. The user can restart manually. If you remove the `pip install numpy`, then I think this is ready to go.

KaleabTessera:
> Thanks @callumtilbury for the changes, there is a minor issue with choosing numpy to 1.22.0 that it doesn't work in 3.7 as they make this version support 3.8-3.10, Python 3.7 has been dropped. https://numpy.org/devdocs/release/1.22.0-notes.html @KaleabTessera shoud we change numpy version or drop the test related to 3.7?

I think we should drop 3.7. I am working on a PR for this.

examples/quickstart.ipynb
callumtilbury: Okay cool! One more problem: when the output is `%%capture`d, the prompt to restart the runtime doesn't show. But even when it shows, it's hidden halfway through a wall of text 🥴  I'll add an instruction to the user to restart the runtime manually, at least as an interim solution? 👍🏻

callumtilbury: Completed in https://github.com/instadeepai/Mava/pull/865/commits/813a4fe6cf18eb024445f1bbaffbc3f6711e4c6e :)

KaleabTessera: Great, thanks @callumtilbury !

KaleabTessera (approved):
Thanks @callumtilbury ! Well done on your first PR!

OmaymaMahjoub (approved):
Thanks @callumtilbury once we merge #872 this can be ready to go :raised_hands:

PR #868, opened 2023-01-07
bugfix: Normalisation components being initialised when disabled.

### LINKED ISSUE #867: [BUG] Normalisation components being initialised when disabled.
### Describe the bug
Even if normalization is disabled `on_building_init_end` will still be called in the normalization components. This creates an error if the datatype of the observation space is not an array as it attempts to set `obs_shape` based on the shape of the observation space.

### To Reproduce
1. Run `run_ippo.py` with normalization disabled using an environment that does not use an array as the observation space.

### Expected behavior
If normalization is disabled it should not be initialised.

siddarthsingh1:
## What?
Add check for normalisation to `on_building_init_end` for each normalisation component. 
## Why?
Currently if normalisation is disabled, the system attempts to use `on_building_init_end` which fails if the observation is not an array.
## How?
If statements added to the builder functions in the normalisation components.
## Extra
Closes #867 

Merge in  [Policy state mismatch](https://github.com/instadeepai/Mava/pull/850) first. This PR is tagged with XL because it is branched from that branch. The bugfix is 2 lines.


DriesSmit:

PR #872, opened 2023-01-10
Feature/upgrade docker and python

KaleabTessera:
## What?
- Remove p3.7 tests as this has been causing issues for a while now.
- Upgrade docker container. The previous container we used is now deprecated. 
## Why?
-
## How?
-
## Extra
- Still need to run some exps/sanity checks on the changes to the dockerfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ You can also install directly from source:
 pip install "id-mava[reverb,jax,envs] @ git+https://github.com/instadeepai/mava.git"
 ```
 
-We have tested `mava` on Python 3.7, 3.8 and 3.9. Note that because the installation of JAX differs depending on your hardware accelerator,
+We have tested `mava` on Python 3.8 and 3.9. Note that because the installation of JAX differs depending on your hardware accelerator,
 we advise users to explicitly install the correct JAX version (see the [official installation guide](https://github.com/google/jax#installation)). For more in-depth instalations guides including Docker builds and virtual environments, please see our [detailed installation guide](DETAILED_INSTALL.md).
 
 ## Quickstart ⚡
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        docker-image: ["python:3.7", "python:3.8", "python:3.9"]
+        docker-image: ["python:3.8", "python:3.9", "python:3.10"]
     steps:
       - name: Checkout mava
         uses: actions/checkout@v2
--- a/.github/workflows/ci_main.yaml
+++ b/.github/workflows/ci_main.yaml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        docker-image: ["python:3.7", "python:3.8", "python:3.9"]
+        docker-image: ["python:3.8", "python:3.9", "python:3.10"]
     steps:
       - name: Checkout mava
         uses: actions/checkout@v2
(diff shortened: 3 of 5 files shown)


.github/codecov.yml
```diff
@@ -2,12 +2,10 @@ coverage:
   status:
     project:
       default:
-        target: auto
         # Only fail PR, if overall project codecov drops by 0.5%.
         threshold: 0.5%
```
sash-a, line 6: I think to be lenient we should make between 2-5%. If the patch stays at 80% I think that means we should never loose too much coverage, but still be quite lenient

sash-a (changes requested):
Just the one comment on codecov, otherwise I think it looks good!

.github/codecov.yml
sash-a: Also are you sure omitting `target: auto` is correct? I did a test on the madqn branch and got it working with these settings:

```
  status:
    project:
      default:
        target: auto
        threshold: 5%
    patch:
      default:
        target: 80%
```

However (and importantly) I put the file in the root dir, not in the .github folder

KaleabTessera: > I think to be lenient we should make between 2-5%. If the patch stays at 80% I think that means we should never loose too much coverage, but still be quite lenient

Let's make it 1% for now and see if it causes issues.

KaleabTessera: > Also are you sure omitting `target: auto` is correct? I did a test on the madqn branch and got it working with these settings:
> 
> ```
>   status:
>     project:
>       default:
>         target: auto
>         threshold: 5%
>     patch:
>       default:
>         target: 80%
> ```
> 
> However (and importantly) I put the file in the root dir, not in the .github folder

So when I validated the config (https://docs.codecov.com/docs/codecov-yaml#validate-your-repository-yaml), it said `target: auto` was a incorrect parameter. This was causing codecov to not read these settings in the past, now it does. 

Maybe try validate your madqn config?

sash-a: just validated and my config is valid, [here's the one on dqn](https://github.com/instadeepai/Mava/pull/843/files#diff-2b2bda0c32d4aa698c35a605120c7745c3433d4ad4210996bfc7f32e9fe9f078) and the codecov is passing even though project cov is down 0.89% 

```
   project:
      default:
        target: auto
        threshold: 1%
    patch:
      default:
        target: 80%
```

README.md
```diff
@@ -72,7 +72,7 @@ You can also install directly from source:
 pip install "id-mava[reverb,jax,envs] @ git+https://github.com/instadeepai/mava.git"
 ```
 
-We have tested `mava` on Python 3.7, 3.8 and 3.9. Note that because the installation of JAX differs depending on your hardware accelerator,
+We have tested `mava` on Python 3.8 and 3.9. Note that because the installation of JAX differs depending on your hardware accelerator,
```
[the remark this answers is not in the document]
KaleabTessera: Updated, thanks! :)

KaleabTessera:
> Thanks @KaleabTessera, just minor question is this [block](https://github.com/instadeepai/Mava/blob/develop/README.md?plain=1#L12) will be updated automatically when we merge this PR, besides the python versions in this [link](https://pypi.org/project/id-mava/)?

No. So those versions only get updated once we push a new mava version to pypi.

OmaymaMahjoub:
> > Thanks @KaleabTessera, just minor question is this [block](https://github.com/instadeepai/Mava/blob/develop/README.md?plain=1#L12) will be updated automatically when we merge this PR, besides the python versions in this [link](https://pypi.org/project/id-mava/)?
> 
> No. So those versions only get updated once we push a new mava version to pypi.

Oh I see, then I think the PR ready to go :rocket:

OmaymaMahjoub (approved):

sash-a (approved):
Nice, thanks @KaleabTessera!

OmaymaMahjoub (approved):
Thanks @KaleabTessera

PR #886, opened 2023-01-16
Feature/single file structure

KaleabTessera:
## What?
- Added basic structure for single file implementations. 
- Added requirements and dockerfile.  
## Why?
-
## How?
-
## Extra
Closes https://github.com/instadeepai/Mava/issues/885 .


sash-a (approved):
Thanks for this @KaleabTessera, looks great! 

Just one thing I would change is the filename ddpg to iddpg, just so it's not confused with a single agent version and to keep it consistent with ippo and idqn, but we can do this in our implementation :smile:

KaleabTessera:
> Thanks for this @KaleabTessera, looks great!
> 
> Just one thing I would change is the filename ddpg to iddpg, just so it's not confused with a single agent version and to keep it consistent with ippo and idqn, but we can do this in our implementation smile

Thanks for the review @sash-a ! I am happy for the people doing the implementation to rename it. I just haven't seen `iddpg` anywhere, so maybe when implementing, people can research the name.

siddarthsingh1 (changes requested):
Thanks @KaleabTessera , looking forward to working on this. Just a quick thing though. I don't think we need both idqn and maqdn. They usually refer to the same algorithm in literature as far as I'm aware.

KaleabTessera:
> Thanks @KaleabTessera , looking forward to working on this. Just a quick thing though. I don't think we need both idqn and maqdn. They usually refer to the same algorithm in literature as far as I'm aware.

So madqn, I was thinking DQN with joint action value function across all agents, but that is not quite common or part of ctde, so I changed it to rdqn (recurrent dqn).

siddarthsingh1 (dismissed):
@KaleabTessera , thanks for the update.

educational/systems/random_agent.py
```diff
@@ -0,0 +1,370 @@
...
+            random_seed=config.seed,
+        )
+    elif config.type == "smac":
+        env, _ = smac_utils.make_environment(
+            map_name=config.env_name, random_seed=config.seed
+        )
+    return env, config
+
+
+def make_system(
```
sash-a, line 127: I still think this is too much boilerplate, but I'll approve here and we can deal with it in #887

sash-a (approved):
LGTM other than the random agent boilerplate, but would rather deal with that in another PR

educational/systems/random_agent.py
KaleabTessera: I accidentally added this here. I have removed it, it should just be in the random agent PR.

sash-a (approved):
Thanks for removing the random agent stuff, looks great!

siddarthsingh1 (approved):
Looks great! @KaleabTessera.

PR #887, opened 2023-01-18
Feature/random single file agent

KaleabTessera:
## What?
- Single file random agent that works on debug and smac. 
## Why?
- 
## How?
- 
## Extra
- Closes https://github.com/instadeepai/Mava/issues/881.


educational/systems/random_agent.py
```diff
@@ -0,0 +1,370 @@
...
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string("system", "test agent", "What agent is running.")
+flags.DEFINE_string(
+    "base_dir", "~/mava", "Base dir to store experiment data e.g. checkpoints."
+)
+flags.DEFINE_string(
+    "env_name",
+    "simple_spread",
+    "Env name e.g. sinple_spread, 3m or 8m.",
```
siddarthsingh1, line 39: Thanks @KaleabTessera! All looks good. `sinple_spread` just needs to be corrected to `simple_spread`

siddarthsingh1 (changes requested):

educational/systems/random_agent.py
```diff
@@ -0,0 +1,370 @@
...
+        config : system config.
+
+    Returns:
+        system.
+    """
+    init_rng = jax.random.PRNGKey(config.seed)
+    agent_specs = environment_spec.get_agent_environment_specs()
+
+    @dataclass
+    class RandomExecutor:
```
sash-a, line 143: Ok after starting implementation I think this is too much boilerplate for single file systems, I think they should just be pretty messy to start with and once they are all implemented we can decide on a common interface.

Especially the observe and observe first, since a simple replay buffer will just have a single add, not necessarily an add first.

I think stand alone global methods, like `get_actions(obs, actor, params)` and `learn(batch, actor, actor_params, critic, ...)` would be a better approach, just to put as few constraints on design as possible and then we can refactor all implementations to a common interface later, once we've seen all implementations and can take the best parts from the ones we like.

What do you think?

sash-a (changes requested):

educational/systems/random_agent.py
```diff
@@ -0,0 +1,370 @@
...
+        """Func which checks if we should stop running.
+
+        Args:
+            episode_count : max episodes.
+            step_count : max steps.
+
+        Returns:
+            bool indicating if we should terminate.
+        """
+        should_stop = (num_episodes is not None and episode_count >= num_episodes) or (  # type: ignore # noqa: E501
```
sash-a, line 300: `num_episodes` -> `max_num_episodes` for consistency

educational/systems/random_agent.py
```diff
@@ -0,0 +1,370 @@
...
+        def init(state) -> None:
+            """Init executor.
+
+            Args:
+                state : state that holds variables.
+            """
+            pass
+
+        @jax.jit
+        def select_actions(
```
sash-a, line 155: I still have an issue with how much boilerplate this all is for a single file system, when all you want is a select actions function

sash-a (commented):
Still have an issue with how much boilerplate there is just for a select actions function. Not sure what the best solution to this is though?

educational/systems/random_agent.py
ulricharmel: I agree with @sash-a too much boilerplate code. I think we really want to make these single file implementations as simple as possible.

DriesSmit:

PR #888, opened 2023-01-19
feat: Add s3 support to tensorboard logging.

KaleabTessera:
## What?
- Support tensorboard logs to s3 url.
## Why?
- So we use s3 urls for tb logs. 
## How?
- Don't parse s3 urls and add `import tensorflow_io`. 
## Extra
-

### DIFF
--- a/mava/utils/loggers/base.py
+++ b/mava/utils/loggers/base.py
@@ -20,6 +20,12 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, Optional, Union
 
+try:
+    # https://github.com/tensorflow/tensorboard/issues/5480#issuecomment-1016612630
+    import tensorflow_io  # noqa: F401
+except ModuleNotFoundError:
+    pass
+
 from acme.utils import loggers, paths
 from acme.utils.loggers import base
 
@@ -61,8 +67,12 @@ class Logger(MavaLogger):
         """Initialise logger."""
         self._label = label
 
-        if not isinstance(directory, Path):
-            directory = Path(directory)
+        is_s3_url = isinstance(directory, str) and "s3" in directory
+        # Don't modify path if s3 url, use url string.
+        # Note this requires tensorflow_io to be installed.
+        if not is_s3_url:
+            if not isinstance(directory, Path):
+                directory = Path(directory)
 
         self._directory = directory
         self._time_stamp = time_stamp if time_stamp else str(datetime.now())
@@ -149,7 +159,10 @@ class Logger(MavaLogger):
                 json_path = extra_logger_kwargs["json_path"]
 
             else:
-                json_path = str(self._directory / self._time_stamp)
+                if isinstance(self._directory, Path):
+                    json_path = str(self._directory / self._time_stamp)
+                else:
+                    json_path = f"{self._directory}/{self._time_stamp}"
 
             del extra_logger_kwargs["json_path"]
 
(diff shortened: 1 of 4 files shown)


RuanJohn (approved):
Thanks @KaleabTessera. Looks good and the tb logs are visible on AIchor now. Massive win! :partying_face:

DriesSmit (dismissed):
Thanks @KaleabTessera :raised_hands: Really nice work :slightly_smiling_face:

mava/utils/loggers/base.py
```diff
@@ -173,13 +186,20 @@ def make_logger(
...
+                path = f"{self._directory}/{self._time_stamp}/{subdir}/{ self._label}"
+            else:
+                path = f"{self._directory}/{self._time_stamp}/{self._label}"
+            return path
         else:
-            path = str(self._directory / self._time_stamp / self._label)
+            if subdir:
+                path = str(self._directory / self._time_stamp / subdir / self._label)
+            else:
+                path = str(self._directory / self._time_stamp / self._label)
```
sash-a, line 199: not following what's happening here, if it's not a string what is it and what does the divide operator do?

mava/utils/loggers/base.py
```diff
@@ -173,13 +186,20 @@ def make_logger(
         return logger
 
     def _path(self, subdir: Optional[str] = None) -> str:
-        if subdir:
-            path = str(self._directory / self._time_stamp / subdir / self._label)
+        if isinstance(self._directory, str):
+            if subdir:
+                path = f"{self._directory}/{self._time_stamp}/{subdir}/{ self._label}"
+            else:
+                path = f"{self._directory}/{self._time_stamp}/{self._label}"
```
sash-a, line 193: Maybe would be cleaner to make a list of this stuff, ignore the empty/None items and then `"/".join(path)`

sash-a (changes requested):
Minor code style stuff but thanks for this major QOL improvement

mava/utils/loggers/base.py
```diff
@@ -61,8 +67,12 @@ def __init__(
         """Initialise logger."""
         self._label = label
 
-        if not isinstance(directory, Path):
-            directory = Path(directory)
+        is_s3_url = isinstance(directory, str) and "s3" in directory
+        # Don't modify path if s3 url, use url string.
+        # Note this requires tensorflow_io to be installed.
+        if not is_s3_url:
+            if not isinstance(directory, Path):
```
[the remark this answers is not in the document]
KaleabTessera: Done.

mava/utils/loggers/base.py
KaleabTessera: So the divider does a similiar thing to the string case. You just can't do `self._directory / ...` with strings (without f strings). Here for a path, you can do this without converting the path object to a string and a couple lines below we do some path cleanup that ensures it is a valid path (`paths.process_path`).

mava/utils/loggers/base.py
KaleabTessera: I agree that would be cleaner, but I prefer the readability of how it looks now. Do you think the cleaner implementation is worth the potential readability hit?

sash-a: Ye I thought it might be the same. So can leave as is

mava/utils/loggers/base.py
sash-a: Ok nice

sash-a (approved):
LGTM @KaleabTessera :fire:

RuanJohn (approved):
Thanks @KaleabTessera :raised_hands:

PR #894, opened 2023-01-27
Feature: masking padded sequence

ulricharmel:
## What?

- Implement the padded sequence masking on top of the develop branch so that this can easily be merged with other PRs.

## Why?
- Masking padded sequences improves results.
## How?
- Added code in step.py to create a mask from the discounts and apply this mask to all training data variables.
## Extra
[Benchmarks](https://www.notion.so/instadeep/Sequence-padding-and-large-batch-sizes-f70ce4feb33340c3939b5df6114a26e6)


mava/components/training/losses.py
```diff
@@ -19,22 +19,68 @@
...
+            update.
+    """
+    chex.assert_rank([prob_ratios_t, adv_t], [1, 1])
+    chex.assert_type([prob_ratios_t, adv_t], [float, float])
+
+    adv_t = jax.lax.select(use_stop_gradient, jax.lax.stop_gradient(adv_t), adv_t)
+    clipped_ratios_t = jnp.clip(prob_ratios_t, 1.0 - epsilon, 1.0 + epsilon)
+    clipped_objective = jnp.fmin(prob_ratios_t * adv_t, clipped_ratios_t * adv_t)
+
+    return -jnp.sum(clipped_objective * mask) / jnp.sum(mask)
```
OmaymaMahjoub, line 77: This could lead to an error that happen to us priorly, at some cases you may divide by zero `jnp.sum(mask)=0`(I know it may be rare but at least add a small factor next to it)

mava/components/training/losses.py
```diff
@@ -259,20 +313,20 @@ def policy_loss_fn(
                             policy_params, observations
                         )
 
-                    log_probs = network.log_prob(distribution_params, actions)
-                    entropy = network.entropy(distribution_params)
+                    log_probs = network.log_prob(distribution_params, actions) * mask
+                    entropy = network.entropy(distribution_params) * mask
```
OmaymaMahjoub, line 317: If you can make the masking optional so in case we don't need it we set the flag to false

mava/components/training/losses.py
```diff
@@ -357,11 +416,11 @@ def critic_loss_fn(
...
                             clipped_value_error
                         )
-                        value_loss = jnp.mean(
+                        value_loss = jnp.sum(
                             jnp.fmax(unclipped_value_loss, clipped_value_loss)
-                        )
+                        ) / jnp.sum(mask)
                     else:
-                        value_loss = jnp.mean(unclipped_value_loss)
+                        value_loss = jnp.sum(unclipped_value_loss) / jnp.sum(mask)
```
OmaymaMahjoub, line 423: same as above dividing by zero in rare cases

mava/components/training/losses.py
```diff
@@ -357,11 +416,11 @@ def critic_loss_fn(
                         clipped_value_loss = trainer.store.value_loss_fn(
                             clipped_value_error
                         )
-                        value_loss = jnp.mean(
+                        value_loss = jnp.sum(
                             jnp.fmax(unclipped_value_loss, clipped_value_loss)
-                        )
+                        ) / jnp.sum(mask)
```
OmaymaMahjoub, line 421: same as above dividing by zero in rare cases

mava/components/training/losses.py
```diff
@@ -259,20 +313,20 @@ def policy_loss_fn(
...
 
-                    policy_loss = rlax.clipped_surrogate_pg_loss(
-                        rhos, advantages, clipping_epsilon
+                    policy_loss = clipped_surrogate_pg_loss(
+                        rhos, advantages, clipping_epsilon, mask
                     )
 
                     # Entropy regulariser.
-                    entropy_loss = -jnp.mean(entropy)
+                    entropy_loss = -jnp.sum(entropy) / jnp.sum(mask)
```
OmaymaMahjoub, line 329: same as above dividing by zero in rare cases

OmaymaMahjoub (changes requested):
Thanks @ulricharmel, just few changes to edit

mava/components/training/losses.py
ulricharmel: Yep. I just pushed some updates for with and also added a flag for the sequence masking.

OmaymaMahjoub (dismissed):
Thanks @ulricharmel, I think it is ready to go :fire:

mava/components/training/losses.py
OmaymaMahjoub: Cool thanks :slightly_smiling_face:

examples/smac/feedforward/decentralised/run_ippo_smac_death_masking.py
```diff
@@ -73,15 +74,16 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
...
     )
 
     critic_optimiser = optax.chain(
-        optax.clip_by_global_norm(40.0), optax.scale_by_adam(), optax.scale(-1e-4)
+        optax.clip_by_global_norm(40.0), optax.scale_by_adam(), optax.scale(-5e-4)
     )
 
     # Create the system.
     system = ippo.IPPOSystem()
+    system.update(HuberValueLoss)
```
DriesSmit, line 86: Do we need Huber loss when testing the death masking code? Dit is not work with vanilla IPPO?

examples/smac/feedforward/decentralised/run_ippo_smac_death_masking.py
```diff
@@ -92,10 +94,21 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
         policy_optimiser=policy_optimiser,
         critic_optimiser=critic_optimiser,
         run_evaluator=True,
-        epoch_batch_size=5,
+        epoch_batch_size=32,
         num_epochs=15,
         num_executors=1,
-        multi_process=True,
+        multi_process=False,
```
DriesSmit, line 100: Why did we disable multi-process?

examples/smac/feedforward/decentralised/run_ippo_smac_death_masking.py
```diff
@@ -92,10 +94,21 @@ def network_factory(*args: Any, **kwargs: Any) -> Any:
...
+        sequence_length=121,
+        period=121,
```
DriesSmit, line 102: Did we decide that this is the only fix to our problem? There was no solution that still allows for smaller sequences. I think the code uses smaller sequences, but the full sequence is used to calculate GAE? Or do they also train on the full sequence?

mava/components/training/losses.py
```diff
@@ -259,20 +315,21 @@ def policy_loss_fn(
                             policy_params, observations
                         )
 
-                    log_probs = network.log_prob(distribution_params, actions)
-                    entropy = network.entropy(distribution_params)
+                    log_probs = network.log_prob(distribution_params, actions) * mask
```
DriesSmit, line 318: Is it necessary to mask the log_probs f we mask the clipped_surrogate_pg_loss? Maybe it is necessary. Just confirm this.

mava/components/training/losses.py
```diff
@@ -332,10 +391,13 @@ def critic_loss_fn(
                     observations: Any,
                     target_values: jnp.ndarray,
                     behavior_values: jnp.ndarray,
+                    mask: jnp.ndarray,
                 ) -> Tuple[jnp.ndarray, Dict[str, jnp.ndarray]]:
                     """Inner critic loss function: see outer function for parameters."""
 
-                    values = network.critic_network.apply(critic_params, observations)
+                    values = (
+                        network.critic_network.apply(critic_params, observations) * mask
```
DriesSmit, line 399: Again is it necessary to mask here if we mask in the loss function?

DriesSmit (commented):
Thanks @ulricharmel :fire: This work looks really nice. I am glad the masking is working and that we are finally getting some results on SMAC :slightly_smiling_face:

mava/components/training/losses.py
ulricharmel: Great catch. This is not necessary because all the operations that follow are element wise. So we just need to mask right at the end. Same as before

examples/smac/feedforward/decentralised/run_ippo_smac_death_masking.py
ulricharmel: We probably need to run an experiment to check that but in that case we will be computing truncated GAE that do not benefit from the full episode advantages.  In the onpolicy implementation they do the truncation in shorter sequences for the recurrent network after computing the GAE and the returns.

examples/smac/feedforward/decentralised/run_ippo_smac_death_masking.py
ulricharmel: I generally just disable this when I am running experiments locally. I will set it back to True

mava/components/training/losses.py
ulricharmel: Not necessary

examples/smac/feedforward/decentralised/run_ippo_smac_death_masking.py
DriesSmit: Agreed. Maybe in the future we can figure out a way to still use shorter sequences without affecting the GAE.

DriesSmit (approved):
Thanks @ulricharmel :raised_hands: The PR looks good to me :+1:

OmaymaMahjoub (approved):
Thanks @ulricharmel :fire:

PR #903, opened 2023-07-11
Feat: Logger and config.yaml

OmaymaMahjoub:
## What?
Add logging util functions.
## Why?
By adding this logger component we can now have the results in json files and tensorboard
## How?
Added Logger_tools.py
## Extra
This branch is based on `feat/tpu_setup` we need to close #902 before reviewing this.


OmaymaMahjoub:
This PR will be closed as #906 addresses the same issue and is adaptable to the current ippo implementation.

PR #905, opened 2023-07-28
feat: Feedforward IPPO JAX Agent implemented

RuanJohn:
## What?
First fully end to end JAX IPPO Agent that is able to train on Rware tiny-4ag. 

## Note
For now we are not fully overhauling the repo yet but working inside the `jax_distribution`. 
Since agents are homogeneous we vmap over them. 

This is joint work done with @OmaymaMahjoub and @DriesSmit.


jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+        """Update the network."""
+
+        def _env_step(runner_state: Tuple, unused: Any) -> Tuple:
+            """Step the environment."""
+            params, opt_state, rng, env_state, last_timestep = runner_state
+
+            # SELECT ACTION
+            rng, _rng = jax.random.split(rng)
+            actor_policy, value = apply_fn(params, last_timestep.observation)
+            action = actor_policy.sample(seed=_rng)
```
sash-a, line 105: Name the keys so we know what they're used for

```suggestion
            rng, policy_rng = jax.random.split(rng)
            actor_policy, value = apply_fn(params, last_timestep.observation)
            action = actor_policy.sample(seed=policy_rng)
```

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+            # LOG EPISODE METRICS
+            done, reward = jax.tree_util.tree_map(
+                lambda x: jnp.repeat(x, config["NUM_AGENTS"]).reshape(
+                    config["NUM_ENVS"], -1
+                ),
+                (timestep.last(), timestep.reward),
+            )
+            info = {
+                "returned_episode_returns": env_state.returned_episode_returns,
+                "returned_episode_lengths": env_state.returned_episode_lengths,
```
sash-a, line 120: Unclear what this is from the name, can you clarify what this is so we can come up with a better name

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+    def __call__(
+        self, observation: Observation
+    ) -> Tuple[distrax.Categorical, chex.Array]:
+        """Forward pass."""
+        x = observation.agents_view
+
+        if self.activation == "relu":
+            activation = nn.relu
+        else:
+            activation = nn.tanh
```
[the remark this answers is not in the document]
sash-a: Maybe this is a bit overkill for this PR, if so just add a todo above that this should be done outside of the neural network class

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+                total_loss, grads = grad_fn(
+                    params, opt_state, traj_batch, advantages, targets
+                )
+
+                grads, total_loss = jax.lax.pmean(
+                    (grads, total_loss), axis_name="batch"
+                )
+                grads, total_loss = jax.lax.pmean(
+                    (grads, total_loss), axis_name="device"
+                )
```
sash-a, line 231: Is `pmean`ing twice here correct, if you have a device axis then don't you want to just mean over that axis?

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+                new_params = optax.apply_updates(params, updates)
+
+                return (new_params, new_opt_state), total_loss
+
+            params, opt_state, traj_batch, advantages, targets, rng = update_state
+            rng, _rng = jax.random.split(rng)
+
+            # SHUFFLE MINIBATCHES
+            batch_size = config["ROLLOUT_LENGTH"] * config["NUM_ENVS"]
+            permutation = jax.random.permutation(_rng, batch_size)
```
sash-a, line 243: Name the rng keys!

```suggestion
            rng, shuffle_rng = jax.random.split(rng)

            # SHUFFLE MINIBATCHES
            batch_size = config["ROLLOUT_LENGTH"] * config["NUM_ENVS"]
            permutation = jax.random.permutation(shuffle_rng, batch_size)
```

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+            batch = (traj_batch, advantages, targets)
+            batch = jax.tree_util.tree_map(
+                lambda x: x.reshape((batch_size,) + x.shape[2:]), batch
+            )
```
sash-a, line 247: This is done quite a lot in general, I know [haiku has](https://github.com/deepmind/dm-haiku/blob/main/haiku/_src/basic.py#L207) `merge_leading_dims` I couldn't find it for flax. If it definitely doesn't have this then we should add our own implementation in a utils file so that we can do:

```python
batch = jax.tree_util.tree_map(lambda x: merge_leading_dims(x, 2), batch)
```

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+        return runner_state, metric
+
+    def learner_fn(
+        params: FrozenDict,
+        opt_state: OptState,
+        rng: chex.PRNGKey,
+        env_state: State,
+        timesteps: TimeStep,
+    ) -> dict:
+        """Learner function."""
```
sash-a, line 285: This should be more descriptive as it's the main function we return

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+    env: jumanji.Environment, apply_fn: Callable, update_fn: Callable, config: dict
+) -> Callable:
+    """Get the learner function."""
+
+    def _update_step(runner_state: Tuple, unused_target: Any) -> Tuple:
+        """Update the network."""
+
+        def _env_step(runner_state: Tuple, unused: Any) -> Tuple:
+            """Step the environment."""
+            params, opt_state, rng, env_state, last_timestep = runner_state
```
sash-a, line 100: We use this runner state in quite a few places, it would be nice to make a type for this in `types.py`, and then use it in this file instead of just `Tuple`. I think the following is correct

```python
RunnerState = Tuple[FrozenDict, OptState, chex.PRNGKey, State (jumanji's state?), TimeStep]
```

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+
+    reshape_states = lambda x: x.reshape(
+        (cores_count, config["UPDATE_BATCH_SIZE"], config["NUM_ENVS"]) + x.shape[1:]
+    )
+    env_states = jax.tree_util.tree_map(
+        reshape_states, env_states
+    )  # add dimension to pmap over.
+    timesteps = jax.tree_util.tree_map(
+        reshape_states, timesteps
+    )  # add dimension to pmap over.
```
sash-a, line 370: nitpick

```suggestion
    # add dimension to pmap over.
    env_states = jax.tree_util.tree_map(reshape_states, env_states)
    # add dimension to pmap over.
    timesteps = jax.tree_util.tree_map(reshape_states, timesteps)
```

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+
+    # RUN EXPERIMENT
+    with TimeIt(tag="COMPILATION"):
+        learn(params, opt_state, step_rngs, env_states, timesteps)
```
sash-a, line 374: This seems like you're running a full experiment before running another, we should definitely tell it to do only a single step of learning and executing, that will compile everything and take much less time.

But in general do we want this here? I understand this is useful when timing stuff, but when doing experiments I don't want to do extra work, the compilation is just part of the runtime, so this probably shouldn't be in the final product.

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+        metric = traj_batch.info
+        return runner_state, metric
+
+    def learner_fn(
+        params: FrozenDict,
+        opt_state: OptState,
+        rng: chex.PRNGKey,
+        env_state: State,
+        timesteps: TimeStep,
+    ) -> dict:
```
sash-a, line 284: If you implement runner state as mentioned above.

```suggestion
    ) -> Dict[str, Union[chex.Array, RunnerState]]:
```

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+
+        runner_state, metric = jax.lax.scan(
+            batched_update_step, runner_state, None, config["NUM_UPDATES"]
+        )
+        return {"runner_state": runner_state, "metrics": metric}
+
+    return learner_fn
+
+
+def run_experiment(env: Environment, config: dict) -> dict:
```
sash-a, line 300: In `types.py` we should have a config type something like `Dict[str, Union[chex.Array, str]`. Also could we update the return type from `dict` to `Dict[...]`

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,407 @@
...
+        "ENV_NAME": "RobotWarehouse-v0",
+        "SEED": 42,
+    }
+
+    env = jumanji.make(config["ENV_NAME"])
+    env = RwareMultiAgentWrapper(env)
+    env = AutoResetWrapper(env)
+    env = LogWrapper(env)
+
+    output = run_experiment(env, config)
```
sash-a, line 407: Not using output here, so either just do `run_experiment(env, config)` or do some sort of final print out using the output. Should probably have a final print anyways, like `print("IPPO experiment completed")`

sash-a (changes requested):
Looks really great guys! Lots of comments, but most of them are just style things :smile:

jax_distribution/systems/ff_ippo_rware.py
sash-a: Even better if this function returns a `NamedTuple` instead of a dict, since you're only ever accessing the outputs, you're never adding more fields to the dict

jax_distribution/wrappers/jumanji.py
```diff
@@ -1,121 +1,105 @@
...
-        # TODO: Fix this. This is just a basic single agent hack
-        self.n = jnp.sum(self.max_val)
-
-    def sample(self, rng: chex.PRNGKey) -> chex.Array:
-        """Sample random action uniformly from set of categorical choices."""
-        return jax.random.randint(
-            rng, shape=self.shape, minval=self.min_val, maxval=self.max_val + 1
-        ).astype(self.dtype)
+@struct.dataclass
+class LogEnvState:
```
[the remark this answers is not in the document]
OmaymaMahjoub: We need these values to log the metrics while running the trainer, we will update the docs

jax_distribution/wrappers/jumanji.py
```diff
@@ -1,121 +1,105 @@
...
-            dtype=jnp.int32,
+    def reset(self, key: chex.PRNGKey) -> Tuple[State, TimeStep]:
+        """Reset the environment. Updates the step count."""
+        state, timestep = self._env.reset(key)
+        timestep.observation = Observation(
+            agents_view=timestep.observation.agents_view,
+            action_mask=timestep.observation.action_mask,
+            step_count=jnp.repeat(
+                timestep.observation.step_count, self._env.num_agents
+            ),
```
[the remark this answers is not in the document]
OmaymaMahjoub: we tried "_replace" but it is a time consumer compared to create a new one that's why we used a new obs

jax_distribution/wrappers/jumanji.py
```diff
@@ -1,121 +1,105 @@
...
-            high=jnp.inf,
-            shape=(
-                jumanji_observation_space.agents_view.shape[0]
-                * jumanji_observation_space.agents_view.shape[1],
+        return state, timestep
+
+    def step(self, state: State, action: jnp.ndarray) -> Tuple[State, TimeStep]:
+        """Step the environment. Updates the step count."""
+        state, timestep = self._env.step(state, action)
+        timestep.observation = Observation(
```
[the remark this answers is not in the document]
OmaymaMahjoub: similar as above it is related to execution time

jax_distribution/wrappers/jumanji.py
```diff
@@ -1,121 +1,105 @@
...
-env = jumanji.make("RobotWarehouse-v0")
-
-wrapped_env = JumanjiToGymnaxWrapper(env)
-x = 0
+        return state, timestep
+
+    def observation_spec(self) -> specs.Spec[Observation]:
+        """Specification of the observation of the `RobotWarehouse` environment."""
+        step_count = specs.BoundedArray(
+            (4,),
```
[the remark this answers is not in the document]
OmaymaMahjoub: oh my bad yes will change it :sweat_smile:

jax_distribution/utils/timing_utils.py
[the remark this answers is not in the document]
OmaymaMahjoub: I will check if it is working directly without create it

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: I am not sure about this one maybe @RuanJohn or @DriesSmit  can have a better answer for that

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: Interesting, yes you made a good point, at the start we did this for testing but I guess we should either make it for one step or remove it completely

jax_distribution/wrappers/jumanji.py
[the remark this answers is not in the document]
OmaymaMahjoub: yes we will rename them :)

jax_distribution/utils/timing_utils.py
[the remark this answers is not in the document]
OmaymaMahjoub: TimeIt use timeit package but since timeit instances should be used as timeit.timeit("method"...) TimeIt give a better and more customized way to show the used time also I think we need to refer to the original implementation of it which I believe from the anakin notebook

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: for now I guess we should keep the config as dict and later on we will change it as it will be updated in the next PR where we will add the logger and evaluator

jax_distribution/types.py
```diff
@@ -13,3 +18,20 @@ class PPOTransition(NamedTuple):
...
+    """State of the `Runner`."""
+
+    params: FrozenDict
+    opt_state: OptState
+    key: chex.PRNGKey
+    env_state: LogEnvState
+    timestep: TimeStep
+
+
+class Output(NamedTuple):
```
sash-a, line 33: I think being slightly more descriptive and calling this `ExperimentOutput` is a good idea.

jax_distribution/wrappers/jumanji.py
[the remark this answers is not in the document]
sash-a: Quite strange, but if that's the case then we can leave it :+1:

sash-a (commented):
Awesome I'm super happy with all the changes, thanks @OmaymaMahjoub! :fire: 

Just waiting for @RuanJohn or @DriesSmit to look at the `pmap`ing and thinking about if we should keep the compilation stuff in there :+1:

jax_distribution/types.py
OmaymaMahjoub: Thanks for all the review points :raised_hands: I will update the name

jax_distribution/systems/ff_ippo_rware.py
RuanJohn: I got this from this [Anakin demo notebook](https://colab.research.google.com/drive/1974D-qP17fd5mLxy6QZv-ic4yxlPJp-G?usp=sharing). The reason for `pmean`-ing twice here is because we are `vmap`-ing over the vectorised updates as well and broadcasting the parameters to both the vectorised update functions and the devices.

jax_distribution/systems/ff_ippo_rware.py
sash-a: Sounds good :+1:

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -0,0 +1,441 @@
...
+
+def run_experiment(env: Environment, config: Dict) -> ExperimentOutput:
+    """Runs experiment."""
+    # Setup learner.
+    learn, initial_runner_state = learner_setup(env, config)
+
+    # Calculate total timesteps.
+    cores_count = len(jax.devices())
+    total_timesteps = (
+        cores_count
```
sash-a, line 398: Technically this returns how many CPU cores/GPUs/TPU cores so rather call it devices. As you could have 2 GPUs for example.

```suggestion
    n_devices = len(jax.devices())
    total_timesteps = (
        n_devices
```

sash-a (approved):
2 very minor remaining comments, add the comment above `pmap` and changes `cores_count` to `n_devices` otherwise looks great!

sash-a (approved):
:fire:

PR #906, opened 2023-07-31
Feat: Add evaluator and logger to  FeedForward IPPO 

OmaymaMahjoub:
## What?
Adding Evaluator and logger 
## Why?
We need the evaluator to use the MARL evaluation guideline, and the logger to log the evaluator outputs into json and tensorboard.
## Extra
We need to close #905  before reviewing this PR


sash-a (commented):
Haven't properly looked at the IPPO file yet.

Everything looks solid :fire: Have some concerns especially the file length, we should consider splitting non-essential stuff into new files very soon.

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -389,55 +512,169 @@ def learner_setup(env: Environment, config: Dict) -> Tuple[callable, RunnerState
...
-
-    return learner_output
+        learn(runner_state)
+    with TimeIt(tag="COMPILATION"):
+        evaluator(trained_params, eval_rngs)
+
+    # Run experiment for a total number of evaluations.
+    max_episode_return = jnp.float32(0.0)
+    best_params = None
+    for i in range(config["NUM_EVALUATION"]):
```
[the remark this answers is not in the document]
sash-a: nvm there's a lot of IO here, so I guess no scan makes sense

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
-        return {"runner_state": runner_state, "metrics": metric}
+        total_loss, (value_loss, loss_actor, entropy) = loss_info
+        return {
+            "runner_state": runner_state,
+            "metrics": metric,
+            "total_loss": total_loss,
+            "value_loss": value_loss,
+            "loss_actor": loss_actor,
+            "entropy": entropy,
+        }
```
sash-a, line 333: When things get this long and the output is static consider making them `NamedTuples` in this case only do it if MAPPO and recurrent have the same output though

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
 
 
-def learner_setup(env: Environment, config: Dict) -> Tuple[callable, RunnerState]:
+def get_evaluator_fn(env: Environment, apply_fn: callable, config: dict) -> callable:
+    """Get the evaluator function."""
+
+    def eval_one_episode(params, runner_state) -> Tuple:
+        """Evaluate one episode. It is vectorized over the number of evaluation episodes."""
+
+        def _env_step(runner_state: Tuple) -> Tuple:
```
sash-a, line 344: Seems like this is duplicate code of [here](https://github.com/instadeepai/Mava/blob/1e9f67a5c9a1756a59f51598fc8b2bcd60729135/jax_distribution/systems/ff_ippo_rware.py#L122), if you pulled out to a common method you'd just need to pass in params, runner state and eval mode

Also you have a runner state type, so should that not be used instead of tuple?

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
+            """Step the environment."""
+            # PRNG keys.
+            rng, env_state, last_timestep, step_count_, return_ = runner_state
```
sash-a, line 347: Why do step_count_ and return_ have an _ at the end?

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
+            # Step environment.
+            env_state, timestep = env.step(env_state, action)
+
+            # Log episode metrics.
+            return_ += timestep.reward
+            step_count_ += 1
+            runner_state = (rng, env_state, timestep, step_count_, return_)
+            return runner_state
+
+        def is_done(carry: Tuple) -> jnp.bool_:
```
sash-a, line 367: As mentioned previous avoid the `jnp.xyz` for types. Use chex.Numeric/jax.typing.ArrayLike or in this case just `bool`

```suggestion
        def is_done(carry: Tuple) -> bool:
```

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
+            """Check if the episode is done."""
+            timestep = carry[2]
+            return ~timestep.last()
```
sash-a, line 370: Is this a bug? If it's the last timestep then you are done?

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
+
+        rng, env_state, timestep = runner_state
+        return_ = jnp.array(0, float)
+        step_count_ = jnp.array(0, int)
```
sash-a, line 374: Again why the `_`. Also you can just do 0.0 and 0, as long as you type stuff as mentioned previously it will all be converted when passed to `jax.jit`

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
+
+        final_runner = jax.lax.while_loop(
```
sash-a, line 376: In jax while loops can be quite slow. Have you done a speed comparison with a fixed number of steps and then masked the output?

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
+    ) -> Dict[str, Dict[str, chex.Array]]:
+        """Evaluator function."""
+
+        # Initialise environment states and timesteps.
+        n_devices = len(jax.devices())
+
+        eval_batch = config["NUM_EVAL_EPISODES"] // n_devices
+
+        rng, *env_rngs = jax.random.split(rng, eval_batch + 1)
+        env_states, timesteps = jax.vmap(env.reset, in_axes=(0))(
```
sash-a, line 400: in_axes=0 is default

```suggestion
        env_states, timesteps = jax.vmap(env.reset)(
```

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
+        step_rngs = reshape_step_rngs(jnp.stack(step_rngs))
+        runner_state = (step_rngs, env_states, timesteps)
+
+        eval_metrics = jax.vmap(
+            eval_one_episode, in_axes=(None, 0), axis_name="eval_batch"
+        )(trained_params, runner_state)
+
+        return {"metrics": eval_metrics}
+
+    def absolute_evaluator_fn(
```
sash-a, line 416: These two functions seem identical to me

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -389,55 +512,169 @@ def learner_setup(env: Environment, config: Dict) -> Tuple[callable, RunnerState
...
     with TimeIt(tag="COMPILATION"):
-        learn(initial_runner_state)
-
-    with TimeIt(tag="EXECUTION", environment_steps=total_timesteps):
-        learner_output = learn(initial_runner_state)
-
-    return learner_output
+        learn(runner_state)
+    with TimeIt(tag="COMPILATION"):
+        evaluator(trained_params, eval_rngs)
```
sash-a, line 587: For @RuanJohn, did we not discuss and say we would remove the explicit compilation step?

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -389,55 +512,169 @@ def learner_setup(env: Environment, config: Dict) -> Tuple[callable, RunnerState
...
+            t_env=timesteps_per_training * (i + 1),
+        )
+        if config["ABSOLUTE_METRIC"] and max_episode_return <= episode_return:
+            best_params = copy.deepcopy(trained_params)
+            max_episode_return = episode_return
+
+        # Update runner state to continue training.
+        runner_state = learner_output["runner_state"]
+
+    if config["ABSOLUTE_METRIC"]:
```
sash-a, line 627: I don't think this should even be a config option, we should just do it. If it must be a config option I think "final_eval" is a better name

sash-a (changes requested):
Looks great :fire: Two changes I think would greatly help reduce the duplicate code: 

* Make the `eval_one_episode` method more global so that the learner can use it also.
* Only have a single evaluator fn, they look the same to me

jax_distribution/logger.py
```diff
@@ -0,0 +1,136 @@
...
+            metrics_info = metrics
+
+        # Flatten metrics info.
+        episodes_return = jnp.ravel(metrics_info["episode_return"])
+        episodes_length = jnp.ravel(metrics_info["episode_length"])
+
+        # Log metrics.
+        if config["USE_SACRED"] or config["USE_TF"]:
+            logger.log_stat(
+                prefix.lower() + "mean_test_episode_returns",
```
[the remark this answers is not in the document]
OmaymaMahjoub: nice catch we need to remove test

jax_distribution/logger.py
```diff
@@ -0,0 +1,136 @@
...
+            value_loss = metrics["value_loss"]
+            loss_actor = metrics["loss_actor"]
+            entropy = metrics["entropy"]
+        else:
+            prefix = ""
+            metrics_info = metrics
+
+        # Flatten metrics info.
+        episodes_return = jnp.ravel(metrics_info["episode_return"])
+        episodes_length = jnp.ravel(metrics_info["episode_length"])
```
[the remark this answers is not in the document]
OmaymaMahjoub: the reason behind that is because this variable include multiple episodes return this is why we kept  it episodes_return, what do you think?

jax_distribution/logger.py
```diff
@@ -0,0 +1,136 @@
...
+        log_string += "| Max Episode Returns {:.3f} ".format(
+            float(np.max(episodes_return))
+        )
+        log_string += "| Mean Episode Length {:.3f} ".format(
+            float(np.mean(episodes_length))
+        )
+        log_string += "| Std Episode Length {:.3f} ".format(
+            float(np.std(episodes_length))
+        )
+        log_string += "| Max Episode Length {:.3f} ".format(
```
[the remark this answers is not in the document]
OmaymaMahjoub: Certainly, no problem at all!

jax_distribution/logger.py
```diff
@@ -0,0 +1,136 @@
...
+            float(np.max(episodes_length))
+        )
```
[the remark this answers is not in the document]
OmaymaMahjoub: This is because we wanted to have round the values by having 3 after the "," however we can change it to f"string" if you want and we use by that the `round` function to achieve what we wanted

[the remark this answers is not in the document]
OmaymaMahjoub: I updated it to look like this: 
```
log_string = (
            "Timesteps {:07d} | "
            "Mean Episode Return {:.3f} | "
            "Std Episode Return {:.3f} | "
            "Max Episode Return {:.3f} | "
            "Mean Episode Length {:.3f} | "
            "Std Episode Length {:.3f} | "
            "Max Episode Length {:.3f}"
        ).format(
            t_env,
            float(np.mean(episodes_return)),
            float(np.std(episodes_return)),
            float(np.max(episodes_return)),
            float(np.mean(episodes_length)),
            float(np.std(episodes_length)),
            float(np.max(episodes_length))
        )
```
I don't know if this solved the issue of "+=" because using """string""" causing some spacing issues in the terminal output

jax_distribution/utils/logger_tools.py
```diff
@@ -0,0 +1,108 @@
...
+    def setup_sacred(self, sacred_run_dict: Dict) -> None:
+        """Set up sacred logging."""
+        self._run_obj = sacred_run_dict
+        self.sacred_info = sacred_run_dict.info
+        self.use_sacred = True
+
+    def setup_neptune(self, neptune_run_dict: Dict) -> None:
+        pass
+
+    def log_stat(self, key: str, value: float, t: int, to_sacred: bool = True) -> None:
```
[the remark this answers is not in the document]
OmaymaMahjoub: Yeah  nice catch will update that

jax_distribution/utils/logger_tools.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    ch.setFormatter(formatter)
+    logger.addHandler(ch)
+    # Set to info to suppress debug outputs.
+    logger.setLevel("INFO")
+
+    return logger
+
+
+def recursive_dict_update(d: Dict, u: Dict) -> Dict:
```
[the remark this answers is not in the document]
OmaymaMahjoub: I guess we can even eliminate this as we don't even used it

jax_distribution/utils/logger_tools.py
```diff
@@ -0,0 +1,108 @@
...
+    """Recursively update a dictionary."""
+    for k, v in u.items():
+        if isinstance(v, collections.Mapping):
+            d[k] = recursive_dict_update(d.get(k, {}), v)
+        else:
+            d[k] = v
+    return d
+
+
+def config_copy(config: Dict) -> Dict:
```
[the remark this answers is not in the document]
OmaymaMahjoub: This is because the _config from sacred is set to be read-only however we make changes to config while running the system to get different needed information shared between the different functions thanks to config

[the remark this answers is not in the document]
OmaymaMahjoub: if you have another option happy to give it a try

jax_distribution/wrappers/jumanji.py
```diff
@@ -29,11 +29,11 @@ class LogEnvState:
     """State of the `LogWrapper`."""
 
     env_state: State
-    episode_returns: float
-    episode_lengths: int
+    episode_returns: jnp.float32
```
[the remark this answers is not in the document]
OmaymaMahjoub: maybe we can have a chat about this because we mainly did this to force having consistent type as skipping this caused us to have a function traced multiple time

jax_distribution/systems/ff_ippo_rware.py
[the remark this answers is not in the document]
OmaymaMahjoub: also pmapping inside a scan can cause some issues

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -49,19 +60,14 @@ def __call__(
...
-        if self.activation == "relu":
-            activation = nn.relu
-        else:
-            activation = nn.tanh
-
         actor_output = nn.Dense(
             64, kernel_init=orthogonal(np.sqrt(2)), bias_init=constant(0.0)
         )(x)
-        actor_output = activation(actor_output)
+        actor_output = nn.relu(actor_output)
```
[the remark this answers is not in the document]
OmaymaMahjoub: we just fixed the activation to be relu but happy to bring it back to be a part of the config

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: yeah this is the difference between this step and the old function, this one we can directly cumulative the val of step count and return therefore they are a part of the runner_state

OmaymaMahjoub: I will update the name of runner_state to eval_state and make it a named tuple

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: also changed the value types to chex.Numeric thanks for the suggestion

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -317,15 +319,133 @@ def learner_fn(runner_state: RunnerState) -> ExperimentOutput:
...
+            """Check if the episode is done."""
+            timestep = carry[2]
+            return ~timestep.last()
+
+        rng, env_state, timestep = runner_state
+        return_ = jnp.array(0, float)
+        step_count_ = jnp.array(0, int)
+
+        final_runner = jax.lax.while_loop(
+            is_done,
```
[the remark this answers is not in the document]
OmaymaMahjoub: oh my bad yeah :sweat_smile:

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: I will test it, however this is only used for evaluation phase so it won't bother with the training speed

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: the absolute metric does exist in theory with this name as it is a long evaluation but with the best trained network, also we need to keep it optional as we just use it if we need to produce eval_marl plots only

jax_distribution/config/rware_env.py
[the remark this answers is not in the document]
OmaymaMahjoub: We update it to be in yaml file, if you can give it a look and give feedback related to it :pray:

jax_distribution/systems/ff_ippo_rware.py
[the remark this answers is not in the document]
OmaymaMahjoub: I moved the config to be in a yaml, however further reduction we need @RuanJohn opinion too

jax_distribution/config/rware_env.py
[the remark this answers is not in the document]
sash-a: That's great was going to say to do this, but wasn't sure if you guys needed it as a python file

jax_distribution/logger.py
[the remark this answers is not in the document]
sash-a: Would this work? I just think `f"string"` is much easier to read

```python
log_string = (
            f"Timesteps {t_env:07d} | "
            f"Mean Episode Return {float(np.mean(episodes_return)):.3f} | "
            f"Std Episode Return {float(np.std(episodes_return)):.3f} | "
            f"Max Episode Return {float(np.max(episodes_return)):.3f} | "
            f"Mean Episode Length {float(np.mean(episodes_length)):.3f} | "
            f"Std Episode Length {float(np.std(episodes_length)):.3f} | "
            f"Max Episode Length {float(np.max(episodes_length)):.3f}"
        )
```

jax_distribution/systems/ff_ippo_rware.py
sash-a: Don't worry too much about this, was just curious

jax_distribution/configs/default.yaml
sash-a, line 1: in general I think keys in yaml files should be lower case

jax_distribution/logger.py
```diff
@@ -82,37 +81,42 @@ def log(
...
+            float(np.max(episodes_return)),
+            float(np.mean(episodes_length)),
+            float(np.std(episodes_length)),
+            float(np.max(episodes_length)),
         )
 
         if absolute_metric:
-            logger.console_logger.info("ABSOLUTE METRIC: " + log_string)
+            logger.console_logger.info(
+                f"{Fore.BLUE}{Style.BRIGHT}ABSOLUTE METRIC: {log_string}{Style.RESET_ALL}"
```
sash-a, line 104: Like the colours, those are useful!

jax_distribution/types.py
```diff
@@ -42,8 +43,22 @@ class RunnerState(NamedTuple):
     timestep: TimeStep
 
 
+class EvalState(NamedTuple):
```
sash-a, line 46: Sorry for all the renames, thinking that if you now have `EvalState` we should maybe rename `RunnerState` to `LearnerState` or maybe `ActorState`, what do you think?

jax_distribution/systems/ff_ippo_rware.py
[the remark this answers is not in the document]
RuanJohn: We ended up using only relu, so it didn't make sense to include the option for tanh anymore in the config. 
It is something easy to bring back if we wanted to, but for now it makes more sense to just hardcode relu.

jax_distribution/systems/ff_ippo_rware.py
RuanJohn: Very nice!

sash-a (commented):
Looking great! Some minor comments, will chat offline about the more tricky issues

jax_distribution/logger.py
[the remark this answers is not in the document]
sash-a: ah my bad keep it as is then

jax_distribution/systems/ff_ippo_rware.py
sash-a: ah but I think it is returning **not** done, so maybe just rename the function?

OmaymaMahjoub: yes I already did :sweat_smile:

jax_distribution/types.py
OmaymaMahjoub: yes totally agree we will change it to LearnerState

jax_distribution/configs/default.yaml
OmaymaMahjoub: yeah it makes sense, we will update them :)

jax_distribution/utils/logger_tools.py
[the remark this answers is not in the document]
OmaymaMahjoub: We will add it as todo in the future

jax_distribution/wrappers/jumanji.py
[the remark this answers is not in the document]
OmaymaMahjoub: We didn't get an issue by changing the variables types to chex.Numeric however when we define it we need to keep it the float ones` jnp.float32(0.0)`

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: We removed the extra iteration

jax_distribution/systems/ff_ippo_rware.py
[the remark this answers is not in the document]
OmaymaMahjoub: We moved the evaluator too to a different file

jax_distribution/requirements.txt
```diff
@@ -7,4 +7,8 @@ optax
 flax
 numpy
 brax
+sacred
+tensorboard_logger
+hydra-core==1.3.2
+omegaconf
```
RuanJohn, line 13: I think `omegaconf` comes with Hydra by default. 
Something else we'd have to set now as well is protobuf. Our code won't run if we have `protobuf > 3.20.x`. Could you try running the code in a clean Conda env to just check that our requirements file works? 
We can also remove `brax`.

RuanJohn: I ended up just making a slimmer requirements file.

jax_distribution/requirements.txt
```diff
@@ -1,14 +1,12 @@
-jax>=0.2.26
-jaxlib>=0.1.74
-gymnax
-evosax
+jax<=0.4.13
+jaxlib<=0.4.13
```
sash-a, line 2: Why less than, is there something wrong with 0.4.14?

jax_distribution/types.py
```diff
@@ -57,7 +57,7 @@ class ExperimentOutput(NamedTuple):
     """Experiment output."""
 
     episodes_info: Dict[str, chex.Array]
-    runner_state: RunnerState = None
+    learner_state: Union[LearnerState, None] = None
```
sash-a, line 60:
```suggestion
    learner_state: Optional[LearnerState] = None
```

jax_distribution/types.py
```diff
@@ -10,7 +10,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict, NamedTuple
+from typing import Dict, NamedTuple, Union
```
sash-a, line 13:
```suggestion
from typing import Dict, NamedTuple, Optional
```

jax_distribution/configs/default.yaml
```diff
@@ -6,32 +6,35 @@ defaults:
...
+vf_coef: 0.5 # Critic weight in
+max_grad_norm: 0.5 # Maximum norm of the gradients for a weight update.
 
 # --- Evaluation ---
-NUM_EVAL_EPISODES: 32
-NUM_EVALUATION: 2
-EVALUATION_GREEDY: False
-ABSOLUTE_METRIC: True
+num_eval_episodes: 32 # Episodes per evaluation
+num_evaluation: 2 # Number of evaluation intervals during training.
```
sash-a, line 31: Mostly just a question, is this `evaluation_frequency`/`evaluation_period`. Docs aren't quite clear, what do you mean by number of evaluation intervals?

sash-a (approved):
Looks good, no big concerns so when you fix the minor issues I have you can merge :fire:

jax_distribution/requirements.txt
RuanJohn: Indeed, 0.4.14 gives the following error: 
`AttributeError: module 'jax' has no attribute 'xla'`

jax_distribution/configs/default.yaml
OmaymaMahjoub: yeah maybe evaluation intervals is not clear, we either will change the name of the var to evaluation_frequency or update the comment to mention that it is related to the evaluation frequency, @RuanJohn what do you think ?

RuanJohn: `num_eval_episodes` is the number of evaluation episodes we want to do each time we evaluate and `num_evaluations` is the amount of evenly spaced evaluation to do during training.

PR #908, opened 2023-08-01
Feat: feedforward mappo vmapped agents with evaluator and logger

OmaymaMahjoub:
## What?
First fully end to end JAX MAPPO Agent that is able to train on Rware tiny-4ag.

## Note
This is joint work done with @RuanJohn and @DriesSmit 

## Extra:
We need to close #905 and #906 before reviewing this PR


jax_distribution/wrappers/jumanji.py
```diff
@@ -24,6 +24,24 @@
...
+        (up, right, down, left) is legal.
+    global_state: the global state of the environment, which is the
+        concatenation of the agents' views.
+    step_count: the number of steps elapsed since the beginning of the episode.
+    """
+
+    agents_view: chex.Array  # (num_agents, num_obs_features)
+    action_mask: chex.Array  # (num_agents, 5)
+    global_state: chex.Array  # (num_agents * num_obs_features,)
+    step_count: chex.Array  # ()
```
RuanJohn, line 42: Small fix here, the shape should be `# (num_agents, )`

OmaymaMahjoub: yes and also I need to fix global_state comment `(num_agents,num_agents* num_obs_features)`

RuanJohn: Shouldn't global state only be a single value?

RuanJohn: Ah, I had a quick look. We duplicate the global state per agent. I think this should be fine, we can keep it as is :)

jax_distribution/wrappers/jumanji.py
```diff
@@ -112,3 +130,72 @@ def observation_spec(self) -> specs.Spec[Observation]:
             "step_count",
         )
         return self._env.observation_spec().replace(step_count=step_count)
+
+
+class RwareMultiAgentWithGlobalStateWrapper(Wrapper):
+    """Multi-agent wrapper for the Robotic Warehouse environment."""
```
RuanJohn, line 136: Just a slightly more descriptive docstring. :) 
```suggestion
    """Multi-agent wrapper for the Robotic Warehouse environment that includes a global environment state
    to be used by the centralised critic. Note here that since robotic warehouse does not have a global state, 
    we create one by concatenating the observations of all agents. 
    """
```

jax_distribution/systems/ff_mappo_rware.py
```diff
@@ -0,0 +1,523 @@
...
+                    )
+                    return total_loss, (value_loss, loss_actor, entropy)
+
+                grad_fn = jax.value_and_grad(_loss_fn, has_aux=True)
+                loss_info, grads = grad_fn(
+                    params, opt_state, traj_batch, advantages, targets
+                )
+
+                # Compute the parallel mean (pmean) over the batch.
+                # This calculation is inspired by the Anakin architecture demo.
```
RuanJohn, line 257: NIT: 
```suggestion
                # This calculation is inspired by the Anakin architecture demo notebook.
                # available at https://tinyurl.com/26tdzs5x
```

jax_distribution/systems/ff_mappo_rware.py
```diff
@@ -0,0 +1,523 @@
...
+                # This pmean could be a regular mean as the batch axis is on all devices.
```
RuanJohn, line 258: NIT: 
```suggestion
                # This pmean could be a regular mean as the batch axis is on the same device.
```

jax_distribution/systems/ff_mappo_rware.py
```diff
@@ -0,0 +1,523 @@
...
+        * config["update_batch_size"]
+        * config["num_envs"]
+    )
+
+    # Run experiment for a total number of evaluations.
+    max_episode_return = jnp.float32(0.0)
+    best_params = None
+    for i in range(config["num_evaluation"]):
+        # Train.
+        with TimeIt(
```
RuanJohn, line 456: Quite nice to have the compilation time be reported on the fly like this!

jax_distribution/wrappers/jumanji.py
OmaymaMahjoub: so this will help in getting things standardized however if you think that this will cause any issues if we extend the number of agents or/and envs I can make some changes to it

jax_distribution/systems/ff_mappo_rware.py
```diff
@@ -0,0 +1,523 @@
...
+        rng_e, *eval_rngs = jax.random.split(rng_e, n_devices + 1)
+        eval_rngs = jnp.stack(eval_rngs)
+        eval_rngs = eval_rngs.reshape(n_devices, -1)
+
+        # Evaluate.
+        evaluator_output = evaluator(trained_params, eval_rngs)
+        jax.block_until_ready(evaluator_output)
+
+        # Log the results
+        log(
```
RuanJohn, line 476: Minor detail, but we should log the learner output before the evaluation happens.

jax_distribution/systems/ff_mappo_rware.py
```diff
@@ -0,0 +1,523 @@
...
+    ex.captured_out_filter = utils.apply_backspaces_and_linefeeds
+    results_path = os.path.join(dirname(dirname(abspath(__file__))), "results")
+
+    file_obs_path = os.path.join(results_path, f"sacred/{cfg['env_name']}")
+    ex.observers = [observers.FileStorageObserver.create(file_obs_path)]
+    ex.add_config(OmegaConf.to_container(cfg, resolve=True))
+    ex.main(run_experiment)
+    ex.run(config_updates={})
+
+    print(f"{Fore.CYAN}{Style.BRIGHT}MAPPO experiment completed{Style.RESET_ALL}")
```
RuanJohn, line 519: Really really like the colours! :fire:

jax_distribution/systems/ff_mappo_rware.py
OmaymaMahjoub: Nice catch will update it for the ff-ippo too

jax_distribution/evaluator.py
```diff
@@ -130,22 +135,25 @@ def evaluator_setup(
...
     # Get available TPU cores.
     n_devices = len(jax.devices())
 
-    # Vmap it over number of agents.
-    vmapped_eval_network_apply_fn = jax.vmap(
-        network.apply,
-        in_axes=(None, 0),
-    )
+    # Vmap it over number of agents based on the type of network.
+    if centralised:
```
RuanJohn, line 145: Very minor suggestion: 
```suggestion
    if centralised_critic:
```

jax_distribution/evaluator.py
```diff
@@ -130,22 +135,25 @@ def evaluator_setup(
...
+        vmapped_eval_network_apply_fn = jax.vmap(
+            network.apply,
+            in_axes=(None, ObservationGlobalState(0, 0, None, 0)),
```
RuanJohn, line 148: Very nice! :sunglasses:

jax_distribution/systems/ff_mappo_rware.py
```diff
@@ -0,0 +1,537 @@
...
+                    )
+                    return total_loss, (value_loss, loss_actor, entropy)
+
+                grad_fn = jax.value_and_grad(_loss_fn, has_aux=True)
+                loss_info, grads = grad_fn(
+                    params, opt_state, traj_batch, advantages, targets
+                )
+
+                # Compute the parallel mean (pmean) over the batch.
+                # This calculation is inspired by the Anakin architecture demo notebook.
```
RuanJohn, line 257: Please add underneath here in a comment as well 
`# available at https://tinyurl.com/26tdzs5x`

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -229,8 +229,8 @@ def _loss_fn(
                 loss_info, grads = grad_fn(params, opt_state, traj_batch, advantages, targets)
 
                 # Compute the parallel mean (pmean) over the batch.
-                # This calculation is inspired by the Anakin architecture demo.
-                # This pmean could be a regular mean as the batch axis is on all devices.
+                # This calculation is inspired by the Anakin architecture demo notebook.
```
RuanJohn, line 232: Similarly a comment underneath here with a link to the notebook: 
`# available at https://tinyurl.com/26tdzs5x`

jax_distribution/systems/ff_mappo_rware.py
```diff
@@ -0,0 +1,508 @@
...
+        critic_output = nn.relu(critic_output)
+        critic_output = nn.Dense(64, kernel_init=orthogonal(np.sqrt(2)), bias_init=constant(0.0))(
+            critic_output
+        )
+        critic_output = nn.relu(critic_output)
+        critic_output = nn.Dense(1, kernel_init=orthogonal(1.0), bias_init=constant(0.0))(
+            critic_output
+        )
+
+        return actor_policy, jnp.squeeze(critic_output, axis=-1)
```
sash-a, line 92: Interesting this uses the same network? I would have though we'd only need to call the critic once per all agents?

jax_distribution/systems/ff_mappo_rware.py
```diff
@@ -0,0 +1,508 @@
...
+        update_state, loss_info = jax.lax.scan(
+            _update_epoch, update_state, None, config["ppo_epochs"]
+        )
+
+        params, opt_state, traj_batch, advantages, targets, rng = update_state
+        learner_state = LearnerState(params, opt_state, rng, env_state, last_timestep)
+        metric = traj_batch.info
+        return learner_state, (metric, loss_info)
+
+    def learner_fn(learner_state: LearnerState) -> ExperimentOutput:
```
sash-a, line 285: Nice, thought you might have to change these types, good to see that they're just the same for this case :fire:

jax_distribution/systems/ff_mappo_rware.py
sash-a: +1, very cool

jax_distribution/wrappers/jumanji.py
```diff
@@ -103,3 +121,65 @@ def observation_spec(self) -> specs.Spec[Observation]:
             "step_count",
         )
         return self._env.observation_spec().replace(step_count=step_count)
+
+
+class RwareMultiAgentWithGlobalStateWrapper(Wrapper):
```
sash-a, line 126: Seems like we're going to need to do this for each env, so in future I assume this would go in `wrappers/jumanji/rware.py`?

sash-a (approved):
Amazing, super clean! Only issue is that I think the name `RwareMultiAgentWithGlobalStateWrapper` is a bit long, but not sure if there's much we can do about it, so I'm happy to merge as is! :rocket:

jax_distribution/wrappers/jumanji.py
OmaymaMahjoub: yes agree!

jax_distribution/systems/ff_mappo_rware.py
OmaymaMahjoub: We used out_axes  None for the critic to ignore/pass the multiple calculation

PR #909, opened 2023-08-01
Feat: recurrent ippo vmapped agents

OmaymaMahjoub:
## What?
First fully end to end JAX recurrent IPPO Agent that is able to train on Rware tiny-4ag. 

## Note
This is joint work done with @RuanJohn  and @DriesSmit.


jax_distribution/types.py
```diff
@@ -43,6 +43,18 @@ class LearnerState(NamedTuple):
...
+class RNNLearnerState(NamedTuple):
+    """State of the `Learner` for recurrent architectures."""
+
+    params: FrozenDict
+    opt_state: OptState
+    key: chex.PRNGKey
+    env_state: LogEnvState
+    timestep: TimeStep
+    dones: chex.Array
+    hstate: chex.Array
```
RuanJohn, line 55: Something that could be worth look at is having this be something like 
`hstates: Union[chex.Array, Tuple[chex.Array, chex.Array]]`
This could enable use to reuse the same `NamedTuple` for rec ippo and rec mappo. It might be a bit confusing in the rec ippo case to the have a plural word here when we only have a single hidden state though. Keen to hear your and @sash-a's thoughts on this.

jax_distribution/types.py
```diff
@@ -53,6 +65,18 @@ class EvalState(NamedTuple):
...
 
 
+class RNNEvalState(NamedTuple):
+    """State of the evaluator for recurrent architectures."""
+
+    key: chex.PRNGKey
+    env_state: State
+    timestep: TimeStep
+    dones: chex.Array
+    init_hstate: chex.Array
```
RuanJohn, line 75: Similarly here to above.

jax_distribution/types.py
sash-a, line 55: Can we make `LearnerState` a `chex.dataclass` and then 
```suggestion
@chex.dataclass
class RNNLearnerState(LearnerState):
    dones: chex.Array
    hstate: chex.Array
```

jax_distribution/types.py
```diff
@@ -53,6 +65,18 @@ class EvalState(NamedTuple):
     return_: chex.Numeric = None
 
 
+class RNNEvalState(NamedTuple):
```
sash-a, line 68: Also same as above

jax_distribution/evaluator.py
```diff
@@ -124,35 +125,185 @@ def evaluator_fn(trained_params: FrozenDict, rng: chex.PRNGKey) -> ExperimentOut
...
+        evaluator = get_ff_evaluator_fn(
+            eval_env,
+            vmapped_eval_network_apply_fn,
+            config,
+        )
+        absolute_metric_evaluator = get_ff_evaluator_fn(
+            eval_env,
+            vmapped_eval_network_apply_fn,
+            config,
+            10,
```
sash-a, line 305:
```suggestion
            10,  # todo: add this to the config
```

sash-a (changes requested):
Couple quite minor things, but otherwise looking great

jax_distribution/evaluator.py
```diff
@@ -124,35 +125,185 @@ def evaluator_fn(trained_params: FrozenDict, rng: chex.PRNGKey) -> ExperimentOut
...
+    return evaluator_fn
+
+
 def evaluator_setup(
     eval_env: Environment,
     rng_e: chex.PRNGKey,
     network: Any,
     params: FrozenDict,
     config: Dict,
+    rnn_net: bool = False,
```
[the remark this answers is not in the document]
RuanJohn: Maybe even something like `recurrent_policy` ?

jax_distribution/evaluator.py
```diff
@@ -124,35 +125,185 @@ def evaluator_fn(trained_params: FrozenDict, rng: chex.PRNGKey) -> ExperimentOut
...
+            vmapped_eval_network_apply_fn,
+            config,
+            ScannedRNN,
+        )
+        absolute_metric_evaluator = get_rnn_evaluator_fn(
+            eval_env,
+            vmapped_eval_network_apply_fn,
+            config,
+            ScannedRNN,
+            10,
```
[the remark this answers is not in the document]
RuanJohn: The absolute metric is suggested to be 10. So I don't even think we need it in the config. 
Maybe just a comment to say that this is 10 because of the default absolute metric requirements.

jax_distribution/evaluator.py
RuanJohn: Same as above.

jax_distribution/systems/rec_ippo_rware.py
```diff
@@ -0,0 +1,665 @@
...
+                lambda x: jnp.swapaxes(
+                    jnp.reshape(
+                        x,
+                        [x.shape[0], config["num_minibatches"], -1] + list(x.shape[2:]),
+                    ),
+                    1,
+                    0,
+                ),
+                shuffled_batch,
+            )
```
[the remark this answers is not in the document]
RuanJohn: Totally agree on this one. 
Maybe just `shaped_batch` -> `reshaped_batch`.

jax_distribution/evaluator.py
OmaymaMahjoub: should we update it now better than keep it to do?

jax_distribution/evaluator.py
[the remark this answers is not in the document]
sash-a: I'd really like the `use` in there, just because it makes it clear that it's a `bool`, so `use_recurrent_policy`?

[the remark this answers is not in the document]
RuanJohn: Makes sense, I think the original `use_recurrent_net` is the best option here. Since we could have recurrent policies and critics.

sash-a (approved):
Looks good to me :smile:

jax_distribution/systems/rec_ippo_rware.py
```diff
@@ -361,17 +352,14 @@ def _loss_fn(
...
             )
-            minibatches = jax.tree_util.tree_map(
-                lambda x: jnp.swapaxes(
-                    jnp.reshape(
-                        x,
-                        [x.shape[0], config["num_minibatches"], -1] + list(x.shape[2:]),
-                    ),
-                    1,
-                    0,
+            reshaped_batch = jax.tree_util.tree_map(
```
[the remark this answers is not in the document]
RuanJohn: Sounds good!

PR #910, opened 2023-08-10
chore: release ci cleanup

sash-a:
Reconfigured the workflow, requirements, tests and the like. Mostly copying jumanji were appropriate


sash-a:
Probably going to need merge this in after all the old mava stuff has been deleted, otherwise it won't pass the checks

.gitignore
```diff
@@ -135,33 +135,31 @@ dmypy.json
...
 logs
 Checkpoints
-mava/workspace.code-workspace
 .DS_Store
-mava_testing
 .devcontainer/devcontainer.json
+3.9
+
+# smac
 3rdparty
```
RuanJohn, line 155: Since we won't be supporting smac we can remove this for now.

setup.py
```diff
@@ -13,129 +12,55 @@
...
-        "record_episode": record_episode_requirements,
-        "sc2": smac_requirements,
-        "envs": pettingzoo_requirements + smac_requirements,
-        "jax": jax_requirements,
+        "dev": _parse_requirements("requirements/requirements-dev.txt"),
     },
+    package_data={"mava": ["py.typed"]},
     classifiers=[
-        "Development Status :: 3 - Alpha",
+        "Development Status :: 4 - Beta",
```
RuanJohn, line 54: No exactly sure about this? Aren't we in alpha at the moment?

RuanJohn (changes requested):
Thanks @sash-a! What a mammoth job! This looks super clean!
Just a few minor requests from my side.

RuanJohn (approved):
Looks good to me! Thanks @sash-a :fire:

RuanJohn:
> 

Very good points, just some comments on them: 
* The troubleshooting will get removed in the readme PR. 
* For now, let's hold off on doing this and have one PR once all the systems are in where we move things around. 
* Very good catch. Can we please bring the Mava logo back @sash-a? My suggestion is that we keep it in the docs folder in `docs/images/`.

RuanJohn (approved):
Thanks @sash-a :raised_hands:

PR #911, opened 2023-08-11
Feat: recurrent mappo vmapped agents

OmaymaMahjoub:
## What?
First fully end to end JAX recurrent MAPPO Agent that is able to train on Rware scenarios. 

## Note
This is joint work done with @RuanJohn.


jax_distribution/evaluator.py
```diff
@@ -138,35 +125,212 @@ def evaluator_fn(trained_params: FrozenDict, rng: chex.PRNGKey) -> ExperimentOut
...
-    )
+    # Vmap it over number of agents and create evaluator_fn.
+    if rnn_net:
+        if centralised_critic:
+            vmapped_eval_network_apply_fn = jax.vmap(
+                network.apply,
+                in_axes=(
+                    None,
+                    (1, None),
+                    (ObservationGlobalState(2, 2, None, 2), None),
```
sash-a, line 295: Howcome you don't need to do this if we're ff centralised?

jax_distribution/logger.py
```diff
@@ -19,26 +19,24 @@
...
 from sacred.run import Run
 
-from jax_distribution.types import ExperimentOutput
 from jax_distribution.utils.logger_tools import Logger
 
 
-# todo: flake is complaining, possibly split into 3 functions
-#  log_absolute, log_trainer, log_evaluator
-def get_logger_fn(logger: Logger, config: Dict) -> Callable:  # noqa: CCR001
+def get_logger_fn(logger: SacredLogger, config: Dict) -> Callable:  # noqa: CCR001
```
sash-a, line 31: I think revert this file as the old stuff is correct, assume it is still being merged

jax_distribution/systems/rec_mappo_rware.py
```diff
@@ -0,0 +1,737 @@
...
+
+            rng, policy_rng = jax.random.split(rng)
+
+            # Add a batch dimension to the observation.
+            batched_observation = jax.tree_util.tree_map(
+                lambda x: x[np.newaxis, :], last_timestep.observation
+            )
+            ac_in = (
+                batched_observation,
+                last_done[:, 0][np.newaxis, :],
```
sash-a, line 198: Only seeing this now, but I think in other places we use `None` to add a new axis, I actually quite like `np.newaxis` as it's clearer what is happening, but we should just stay consistent, so either change these to `None` or change the others to `np.newaxis`

jax_distribution/systems/rec_mappo_rware.py
```diff
@@ -0,0 +1,737 @@
...
+                    # CALCULATE ACTOR LOSS
+                    ratio = jnp.exp(log_prob - traj_batch.log_prob)
+                    gae = (gae - gae.mean()) / (gae.std() + 1e-8)
+                    loss_actor1 = ratio * gae
+                    loss_actor2 = (
+                        jnp.clip(
+                            ratio,
+                            1.0 - config["clip_eps"],
+                            1.0 + config["clip_eps"],
+                        )
```
sash-a, line 357: Also only saw this now, would need to change all files, but might be better to pull this out into `clipped_ratio = ...` for readability

jax_distribution/systems/rec_mappo_rware.py
```diff
@@ -0,0 +1,737 @@
...
+                init_critic_hstate,
+                traj_batch,
+                advantages,
+                targets,
+            )
+            shuffled_batch = jax.tree_util.tree_map(
+                lambda x: jnp.take(x, permutation, axis=1), batch
+            )
+            minibatches = jax.tree_util.tree_map(
+                lambda x: jnp.swapaxes(
```
sash-a, line 413: As commented on previous PR pull out the `reshape`

jax_distribution/systems/rec_mappo_rware.py
sash-a: Having gone through the PR I don't see adding a new axis with `None` anywhere, so happy for you to keep things as they are. Just have a look in the other files as I feel like I saw it somewhere :shrug:

sash-a (commented):
Nice looks great, some more very minor stuff! Just commenting so I don't block the merge later. Good luck!! :fire:

jax_distribution/logger.py
sash-a: Seems like there was a mix up in the merge, definitely revert this file

jax_distribution/evaluator.py
OmaymaMahjoub: We already did the same for the mappo check this[ line](https://github.com/instadeepai/Mava/blob/feat/pure-jax-mava/jax_distribution/systems/ff_mappo_rware.py#L355)

jax_distribution/systems/rec_mappo_rware.py
OmaymaMahjoub: We changed them to jnp.newaxis

jax_distribution/systems/rec_mappo_rware.py
OmaymaMahjoub: We feel like this is already descriptive enough

OmaymaMahjoub: are you okay to reconsider

jax_distribution/systems/rec_mappo_rware.py
```diff
@@ -409,17 +382,14 @@ def _loss_fn(
...
-                    jnp.reshape(
-                        x,
-                        [x.shape[0], config["num_minibatches"], -1] + list(x.shape[2:]),
-                    ),
-                    1,
-                    0,
+            reshaped_batch = jax.tree_util.tree_map(
+                lambda x: jnp.reshape(
+                    x,
+                    [x.shape[0], config["num_minibatches"], -1] + list(x.shape[2:]),
```
sash-a, line 388: Can we change this slightly, here and in the other PR::

```suggestion
                    x, (x.shape[0], config["num_minibatches"], -1, *x.shape[2:])
```

sash-a (approved):
Nice! One very nitpicky change, could you also do it in the other files please :smile:

RuanJohn (approved):
THis looks good to me! Thank you @OmaymaMahjoub! :fire:

PR #912, opened 2023-08-15
Feat/agent id wrapper

RuanJohn:
## What?
Added wrapper that can add one-hot agent ID to agent observations. 
## Why?
When parameter sharing among agents this allows for learning more diverse policies.

### DIFF
--- a/jax_distribution/systems/ff_ippo_rware.py
+++ b/jax_distribution/systems/ff_ippo_rware.py
@@ -44,7 +44,11 @@ from jax_distribution.types import ExperimentOutput, LearnerState, PPOTransition
 from jax_distribution.utils.jax import merge_leading_dims
 from jax_distribution.utils.logger_tools import config_copy, get_logger
 from jax_distribution.utils.timing_utils import TimeIt
-from jax_distribution.wrappers.jumanji import LogWrapper, RwareMultiAgentWrapper
+from jax_distribution.wrappers.jumanji import (
+    AgentIDWrapper,
+    LogWrapper,
+    RwareMultiAgentWrapper,
+)
 
 
 class ActorCritic(nn.Module):
@@ -384,10 +388,12 @@ def run_experiment(_run: run.Run, _config: Dict, _log: SacredLogger) -> None:
     generator = RandomGenerator(**config["rware_scenario"])
     # Create envs
     env = jumanji.make(config["env_name"], generator=generator)
+    env = AgentIDWrapper(env)
     env = RwareMultiAgentWrapper(env)
     env = AutoResetWrapper(env)
     env = LogWrapper(env)
     eval_env = jumanji.make(config["env_name"], generator=generator)
+    eval_env = AgentIDWrapper(eval_env)
     eval_env = RwareMultiAgentWrapper(eval_env)
 
     # PRNG keys.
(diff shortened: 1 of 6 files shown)


jax_distribution/wrappers/jumanji.py
```diff
@@ -88,6 +89,72 @@ def step(
...
+                step_count=timestep.observation.step_count,
+                global_state=timestep.observation.global_state,
+            )
+
+        else:
+            timestep.observation = Observation(
+                agents_view=new_agents_view,
+                action_mask=timestep.observation.action_mask,
+                step_count=timestep.observation.step_count,
+            )
```
sash-a, line 120: Instead of the if/else block could you not just do: `timestep.observation.replace(agents_view=new_agents_view)`? Might be `_replace` for `NamedTuple`, I can never remember

jax_distribution/wrappers/jumanji.py
```diff
@@ -88,6 +89,72 @@ def step(
...
```
sash-a, line 146: Same as above

jax_distribution/systems/ff_ippo_rware.py
```diff
@@ -385,10 +389,14 @@ def run_experiment(_run: run.Run, _config: Dict, _log: SacredLogger) -> None:
     # Create envs
     env = jumanji.make(config["env_name"], generator=generator)
     env = RwareMultiAgentWrapper(env)
+    if config["add_agent_id"]:
```
OmaymaMahjoub, line 392: you can make both env and eval_env under one if condition as the remaining code after env is independent of it

sash-a (changes requested):
Looks good, just the if/else

jax_distribution/wrappers/jumanji.py
OmaymaMahjoub: the issue with _replace it consumes time more than creating new Observation so maybe this is one of @RuanJohn  reasons

RuanJohn: That is exactly the reason. For some reason replacing was slower than just recreating the observation.
I am not sure why this is the case. I can test it again on this PR.

jax_distribution/systems/ff_ippo_rware.py
RuanJohn: The reason for having two if statements is because we need the training environment to be wrapped in the `AutoReset` and `Log` wrappers. So doing both at the end would cause trouble as far as I can see.

jax_distribution/wrappers/jumanji.py
sash-a: After a chat, we've decided to test it after the release

jax_distribution/systems/ff_ippo_rware.py
OmaymaMahjoub: We can write it this way: 
```
env = jumanji.make(config["env_name"], generator=generator)
env = RwareMultiAgentWrapper(env)
eval_env = RwareMultiAgentWrapper(env)
if config["add_agent_id"]:
        env = AgentIDWrapper(env)
        eval_env = AgentIDWrapper(eval_env)
env = AutoResetWrapper(env)
env = LogWrapper(env)
```

PR #914, opened 2023-08-16
Maintenance/move folder structure

RuanJohn:
## What?
Moved the core logic from `jax_distribution` to `mava`. 
## Extra
* Added `multirun` to the `.gitignore` for running `hydra` experiments. 
* Pinned the version of `flax` to `0.6.11` to avoid issues with recurrent networks. 

## Note
Please run all systems in `mava/systems` once before approving this PR. Ideally in a fresh conda env made by running 
```python
conda create -n test-moved-files python=3.9 -y
conda activate test-moved-files
pip install -e .[dev]
```
Please also check that the pre-commits install fine by running 
```
pre-commit install
pre-commit run --all-files --verbose 
```


mava/utils/logger_tools.py
```diff
@@ -36,12 +37,8 @@ def __init__(self, console_logger: logging.Logger) -> None:
         self.use_tb = False
         self.use_sacred = False
 
-        self.tb_logger = None
-        self.sacred_run_dict = None
-        self.sacred_info = None
-
         # defaultdict is used to overcome the problem of missing keys when logging to sacred.
-        self.stats = defaultdict(lambda: [])
+        self.stats: Dict = defaultdict(lambda: [])
```
sash-a, line 41: Also import Dict

```suggestion
        self.stats: Dict[str, List[float]] = defaultdict(lambda: [])
```

mava/logger.py
```diff
@@ -115,7 +115,7 @@ def log(
     return log
 
 
-def logger_setup(_run: Run, config: Dict, _log: SacredLogger):
+def logger_setup(_run: Run, config: Dict, _log: SacredLogger) -> Callable:
```
sash-a, line 118: Would be nice in future to define a type for this and some other methods throughout the code so it's not just `Callable`, but rather `Callable[int, float, etc]`

sash-a (approved):
Just a small thing otherwise LGTM

PR #915, opened 2023-08-17
feat: changed logging path

RuanJohn:
## What?
Previously sacred logs and tensorboard were getting stored in different directories due to using `os.path.dirname` at different levels in the `mava` directory. This PR standardises storing logs in the base directory and enables for a unified `base_exp_path` to be passed to both the tensorboard and sacred loggers. This has further benefits for future running on the internal cluster.

### DIFF
--- a/mava/configs/default.yaml
+++ b/mava/configs/default.yaml
@@ -42,4 +42,4 @@ absolute_metric: True # Whether the absolute metric should be computed. For more
 use_tf: True
 use_sacred: True
 base_exp_path: results
-system_name: replace
+system_name: ~  # this is manually set inside each file depending on which system is run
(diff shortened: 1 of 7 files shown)


mava/configs/default.yaml
```diff
@@ -39,6 +39,7 @@ absolute_metric: True # Whether the absolute metric should be computed. For more
   # on the absolute metric please see: https://arxiv.org/abs/2209.10485
 
 # --- Logging options ---
-use_tf: False
+use_tf: True
 use_sacred: True
-base_exp_path: sacred
+base_exp_path: results
+system_name: replace
```
sash-a, line 45:
```suggestion
system_name: ~  # this is manually set inside each file depending on which system is run
```

sash-a (approved):

PR #916, opened 2023-08-18
Feat/quickstart notebook

OmaymaMahjoub:
## What?
Quickstart notebook.
## Why?
This notebook will make it easy for the users to use and understand Mava
## Extra?
The `open in colab` click won't work until we merge this to pure-jax-mava branch


sash-a (approved):
:1st_place_medal:

PR #917, opened 2023-08-18
Feat/jax mava dockerfile

RuanJohn:
## What?
A new simplified `Dockerfile` and `Makefile`. The `Dockerfile` has been tested on the cluster and runs fine. 
## Extra
Updated Detailed Install and Contributing documentation.


DETAILED_INSTALL.md
```diff
@@ -2,251 +2,79 @@
...
-```
-
-* For windows, replace `$(pwd)` with `$(curdir)`.
-
-* You can replace the example with your custom python file.
-
-#### Building the image yourself
-
-1. Build the correct docker image using the `make` command:
+1. Build the docker image using the `make` command:
```
sash-a, line 7:
```suggestion
1. Build the docker image using the `make build` command:
```

DETAILED_INSTALL.md
sash-a, line 1: Should we remove all the windows stuff from here? I really don't want to support that :joy:

DETAILED_INSTALL.md
```diff
@@ -2,251 +2,79 @@
...
     ```
 
-    1.1  To install the core libraries, including [Reverb](https://github.com/deepmind/reverb) - our storage dataset , Tensorflow and [Launchpad](https://github.com/deepmind/launchpad) - for distributed agent support :
-
-    * Install swig for box2d:
-
-    ```bash
-    sudo apt-get install swig -y
-    ```
+    1.1  To install the core libraries, please run:
```
sash-a, line 45: I'd remove this whole section and just put the `pip install .` inside the bigger block

Dockerfile
```diff
@@ -1,89 +1,45 @@
...
-# PettingZoo Atari envs
-RUN apt-get update
-RUN apt-get install ffmpeg libsm6 libxext6  -y
-RUN apt-get install -y unrar-free
-RUN pip install autorom
-RUN AutoROM -v
-##########################################################
+RUN echo "Installing requirements..."
+RUN pip install --quiet --upgrade pip setuptools wheel &&  \
+    pip install -e .
```
sash-a, line 34: Not really sure what's best here, maybe we should copy jumanji, but what about:
```suggestion
    pip install -r requirements/requirements.txt
```

TROUBLESHOOTING.md
sash-a, line 1: I don't think we need this file until we find something specific to mava that happens often that people ask about :+1:

sash-a (changes requested):
LSK :rice_cracker: 

(Don't mind if you leave out the TPU stuff :smile:)

TROUBLESHOOTING.md
RuanJohn: I also thought about this initially but then the `make: 'build' is up to date.` issue already happened for me again locally, so I think it is worth keeping for now. Happy to remove the Docker not using GPU issue.

CONTRIBUTING.md
```diff
@@ -17,18 +17,18 @@ again.
...
+pip install -e .[dev]
 ```
 
-## Installing Pre-Commit Hooks and Testing Dependencies
+## Installing Pre-Commit Hooks
 
-Install the pre-commit hooks and testing dependencies:
+Install the pre-commit hooks:
 ```bash
-pip install .[testing_formatting]
```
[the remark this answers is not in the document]
RuanJohn: Above it does this as the first step? So here some should only have to install the `pre-commit` hooks.

DETAILED_INSTALL.md
RuanJohn: I think we can keep it as is? Since we say `make build` a few steps down?

DETAILED_INSTALL.md
RuanJohn: Agree, done.

Dockerfile
RuanJohn: I tried this, but then it complains that `mava` isn't a module. I think this works on zathura since jumanji is added in `requirements.txt` there.

Dockerfile
```diff
@@ -1,89 +1,45 @@
...
-RUN pip install .[sc2]
-# We use the pz wrapper for smac
-RUN pip install .[pz]
-ENV SC2PATH /home/app/mava/3rdparty/StarCraftII
-##########################################################
+# Need to use specific cuda versions for jax
+ARG USE_CUDA=true
+RUN if [ "$USE_CUDA" = true ] ; \
+    then pip install "jax[cuda11_pip]<=0.4.13" -f "https://storage.googleapis.com/jax-releases/jax_cuda_releases.html" ; \
+    fi
```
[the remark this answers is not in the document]
RuanJohn: To be able to run on a TPU we would actually need to have a completely separate docker file since running on the TPU won't work with the CUDA base image we are using here. A possible solution would be to have a `Dockerfile.gpu` and `Dockerfile.tpu` but not very many people have got TPU access. So I think it is sufficient to only support GPUs for now.

Makefile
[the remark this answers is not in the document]
RuanJohn: Please see above for why not.

.dockerignore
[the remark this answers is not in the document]
RuanJohn: Good point :sweat_smile: 
Done!

Dockerfile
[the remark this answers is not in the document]
sash-a: Great point

TROUBLESHOOTING.md
sash-a: I disagree with this, this issue isn't really related to mava, it's a `make` issue, so I don't think we need to have troubleshooting stuff for this

sash-a (approved):
Just remove the line about windows in `DETAILED_INSTALL.md` and see my reply about `TROUBLESHOOTING.md`. Otherwise I'm happy

TROUBLESHOOTING.md
RuanJohn: I think it is quite nice to keep it in since it is a common thing someone might run into? 
If you feel really strongly I am happy to remove it, but I have found this helpful in the past.

DETAILED_INSTALL.md
```diff
@@ -2,251 +2,72 @@
...
-#### Building the image yourself
-
-1. Build the correct docker image using the `make` command:
+1. Build the docker image using the `make` command:
 
     For Windows, before the docker image build, we recommend to first install the package manager [chocolatey](https://chocolatey.org/install) and run (to install make):
 
     ```bash
     choco install make
     ```
```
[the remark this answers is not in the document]
RuanJohn: Removed :)

PR #918, opened 2023-08-18
Feat/vmap seeds

OmaymaMahjoub:
## What?
Add the option of vmapping over multiple seeds.
## Why?
We believe this will help in speeding up running experiments compared to run the different seeds sequentially 
## How?
After pmap over the devices, inside each device we firstly vmap over the different seeds.
## Extra
We need to double check that running different seeds simultaneously won't impact the step per second value (sps)
Merge #920 before reviewing this.


mava/logger.py
```diff
@@ -102,19 +115,52 @@ def log(
...
+                f"{Fore.GREEN}{Style.BRIGHT}EVALUATOR{suffix}: {log_string}{Style.RESET_ALL}"
             )
 
         return float(np.mean(episodes_return))
 
-    return log
+    def save_json_by_seed() -> None:
+        """Split the current json file by seed into separate files."""
+        # Make sure the full values are recorded in the JSON file.
+        time.sleep(10)
```
sash-a, line 130: I think this should rather be pulled out of this function otherwise people might try to call this function and expect it to happen instantly. So rather

```python
# Make sure the full values are recorded in the JSON file.
time.sleep(10)
logger.save_json_by_seed()
```

mava/logger.py
```diff
@@ -102,19 +115,52 @@ def log(
...
         else:
             logger.console_logger.info(
-                f"{Fore.GREEN}{Style.BRIGHT}EVALUATOR: {log_string}{Style.RESET_ALL}"
+                f"{Fore.GREEN}{Style.BRIGHT}EVALUATOR{suffix}: {log_string}{Style.RESET_ALL}"
             )
 
         return float(np.mean(episodes_return))
 
-    return log
+    def save_json_by_seed() -> None:
```
sash-a, line 127: A thought on a possibly easier way to do this: instead of having this method, just creating different loggers for each seed. So now instead of a single logger you could have `Dict[int, Logger]` and then when logging you can keep it all the same just instead of `logger.xyz()` you could do `logger[seed].xyz()`.

I know this isn't a trivial change and so maybe we make an issue and address it next week?

mava/logger.py
```diff
@@ -102,19 +115,52 @@ def log(
...
+            output_file_path = f"{_run.observers[0].dir}/metrics_seed_{seed_num}.json"
+            with open(output_file_path, "w") as output_file:
+                json.dump(metrics, output_file, indent=2)
+
+    def stop_logger() -> None:
+        """Stop the logger and save the metrics for different seeds."""
+        if logger.use_neptune:
+            logger.neptune_logger.stop()
+        if (len(config["seeds"]) > 1) and config["split_json_by_seed"]:
+            save_json_by_seed()
```
sash-a, line 158:
```suggestion
            # Make sure the full values are recorded in the JSON file.
            time.sleep(10)
            save_json_by_seed()
```

mava/logger.py
```diff
@@ -102,19 +115,52 @@ def log(
...
+                seed_data[seed_num][metric_name] = value
+
+        # Create separate JSON files for each seed's metrics
+        for seed_num, metrics in seed_data.items():
+            output_file_path = f"{_run.observers[0].dir}/metrics_seed_{seed_num}.json"
+            with open(output_file_path, "w") as output_file:
+                json.dump(metrics, output_file, indent=2)
+
+    def stop_logger() -> None:
+        """Stop the logger and save the metrics for different seeds."""
```
sash-a, line 154:
```suggestion
        """Stop the logger and save the metrics for different seeds.
        
        Note: this method stops waits for 10 seconds to be sure all logging 
        has been written to disk
        """
```

sash-a (commented):
I see I was supposed to do #920 first, just leaving the comments here for now

mava/configs/default.yaml
```diff
@@ -3,22 +3,24 @@
 # --- Experiment params ---
 defaults:
   - _self_
-  - rware_scenario: tiny-4ag-easy
+  - rware_scenario: tiny-2ag
```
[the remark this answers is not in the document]
OmaymaMahjoub: happy to revert it the default in jumanji is small-4ag-easy

mava/configs/default.yaml
```diff
@@ -39,7 +41,12 @@ absolute_metric: True # Whether the absolute metric should be computed. For more
   # on the absolute metric please see: https://arxiv.org/abs/2209.10485
 
 # --- Logging options ---
-use_tf: True
+use_tf: False
 use_sacred: True
+split_json_by_seed: True # Split JSON logs by seed after the full experiment has run.
```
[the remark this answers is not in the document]
OmaymaMahjoub: I thought about making it optional because postprocessing can take some time if you record for many evaluations and if someone don't care about splitting can skip that time (we also made a warning that will shows up if this is false so people be aware of it)

mava/logger.py
OmaymaMahjoub: the issue of moving it outside will need to check with the flag that we will call the function else will be useless 10 seconds of waiting do you have another suggestion? else maybe we can just shows a print related to it "waiting for 10 seconds..."

mava/logger.py
OmaymaMahjoub: we wanted to do that but with creating ex.run in the main sacred won't allow having multiple loggers it will be just multiple references to the same instance :cry:

mava/logger.py
OmaymaMahjoub: oh you mean making the time sleep after this condition yeah sure I will update it

mava/logger.py
OmaymaMahjoub: oh I saw the other review point yeah we can add time.sleep in the stop function

mava/configs/default.yaml
```diff
@@ -39,7 +41,12 @@ absolute_metric: True # Whether the absolute metric should be computed. For more
...
 base_exp_path: results
 system_name: ~  # this is manually set inside each file depending on which system is run
+# Neptune logging
+use_neptune: False
+neptune_tag: test_mava
```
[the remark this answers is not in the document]
OmaymaMahjoub: I guess we need to update this on the other PR

sash-a:
Seemed to be slower than just running sequentially, likely an issue where we were oversubscribing the cores on the accelerator. So closing it for now :shrug:

PR #926, opened 2023-08-28
feat: log environment timestep to x-axis on neptune

RuanJohn:
## What?
Log timesteps to the x-axis on Neptune.
## Why?
It makes plots on Neptune far easier to interpret and also allows for comparing sample efficiency across experiments that ran for different total numbers of timesteps. 

### After the changes:
Screenshot from 2023-08-28 12-20-29

### Before the changes: 
Screenshot from 2023-08-28 12-22-59


OmaymaMahjoub (approved):

sash-a (approved):
:ship:

PR #927, opened 2023-08-28
Fix: fixing global-obs shape in case of no agent-id 

OmaymaMahjoub:
## What?
Fixing the shape of global observation in the case of centralised training and no agent id.
## How?
Duplicate the global obs for each agent as we vmap over the full observation

### DIFF
--- a/mava/wrappers/jumanji.py
+++ b/mava/wrappers/jumanji.py
@@ -107,9 +107,6 @@ class AgentIDWrapper(Wrapper):
         agent_ids = jnp.eye(num_agents)
         new_agents_view = jnp.concatenate([agent_ids, timestep.observation.agents_view], axis=-1)
 
-        # we really should rather do this:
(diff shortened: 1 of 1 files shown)


mava/wrappers/jumanji.py
```diff
@@ -112,8 +112,9 @@ def _add_agent_ids(
         # just need to check that this doesn't affect timing
```
sash-a, line 112: Can you remove these comments above here. I had another look and it's not possible to do this unfortunately

sash-a (approved):
:shipit:

RuanJohn (dismissed):
Looks good to me! 
Good catch. Thank you @OmaymaMahjoub :fire:

RuanJohn (approved):
Thank you :raised_hands:

sash-a (approved):
:ship:

PR #928, opened 2023-08-28
feat: Split Net in the Quickstart notebook

OmaymaMahjoub:
## What?
Split the actor and critic network and tune the model
## Why?
By splitting the net and tune the mode we reached similar performance of 150M timesteps in just 20M


sash-a (dismissed):
:ship:

sash-a (approved):
:ship: :ship:

RuanJohn (approved):

PR #934, opened 2023-08-30
feat: update neptune

sash-a:
* Unpinned neptune version
* Removed old `NeptuneLogger` class
* Use sacred integration for logging to neptune - so neptune has been remove from `Logger` instead it is added to sacred when sacred is setup
* Refactored the common logging stuff from all the files into a method: `get_sacred_exp`
* Removed the `stop_logging` method because sacred should stop the neptune logger now
* Converted the hydra config to a dict right away, this lets us remove the `copy_config` method, plays nicer with typing and neptune doesn't complain

Note to run this locally you need to uninstall neptune-client before you update neptune!

```
pip uninstall neptune-client
pip install neptune[sacred]
```

Closes #930

PR #936, opened 2023-09-04
feat: use rlax

### LINKED ISSUE #929: Use rlax
### Feature
Use `rlax` where we can, so that we don't have to implement and test common methods.

### Proposal
As far as I can see we can use it for policy/critic loss and GAE.

### Benchmarking
* Needs to match the perfromance of the current develop branch.

sash-a:
## What?
Uses `rlax` for gae and policy loss clipping

Depends on #934 
Closes #929

### DIFF
--- a/mava/configs/default.yaml
+++ b/mava/configs/default.yaml
@@ -3,14 +3,14 @@
 # --- Experiment params ---
 defaults:
   - _self_
-  - rware_scenario: tiny-2ag
+  - rware_scenario: small-4ag
 
 # --- Environment ---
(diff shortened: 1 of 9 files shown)


sash-a:
Results for ff-MAPPO on tiny-2ag, tiny-4ag and small-4ag. All runs have 4 seeds, longer lines are new rlax runs and shorter lines are old runs. Performance seems the same if not slightly better

image

sash-a:
Unfortunately this seems to be a bit slower than the manual implementations :cry:

PR #938, opened 2023-09-15
Feat/add recurrent chunks

RuanJohn:
## What?
Add support for splitting environment rollouts into smaller chunks for recurrent systems. This has two benefits: 
* Improves wallclock time performance 
* Improves system stability


mava/systems/rec_ippo_rware.py
```diff
@@ -445,12 +452,27 @@ def _critic_loss_fn(
...
+            )
+            batch = jax.tree_util.tree_map(
+                lambda x: x.reshape(
+                    x.shape[0] // config["recurrent_chunk_size"],
+                    config["recurrent_chunk_size"],
+                    -1,
+                    *x.shape[2:],
+                ).reshape(config["recurrent_chunk_size"], -1, *x.shape[2:]),
+                batch,
+            )
```
OmaymaMahjoub, line 471:
```suggestion
            batch = jax.tree_util.tree_map(
                lambda x: x.reshape(config["recurrent_chunk_size"], (x.shape[0] // config["recurrent_chunk_size"])*x.shape[1], *x.shape[2:]),
                batch,
            )
```

mava/configs/system/rec_mappo.yaml
```diff
@@ -20,3 +20,7 @@ clip_eps: 0.2 # Clipping value for PPO updates and value function.
 ent_coef: 0.01 # Entropy regularisation term for loss function.
 vf_coef: 0.5 # Critic weight in
 max_grad_norm: 0.5 # Maximum norm of the gradients for a weight update.
+
+# --- Recurrent hyperparameters ---
+recurrent_chunk_size: ~ # The size of the chunks in which the recurrent sequences are divided during the training process.
+skip_hstates_update: False # Whether to skip updating the hidden states during training.
```
OmaymaMahjoub, line 26: same as above

mava/systems/rec_ippo.py
```diff
@@ -317,25 +326,31 @@ def _update_minibatch(train_state: Tuple, batch_info: Tuple) -> Tuple:
...
                 ) = batch_info
 
+                if config["system"]["skip_hstates_update"]:
+                    traj_batch = traj_batch._replace(
+                        policy_hidden_state=jnp.zeros_like(traj_batch.policy_hidden_state)
+                    )
+                    traj_batch = traj_batch._replace(
+                        critic_hidden_state=jnp.zeros_like(traj_batch.critic_hidden_state)
+                    )
+
```
OmaymaMahjoub, line 341: if we can remove this

mava/systems/rec_mappo.py
```diff
@@ -310,24 +318,30 @@ def _update_minibatch(train_state: Tuple, batch_info: Tuple) -> Tuple:
...
                 ) = batch_info
 
+                if config["system"]["skip_hstates_update"]:
+                    traj_batch = traj_batch._replace(
+                        policy_hidden_state=jnp.zeros_like(traj_batch.policy_hidden_state)
+                    )
+                    traj_batch = traj_batch._replace(
+                        critic_hidden_state=jnp.zeros_like(traj_batch.critic_hidden_state)
+                    )
+
```
OmaymaMahjoub, line 333: if you can remove this and create an issue for future PR about this

OmaymaMahjoub (dismissed):
Thanks @RuanJohn, just I requested some minor changes :pray:

mava/systems/rec_ippo.py
```diff
@@ -349,12 +355,28 @@ def _critic_loss_fn(
...
             key, shuffle_key = jax.random.split(key)
 
             # SHUFFLE MINIBATCHES
-            permutation = jax.random.permutation(shuffle_key, config.arch.num_envs)
-            batch = (init_policy_hstate, init_critic_hstate, traj_batch, advantages, targets)
+            batch = (
+                traj_batch,
+                advantages,
+                targets,
+            )
```
sash-a: nit:
```suggestion
            batch = traj_batch, advantages,targets,
```

mava/configs/system/rec_ippo.yaml
```diff
@@ -20,3 +20,7 @@ clip_eps: 0.2 # Clipping value for PPO updates and value function.
 ent_coef: 0.01 # Entropy regularisation term for loss function.
 vf_coef: 0.5 # Critic weight in
 max_grad_norm: 0.5 # Maximum norm of the gradients for a weight update.
+
+# --- Recurrent hyperparameters ---
+recurrent_chunk_size: ~ # The size of the chunks in which the recurrent sequences are divided during the training process.
```
sash-a, line 25: Is `None` a good default here. Like the comments but we should also set it to a good value

mava/types.py
```diff
@@ -96,6 +96,20 @@ class PPOTransition(NamedTuple):
     info: Dict
 
 
+class RNNPPOTransition(NamedTuple):
```
sash-a: Nit: too many capitals, just a suggestion, but at least for RNN->Rnn :joy: 
```suggestion
class RnnPpoTransition(NamedTuple):
```

mava/types.py
```diff
@@ -96,6 +96,20 @@ class PPOTransition(NamedTuple):
...
+    """Transition tuple for PPO."""
+
+    done: Done
+    action: Action
+    value: Value
+    reward: chex.Array
+    log_prob: chex.Array
+    obs: chex.Array
+    policy_hidden_state: HiddenState
+    critic_hidden_state: HiddenState
```
sash-a: I think this should be `HiddenStates`, HiddenStates above this so you can use it here

mava/systems/rec_mappo.py
```diff
@@ -584,6 +601,14 @@ def run_experiment(_config: DictConfig) -> None:
     """Runs experiment."""
     config = copy.deepcopy(_config)
 
+    # Set recurrent chunk size.
```
sash-a, line 599: Just a thought, should we require the user to set it so we don't have to have this check? I don't mind either way really

sash-a (changes requested):
Happy with this except for minor changes. Have we got performance comparisons?

mava/configs/system/rec_ippo.yaml
RuanJohn: Yes, we set it to None on purpose so that it recovers the current default Mava behaviour which is to have only a single `chunk` with the full rollout.

mava/types.py
RuanJohn: I prefer the capitals - keen to hear thoughts of others.

mava/systems/rec_mappo.py
RuanJohn: I think what we have at the moment is quite nice because it is easy to recover the current behaviour.

mava/types.py
OmaymaMahjoub: I agree with @RuanJohn

mava/types.py
OmaymaMahjoub: Nice catch!

sash-a (approved):
:shipit:

OmaymaMahjoub (approved):

PR #940, opened 2023-09-15
feat: type callables

### LINKED ISSUE #933: Type callables
### Feature
We have untyped callables throughout the codebase e.g [here](https://github.com/instadeepai/Mava/blob/develop/mava/systems/ff_ippo_rware.py#L103). We should type them and put the types inside `types.py`

sash-a:
## What?
We had a lot of empty `Callable`'s in mava before this e.g `def f(x: Callable) -> Callable`

## Why?
These are bad, types are good

Closes #933
Depends on #936 and #934


mava/types.py
```diff
@@ -12,23 +12,35 @@
...
 
+Action: TypeAlias = chex.Array
+Value: TypeAlias = chex.Array
+Done: TypeAlias = chex.Array
+HiddenState: TypeAlias = chex.Array
+
+# Can't know the exact type of State or Timestep.
+# Is there a better way to do this?
+State: TypeAlias = Any
+Observation: TypeAlias = Any
```
OmaymaMahjoub, line 34: Maybe we can just import State and Observation directly from jumanji `from jumanji.environments.routing.robot_warehouse import Observation, State
`

sash-a: So the reason I don't do this is because we don't want to limit ourselves to rware's obs and state, we want to allow for any type of state and obs

OmaymaMahjoub: Oh I see, maybe we can define a type in jumanji wrapper py and just call it here, where State in the wrapper is the union of states of different envs

mava/types.py
```diff
@@ -94,18 +106,34 @@ class RNNEvalState(NamedTuple):
...
-    step_count_: chex.Numeric = None
-    return_: chex.Numeric = None
+    hstate: HiddenState
+
+
+MavaState = TypeVar("MavaState", LearnerState, RNNLearnerState, EvalState, RNNEvalState)
 
 
-class ExperimentOutput(NamedTuple):
+class ExperimentOutput(NamedTuple, Generic[MavaState]):
```
OmaymaMahjoub, line 118: I think we need to double check this, I don't know if the error happen on my local machine only but this one give the following error {TypeError: Multiple inheritance with NamedTuple is not supported{

mava/types.py
sash-a: But then we have to remember that every time we add a new env we must also update the type.

I'd like there to be something better, but I don't think there is a better option. Ideally I'd want something like this

```python
class Observation:
  any_other_attribs
  ....
  action_mask: chex.Array
  step_count: chex.Array
```

Then we'd get type hints for `action_mask` and `step_count`, but it would also allow for other attributes, just I don't know if this is possible in python :shrug:

OmaymaMahjoub: it would be amazing if we had all the other observation variables under one attribute but yeah I get the issue with this :/ maybe keeping it to any is better for now and maybe create an issue for the future to work on it

mava/types.py
sash-a: Not sure but that seems to be a python 3.8 issue see [here](https://github.com/JustAnotherArchivist/snscrape/issues/111)

OmaymaMahjoub: They mentioned that they fixed it for python 3.9 however I am using 3.9 and still have it :(

sash-a: hmmmm I'm not sure then :thinking:

OmaymaMahjoub: I will try to look if there is another option to keep inheriting from both

mava/types.py
OmaymaMahjoub: @sash-a  if you can remove the comment from this one and maybe create an issue related to it (that we will try to make a PR for it once we support more envs)

OmaymaMahjoub (approved):

PR #942, opened 2023-09-21
feat: logging steps per second

sash-a:
## What?
Log the steps per second to tensorboard, neptune and sacred
## Why?
We were only printing the steps per second, would be nice to have it on neptune
## How?
Added steps per second to metrics dict
## Extra
* Removed `TimeIt` since we aren't using that anymore.
* Changed logging prefix to be e.g "trainer/" instead of "Trainer_"
* Added prefix for evalutor (it used to be blank)
* Changed `timesteps_per_training` to `steps_per_rollout`


mava/systems/ff_ippo_rware.py
```diff
@@ -558,9 +559,11 @@ def run_experiment(_run: run.Run, _config: Dict, _log: SacredLogger) -> None:
...
 
         # Log the results of the evaluation.
+        elapsed_time = time.time() - start_time
+        evaluator_output.episodes_info["steps_per_second"] = steps_per_rollout / elapsed_time
         episode_return = log(
             metrics=evaluator_output,
-            t_env=timesteps_per_training * (i + 1),
+            t_env=steps_per_rollout * (i + 1),
         )
         if config["absolute_metric"] and max_episode_return <= episode_return:
```
OmaymaMahjoub, line 568: We need to add `evaluator_output.episodes_info["steps_per_second"] = steps_per_rollout / elapsed_time` to the absolute metric else it will cause an error of not finding steps_per_seconds within it

OmaymaMahjoub (changes requested):
Looks good, just a minor fix related to logging the absolute metric!

mava/systems/ff_ippo_rware.py
sash-a: Great spot!

RuanJohn (approved):
Very nice that we can log this now. Thanks @sash-a :fire:

mava/systems/ff_ippo_rware.py
```diff
@@ -517,7 +517,7 @@ def run_experiment(_run: run.Run, _config: Dict, _log: SacredLogger) -> None:
     # Calculate total timesteps.
     n_devices = len(jax.devices())
     config["num_updates_per_eval"] = config["num_updates"] // config["num_evaluation"]
-    timesteps_per_training = (
+    steps_per_rollout = (
```
OmaymaMahjoub, line 520: It will be amazing if we can also print at the beginning of the training the expected total timesteps (as it is different depending on device num...)

mava/logger.py
```diff
@@ -74,6 +75,8 @@ def log(
                 float(np.mean(episodes_length)),
                 t_env,
             )
+            logger.log_stat(prefix.lower() + "steps_per_second", steps_per_second, t_env)
+
             if trainer_metric:
```
OmaymaMahjoub, line 80: What do you think about adding a prefix to these info? I'm not sure if it would be beneficial.

OmaymaMahjoub (commented):
Overall it looks good to me, Thanks for the PR! 
Maybe another minor request is to update the quickstart notebook to have `steps_per_rollout` instead of `timesteps_per_training`

mava/logger.py
```diff
@@ -74,6 +75,8 @@ def log(
...
                 logger.log_stat("total_loss", float(np.mean(total_loss)), t_env)
                 logger.log_stat("value_loss", float(np.mean(value_loss)), t_env)
```
OmaymaMahjoub, line 82: What do you think about adding a prefix to these info? I'm not sure if it would be beneficial.

sash-a: What would the prefix be?

OmaymaMahjoub: maybe "trainer/" even though these only measured for the trainer :sweat_smile: ?

mava/systems/ff_ippo_rware.py
OmaymaMahjoub: forget about this as it is solved on seperate config PR

OmaymaMahjoub (approved):
LGTM! Thanks @sash-a

mava/logger.py
```diff
@@ -99,7 +94,8 @@ def log(
             f"Max Episode Return {float(np.max(episodes_return)):.3f} | "
             f"Mean Episode Length {float(np.mean(episodes_length)):.3f} | "
             f"Std Episode Length {float(np.std(episodes_length)):.3f} | "
-            f"Max Episode Length {float(np.max(episodes_length)):.3f}"
+            f"Max Episode Length {float(np.max(episodes_length)):.3f} | "
+            f"Steps Per Second {steps_per_second:.2e}"
```
RuanJohn, line 98:
```suggestion
            f"Steps Per Second {steps_per_second:.2e} "
```

RuanJohn (changes requested):
Thanks @sash-a, very minor suggestion. 
But also a detail, when I tried to run this locally the logging to Neptune seemed to not be working. Can you please double check that it works on your side?

RuanJohn (approved):
:fire:

OmaymaMahjoub (approved):
🥇

PR #943, opened 2023-10-26
feat(lbf): generalise wrapper and logger

WiemKhlifi:
## What?
I've changed the wrapper, logger, logger_utils and default config files
## Why?
To support the features that will be added to support LBF later.


mava/wrappers/jumanji.py
```diff
@@ -236,16 +261,19 @@ def step(self, state: State, action: chex.Array) -> Tuple[State, TimeStep]:
             global_state=global_state,
             step_count=jnp.repeat(timestep.observation.step_count, self._env.num_agents),
         )
+        timestep.reward = jnp.sum(timestep.reward)
```
sash-a, line 264: I don't think this is correct, this makes it always a joint reward, what I think we need to do is create a `MARLRewardWrapper` that takes any environment with only 1 reward and repeats it `num_agents` number of times and then we must just handle this correctly in the loss functions and always wrap environments such as RWARE with this wrapper. 

@RuanJohn  and @OmaymaMahjoub your opinions are needed here

mava/wrappers/jumanji.py
```diff
@@ -236,16 +261,19 @@ def step(self, state: State, action: chex.Array) -> Tuple[State, TimeStep]:
...
     def observation_spec(self) -> specs.Spec[ObservationGlobalState]:
         """Specification of the observation of the `RobotWarehouse` environment."""
+        num_actions = int(self._env.action_spec().num_values[0])
 
         agents_view = specs.Array(
             (self._env.num_agents, self._env.num_obs_features), jnp.int32, "agents_view"
         )
         action_mask = specs.BoundedArray(
-            (self._env.num_agents, 5), bool, False, True, "action_mask"
+            (self._env.num_agents, num_actions), bool, False, True, "action_mask"
```
sash-a, line 276: Nice change, but I think this catches some edge cases where we have multi-discrete action shapes

```suggestion
            self._env.action_spec().shape, bool, False, True, "action_mask"
```

mava/wrappers/jumanji.py
```diff
@@ -223,6 +246,8 @@ def reset(self, key: chex.PRNGKey) -> Tuple[State, TimeStep]:
             global_state=global_state,
             step_count=jnp.repeat(timestep.observation.step_count, self._env.num_agents),
         )
+        timestep.reward = jnp.sum(timestep.reward)
```
sash-a, line 249: Same here

mava/wrappers/jumanji.py
```diff
@@ -190,6 +211,8 @@ def step(self, state: State, action: chex.Array) -> Tuple[State, TimeStep]:
             action_mask=timestep.observation.action_mask,
             step_count=jnp.repeat(timestep.observation.step_count, self._env.num_agents),
         )
+        timestep.reward = jnp.sum(timestep.reward)
```
sash-a, line 214: Same here

mava/wrappers/jumanji.py
```diff
@@ -180,6 +199,8 @@ def reset(self, key: chex.PRNGKey) -> Tuple[State, TimeStep]:
...
```
sash-a, line 202: Same here

mava/logger.py
```diff
@@ -90,6 +90,18 @@ def log(
...
+        if trainer_metric and "LevelBasedForaging" in config["env_name"]:
+            num_eaten = jnp.ravel(episodes_info["num_eaten"])
+            percent_eaten = jnp.ravel(episodes_info["percent_eaten"])
+            logger.log_stat("mean_num_eaten", float(np.mean(num_eaten)), t_env)
+            logger.log_stat("mean_percent_eaten", float(np.mean(percent_eaten)), t_env)
+            log_string = (
+                log_string
+                + f" | Num Eaten {float(np.mean(num_eaten)):.3f} | "
+                + f"Percent Eaten {float(np.mean(percent_eaten)):.3f}"
+            )
```
sash-a, line 103: I really like this! 

However it should be done in the env and then we can grab it and log it here so that we don't have to make if statements for each different env. See how we add these kind of statistics in connector [here](https://github.com/instadeepai/jumanji/blob/main/jumanji/environments/routing/connector/env.py#L272-L283) and [here](https://github.com/instadeepai/jumanji/blob/main/jumanji/environments/routing/connector/env.py#L181-L194)

So then we could change add extras to our episode metrics and always log extras.

mava/configs/default.yaml
```diff
@@ -3,10 +3,12 @@
 # --- Experiment params ---
 defaults:
   - _self_
+# Select the wanted scenario for the corresponding environment.
   - rware_scenario: tiny-2ag
+  - lbf_scenario: Foraging-10x10-3p-3f # [Foraging-10x10-3p-3f, Foraging-grid-10x10-3p-3f, Foraging-8x8-2p-2f-coop]
 
 # --- Environment ---
-env_name: RobotWarehouse-v0
+env_name: RobotWarehouse-v0  # [RobotWarehouse-v0, LevelBasedForaging-v0]
```
sash-a, line 11:
```suggestion
env_name: RobotWarehouse-v0
```

mava/configs/default.yaml
```diff
@@ -46,6 +48,6 @@ base_exp_path: results
 system_name: ~  # this is manually set inside each file depending on which system is run
 # Neptune logging
 use_neptune: False
-neptune_project: Instadeep/Mava
-neptune_tag: [rware]
+neptune_project: InstaDeep/Mava
+neptune_tag: [Foraging-10x10-3p-3f, ff_ippo, 16_envs]
```
sash-a, line 52:
```suggestion
neptune_tag: [rware]
```

mava/configs/default.yaml
```diff
@@ -3,10 +3,12 @@
 # --- Experiment params ---
 defaults:
   - _self_
+# Select the wanted scenario for the corresponding environment.
```
sash-a, line 6: nitpick

```suggestion
  # Select the wanted scenario for the corresponding environment.
```

sash-a (changes requested):
Thanks @WiemKhlifi I really like this PR! Only major concern is about the `jnp.sum(rewards)` in the wrapper :fire:

mava/wrappers/jumanji.py
```diff
@@ -204,7 +227,7 @@ def observation_spec(self) -> specs.Spec[Observation]:
         return self._env.observation_spec().replace(step_count=step_count)
 
 
-class RwareMultiAgentWithGlobalStateWrapper(Wrapper):
+class MultiAgentWithGlobalStateWrapper(Wrapper):
```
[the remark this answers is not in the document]
WiemKhlifi: Didn't pay attention to that thank you!

mava/utils/logger_tools.py
```diff
@@ -140,12 +140,20 @@ def get_sacred_exp(cfg: Dict, system_name: str) -> Experiment:
...
+            f"{logger_type}/{config['system_name']}/{config['env_name']}/"
+            + f"{config['rware_scenario']['task_name']}/envs_{config['num_envs']}/"
+            + f"seed_{config['seed']}"
+        )
+    else:
+        exp_path = (
+            f"{logger_type}/{config['system_name']}/{config['env_name']}/"
+            + f"{config['lbf_scenario']['task_name']}/envs_{config['num_envs']}/"
+            + f"seed_{config['seed']}"
+        )
```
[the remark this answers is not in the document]
WiemKhlifi: Good idea! for this, we'll  put all the config files of each scenario in one folder called "**scenario**" for easy access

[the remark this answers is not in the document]
WiemKhlifi: changed also the name of scenario config file to start with the name of the environment

mava/wrappers/jumanji.py
sash-a: hmmmm, weird, ok then you can leave it as is and if we get an issue in the future we can deal with it then :smile:

mava/wrappers/jumanji.py
WiemKhlifi: This seems like a good idea, we can also use one wrapper for the shared reward and another for the opposite case.
And yes, in this case, we only used the shared reward and added it to get similar results to the benchmark paper where the sum of the individual rewards was used as the shared reward.

mava/logger.py
WiemKhlifi: Done :heavy_check_mark: Please take a look at it :)

mava/wrappers/jumanji.py
```diff
@@ -43,7 +46,8 @@ class Observation(NamedTuple):
     """
 
     agents_view: chex.Array  # (num_agents, num_obs_features)
-    action_mask: chex.Array  # (num_agents, num_actions) num_actions = 5 in RWARE and num_actions = 6 in LBF
+    action_mask: chex.Array  # (num_agents, num_actions)
+    # num_actions = 5 in RWARE and num_actions = 6 in LBF
```
sash-a, line 50: Can just remove this :)

```suggestion
```

sash-a (commented):
Great changes! Just waiting for the other PRs to be approved so we can make the final changes

mava/wrappers/jumanji.py
WiemKhlifi: Yes of course!

PR #944, opened 2023-11-20
chore: unpin jax version

sash-a:
## What?
Unpin the jax version
## Why?
It would be nice to always use the most recent jax
## How?
Change the requirements.txt, had to also switch to distrax master, see https://github.com/google-deepmind/distrax/issues/249


RuanJohn (approved):
:fire:

PR #947, opened 2023-11-21
Maintenance/remove sacred

RuanJohn:
## What?
Removes sacred from Mava
## Why?
It is much cleaner to only use Hydra.  
## Extra
We will add a custom JSON logger in the future which will better support using `marl-eval`.


mava/systems/ff_ippo_rware.py
```diff
@@ -488,11 +485,11 @@ def learner_setup(
     return learn, actor_network, init_learner_state
 
 
-def run_experiment(_run: run.Run, _config: Dict, _log: SacredLogger) -> None:
+def run_experiment(_config: Dict) -> None:
```
sash-a, line 488: Made this method much cleaner!

mava/utils/logger_tools.py
```diff
@@ -13,34 +13,28 @@
...
-        self.console_logger = console_logger
+        self.cfg = cfg
+        self.console_logger = get_python_logger()
 
         self.use_tb = False
-        self.use_sacred = False
-
-        # defaultdict is used to overcome the problem of missing keys when logging to sacred.
-        self.stats: Dict[str, List[Tuple[int, float]]] = defaultdict(lambda: [])
+        self.use_neptune = False
```
sash-a, line 37: Why not grab `use_tb` and `use_neptune` from the config and set it here, then we don't even need to store the config

sash-a (changes requested):
Minor issue, just move the logger setup into the `__init__`

mava/systems/ff_ippo_rware.py
```diff
@@ -621,11 +618,8 @@ def hydra_entry_point(cfg: DictConfig) -> None:
     # Convert config to python dict.
     cfg: Dict = OmegaConf.to_container(cfg, resolve=True)
```
OmaymaMahjoub, line 619: I think we don't need this anymore

sash-a (changes requested):
Very minor things, very happy with this!

mava/utils/logger_tools.py
```diff
@@ -100,48 +83,16 @@ def get_python_logger() -> logging.Logger:
...
-    name = cfg["logger"]["kwargs"]["name"]
     tags = cfg["logger"]["kwargs"]["neptune_tag"]
     project = cfg["logger"]["kwargs"]["neptune_project"]
 
-    run = neptune.init_run(name=name, project=project, tags=tags)
+    run = neptune.init_run(project=project, tags=tags)
 
-    del cfg["logger"]["kwargs"]["neptune_tag"]  # neptune doesn't want lists in run params
-    run["params"] = cfg
+    run["params"] = stringify_unsupported(cfg)
```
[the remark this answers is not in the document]
RuanJohn: For sure, let's set it to config.

sash-a (approved):
Great changes :fire:

mava/systems/ff_ippo_rware.py
RuanJohn: I agree that we should definitely move from a dict to the default `hydra` `.` syntax but this will take quite a bit of admin since all of Mava expects dicts at the moment. Can we make a maintenance issue for updating to this later on?

RuanJohn: I made [this issue](https://github.com/instadeepai/Mava/issues/949) so that we can address this in the future.

OmaymaMahjoub: Sounds great!

OmaymaMahjoub (approved):
:1st_place_medal:

PR #948, opened 2023-11-21
feat: general wrapper framework

sash-a:
## What?
General observation and example rware wrapper which makes it easy to support more jumanji envs in the future :eyes: 
## Why?
To support more jumanji envs
## How?
Create an observation type with known fields. Wrap all new envs so that they return this observation type instead of their own custom observation.

### DIFF
--- a/mava/systems/rec_ippo_rware.py
+++ b/mava/systems/rec_ippo_rware.py
@@ -48,10 +48,11 @@ from mava.types import (
     RecActorApply,
     RecCriticApply,
     RNNLearnerState,
+    RnnObservation,
 )
 from mava.utils.logger_tools import get_sacred_exp
 from mava.utils.timing_utils import TimeIt
-from mava.wrappers.jumanji import AgentIDWrapper, LogWrapper, RwareMultiAgentWrapper
+from mava.wrappers.jumanji import AgentIDWrapper, LogWrapper, RwareWrapper
 
 
 class ScannedRNN(nn.Module):
@@ -91,7 +92,7 @@ class Actor(nn.Module):
     def __call__(
         self,
         policy_hidden_state: chex.Array,
-        observation_done: Tuple[chex.Array, chex.Array],
+        observation_done: RnnObservation,
     ) -> Tuple[chex.Array, distrax.Categorical]:
         """Forward pass."""
         observation, done = observation_done
@@ -130,7 +131,7 @@ class Critic(nn.Module):
     def __call__(
         self,
         critic_hidden_state: Tuple[chex.Array, chex.Array],
-        observation_done: Tuple[chex.Array, chex.Array],
+        observation_done: RnnObservation,
     ) -> Tuple[chex.Array, chex.Array]:
         """Forward pass."""
         observation, done = observation_done
@@ -677,14 +678,14 @@ def run_experiment(_run: run.Run, _config: Dict, _log: SacredLogger) -> None:
     # Create envs
     generator = RandomGenerator(**config["rware_scenario"]["task_config"])
     env = jumanji.make(config["env_name"], generator=generator)
-    env = RwareMultiAgentWrapper(env)
+    env = RwareWrapper(env)
     # Add agent id to observation.
     if config["add_agent_id"]:
         env = AgentIDWrapper(env)
     env = AutoResetWrapper(env)
     env = LogWrapper(env)
     eval_env = jumanji.make(config["env_name"], generator=generator)
-    eval_env = RwareMultiAgentWrapper(eval_env)
+    eval_env = RwareWrapper(eval_env)
     if config["add_agent_id"]:
         eval_env = AgentIDWrapper(eval_env)
 
--- a/mava/systems/rec_mappo_rware.py
+++ b/mava/systems/rec_mappo_rware.py
@@ -48,14 +48,16 @@ from mava.types import (
     RecActorApply,
     RecCriticApply,
     RNNLearnerState,
+    RnnObservation,
 )
 from mava.utils.logger_tools import get_sacred_exp
 from mava.utils.timing_utils import TimeIt
 from mava.wrappers.jumanji import (
     AgentIDWrapper,
+    GlobalStateWrapper,
     LogWrapper,
     ObservationGlobalState,
-    RwareMultiAgentWithGlobalStateWrapper,
+    RwareWrapper,
 )
 
 
@@ -96,7 +98,7 @@ class Actor(nn.Module):
     def __call__(
         self,
         policy_hidden_state: chex.Array,
-        observation_done: Tuple[chex.Array, chex.Array],
+        observation_done: RnnObservation,
     ) -> Tuple[chex.Array, distrax.Categorical]:
         """Forward pass."""
         observation, done = observation_done
@@ -135,7 +137,7 @@ class Critic(nn.Module):
     def __call__(
         self,
         critic_hidden_state: Tuple[chex.Array, chex.Array],
-        observation_done: Tuple[chex.Array, chex.Array],
+        observation_done: RnnObservation,
     ) -> Tuple[chex.Array, chex.Array]:
         """Forward pass."""
         observation, done = observation_done
@@ -686,14 +688,14 @@ def run_experiment(_run: run.Run, _config: Dict, _log: SacredLogger) -> None:
     # Create envs
     generator = RandomGenerator(**config["rware_scenario"]["task_config"])
     env = jumanji.make(config["env_name"], generator=generator)
-    env = RwareMultiAgentWithGlobalStateWrapper(env)
+    env = GlobalStateWrapper(RwareWrapper(env))
     # Add agent id to observation.
     if config["add_agent_id"]:
         env = AgentIDWrapper(env=env, has_global_state=True)
     env = AutoResetWrapper(env)
     env = LogWrapper(env)
     eval_env = jumanji.make(config["env_name"], generator=generator)
-    eval_env = RwareMultiAgentWithGlobalStateWrapper(eval_env)
+    eval_env = GlobalStateWrapper(RwareWrapper(env))
     if config["add_agent_id"]:
         eval_env = AgentIDWrapper(env=eval_env, has_global_state=True)
 
--- a/mava/systems/rec_mappo_rware.py
+++ b/mava/systems/rec_mappo_rware.py
@@ -695,7 +695,7 @@ def run_experiment(_run: run.Run, _config: Dict, _log: SacredLogger) -> None:
     env = AutoResetWrapper(env)
     env = LogWrapper(env)
     eval_env = jumanji.make(config["env_name"], generator=generator)
-    eval_env = GlobalStateWrapper(RwareWrapper(env))
+    eval_env = GlobalStateWrapper(RwareWrapper(eval_env))
     if config["add_agent_id"]:
         eval_env = AgentIDWrapper(env=eval_env, has_global_state=True)
 
--- a/mava/types.py
+++ b/mava/types.py
@@ -21,16 +21,21 @@ from jumanji.types import TimeStep
 from optax._src.base import OptState
 from typing_extensions import NamedTuple, TypeAlias
 
-from mava.wrappers.jumanji import LogEnvState
-
 Action: TypeAlias = chex.Array
 Value: TypeAlias = chex.Array
 Done: TypeAlias = chex.Array
 HiddenState: TypeAlias = chex.Array
 
-# Can't know the exact type of State or Timestep.
+# Can't know the exact type of State.
 State: TypeAlias = Any
-Observation: TypeAlias = Any
+
+
+class Observation(NamedTuple):
+    agents_view: chex.Array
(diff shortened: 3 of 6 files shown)


mava/systems/rec_ippo_rware.py
```diff
@@ -48,10 +48,11 @@
     RecActorApply,
     RecCriticApply,
     RNNLearnerState,
+    RnnObservation,
```
OmaymaMahjoub, line 51: just for consistency can you name it RNNObservation or RecObservation 😅

mava/types.py
```diff
@@ -21,17 +21,58 @@
...
-Observation: TypeAlias = Any
+
+
+class Observation(NamedTuple):
+    agents_view: chex.Array
+    action_mask: chex.Array
+    step_count: chex.Numeric
+
+
+class ObservationGlobalState(NamedTuple):
```
sash-a, line 44: I really like that you moved this into here :+1:

mava/types.py
```diff
@@ -21,17 +21,58 @@
...
+    env_state: State
+    episode_returns: chex.Numeric
+    episode_lengths: chex.Numeric
+    # Information about the episode return and length for logging purposes.
+    episode_return_info: chex.Numeric
+    episode_length_info: chex.Numeric
+
+
 RnnObservation: TypeAlias = Tuple[Observation, Done]
+RnnGlobalObservation: TypeAlias = Tuple[ObservationGlobalState, Done]
```
sash-a, line 75: :fire:

mava/wrappers/jumanji.py
```diff
@@ -169,28 +136,34 @@ def observation_spec(self) -> specs.Spec[Observation]:
...
+    def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
+        n_agents = self._env.num_agents
+        observation = Observation(
             agents_view=timestep.observation.agents_view,
             action_mask=timestep.observation.action_mask,
-            step_count=jnp.repeat(timestep.observation.step_count, self._env.num_agents),
+            step_count=jnp.repeat(timestep.observation.step_count, n_agents),
         )
-        return state, timestep
+        shared_reward = jnp.sum(timestep.reward)
```
sash-a, line 153: Do we need this? I think separate rewards for each agent is a good thing allows them to learn independently

mava/types.py
```diff
@@ -21,17 +21,58 @@
...
 Done: TypeAlias = chex.Array
 HiddenState: TypeAlias = chex.Array
-
-# Can't know the exact type of State or Timestep.
+# Can't know the exact type of State.
 State: TypeAlias = Any
-Observation: TypeAlias = Any
+
+
+class Observation(NamedTuple):
```
sash-a, line 38: Could we take the docs from `ObservationGlobalState` and put them here and then only put the `global_state` doc in the observation global state.
```suggestion
class Observation(NamedTuple):
"""The observation that the agents see. Used for all decentralized systems.

    agents_view: the array that is passed to the networks during acting and learning.
    action_mask: boolean array specifying, for each agent, which action is legal.
    step_count: the number of steps elapsed since the beginning of the episode.
```

mava/types.py
```diff
@@ -21,17 +21,58 @@
...
+    """The observation that the agent sees.
+    agents_view: the agents' view of other agents and shelves within their
+        sensor range. The number of features in the observation array
+        depends on the sensor range of the agent.
+    action_mask: boolean array specifying, for each agent, which action
+        (up, right, down, left) is legal.
+    global_state: the global state of the environment, which is the
+        concatenation of the agents' views.
+    step_count: the number of steps elapsed since the beginning of the episode.
+    """
```
sash-a, line 54: Note you may have to run pre-commit after adding this as I'm not sure about line lenghts

```suggestion
    """The observation that the agents see. Used in centralized systems.
    
    Same as `Observation` except it adds a `global_state` attribute for centralized training.
  
    global_state: the global state of the environment. (Often is concatenation of the agents' views.)
    """
```

sash-a (commented):
I can't approve :cry: but I am happy with this, just one question about the shared reward and a minor docstring thing

sash-a:
Just for bookkeeping #951 was discovered in this PR, but we're deciding to fix it in a separate PR so we can get this one in ASAP.

mava/configs/system/ff_ippo.yaml
```diff
@@ -1,6 +1,6 @@
 # --- Defaults FF-IPPO ---
 
-num_updates: 1000 # Number of updates
+num_updates: 10000 # Number of updates
```
OmaymaMahjoub, line 3: can you reset it to 1000

mava/wrappers/jumanji.py
OmaymaMahjoub: I think we need to add it as a feature in the systems files instead of the wrapper, what do you think?

WiemKhlifi: Oh it's a mistake I added that to test the LBF environment I'll remove it and add this in the LBF wrapper :+1:.
For the idea of using separate rewards or the shared in LBF, we can use a flag as Omayma mentioned that will be helpful in future to test networks with non-shared parameters :crossed_fingers:

mava/configs/system/ff_ippo.yaml
WiemKhlifi: Sure!

OmaymaMahjoub (changes requested):
Can we add a feature called `use_team_reward` as we need it for a fair comparison between our work and existing works that rely on giving the agents the same reward measured by `mean(timestep.reward)`

WiemKhlifi:
I made all the necessary changes in this PR #952, just to make this PR smaller.
> Can we add a feature called `use_team_reward` as we need it for a fair comparison between our work and existing works that rely on giving the agents the same reward measured by `mean(timestep.reward)`

mava/evaluator.py
```diff
@@ -105,7 +105,9 @@ def evaluator_fn(trained_params: FrozenDict, rng: chex.PRNGKey) -> ExperimentOut
         # Add dimension to pmap over.
         step_rngs = jnp.stack(step_rngs).reshape(eval_batch, -1)
 
-        eval_state = EvalState(step_rngs, env_states, timesteps, 0, 0.0)
+        eval_state = EvalState(
+            step_rngs, env_states, timesteps, 0, jnp.zeros_like(timesteps.reward)
```
RuanJohn, line 109: Is this so that we can support team and individual rewards?

mava/systems/rec_ippo_rware.py
```diff
@@ -89,7 +90,7 @@ class Actor(nn.Module):
     def __call__(
         self,
         policy_hidden_state: chex.Array,
-        observation_done: Tuple[chex.Array, chex.Array],
+        observation_done: RnnObservation,
```
RuanJohn, line 93: SImilarly here. We should please go for `RNN` then instead of `Rnn`

mava/systems/rec_mappo_rware.py
```diff
@@ -42,18 +42,20 @@
     ExperimentOutput,
     HiddenStates,
     LearnerFn,
+    ObservationGlobalState,
     OptStates,
     Params,
     PPOTransition,
     RecActorApply,
     RecCriticApply,
+    RnnGlobalObservation,
```
RuanJohn, line 51: Same as above

mava/types.py
```diff
@@ -21,17 +21,60 @@
...
-
-# Can't know the exact type of State or Timestep.
+# Can't know the exact type of State.
 State: TypeAlias = Any
-Observation: TypeAlias = Any
+
+
+class Observation(NamedTuple):
+    """The observation that the agent sees.
+    agents_view: the agents' view of other agents and shelves within their
```
RuanJohn, line 40: Since we want this to be general the docstring should reflect this please. This docstring looks like it is rware specific.

mava/types.py
```diff
@@ -21,17 +21,60 @@
...
+        sensor range. The number of features in the observation array
+        depends on the sensor range of the agent.
+    action_mask: boolean array specifying, for each agent, which action is legal.
+    step_count: the number of steps elapsed since the beginning of the episode.
+    """
+
+    agents_view: chex.Array
```
RuanJohn, line 47: Also for consistency, can we please have shape comments here as well like below.

mava/types.py
```diff
@@ -21,17 +21,60 @@
...
+    action_mask: chex.Array
+    step_count: chex.Numeric
+
+
+class ObservationGlobalState(NamedTuple):
+    """The observation seen by agents in centralized systems.
```
RuanJohn, line 53:
```suggestion
    """The observation seen by agents in centralised systems.
```

mava/types.py
```diff
@@ -21,17 +21,60 @@
...
+    Extends `Observation` by adding a `global_state` attribute for centralized training.
```
RuanJohn, line 54:
```suggestion
    Extends `Observation` by adding a `global_state` attribute for centralised training.
```

mava/wrappers/jumanji.py
```diff
@@ -93,6 +60,46 @@ def step(
         return state, timestep
 
 
+class RwareWrapper(Wrapper):
```
RuanJohn, line 63: I really like this!!

mava/wrappers/jumanji.py
```diff
@@ -169,102 +178,52 @@ def observation_spec(self) -> specs.Spec[Observation]:
...
-            [0] * self._env.num_agents,
-            [self._env.time_limit] * self._env.num_agents,
-            "step_count",
-        )
-        return self._env.observation_spec().replace(step_count=step_count)
-
-
-class RwareMultiAgentWithGlobalStateWrapper(Wrapper):
-    """Multi-agent wrapper for the Robotic Warehouse environment.
+class GlobalStateWrapper(Wrapper):
```
RuanJohn, line 181: I wonder if we shouldn't have a `wrapper/mava.py` file that includes the more general wrappers like the agent IDs and the global state wrappers? Since I foresee us reusing these for any env we wrap.

RuanJohn (changes requested):
Thank you for all the work! Super nice changes! Just some minor comments and suggestions here and there.

mava/systems/rec_ippo_rware.py
WiemKhlifi: I think RecObservation may be better if you agree on this I'll change it 👌

mava/evaluator.py
WiemKhlifi: Actually, this was changed because we removed the reshaping of the reward outside of the system here [https://github.com/instadeepai/Mava/blob/7e500aaa2c657ce818689c589020f39284391da4/mava/systems/ff_ippo_rware.py#L152] and added it the wrapper in the modify_timesteps [https://github.com/instadeepai/Mava/blob/7e500aaa2c657ce818689c589020f39284391da4/mava/wrappers/jumanji.py#L77C6-L77C6] so that in future we can use the same system.

mava/types.py
WiemKhlifi: I missed it thank you Ruan I changed it! 🙏

mava/types.py
WiemKhlifi: Sure!

mava/wrappers/jumanji.py
WiemKhlifi: I really like this idea 🔥 ! We can name the file mava.py (like jumanji.py, gym.py ...) and then we change and adjust the environments wrapper separately 🙌

mava/types.py
```diff
@@ -21,17 +21,60 @@
...
-# Can't know the exact type of State or Timestep.
+# Can't know the exact type of State.
 State: TypeAlias = Any
-Observation: TypeAlias = Any
-RnnObservation: TypeAlias = Tuple[Observation, Done]
+
+
+class Observation(NamedTuple):
+    """The observation that the agent sees.
+    agents_view: the agents' view of other agents and items within their
```
RuanJohn, line 40: Maybe a simpler docstring here is to just say `the agent's view of the environment.`

mava/wrappers/jumanji.py
RuanJohn: @sash-a, keen to hear your thoughts here? I think we should everything jumanji specific in this jumanji file and move general wrappers to another file. Calling the file `mava.py` is a bit misleading since Mava isn't an environment. My suggestion is to call it `shared.py` since it will be wrappers shared between envs.

mava/types.py
WiemKhlifi: Thanks for suggesting this I'll change it 🙌

RuanJohn (approved):
Thank you @WiemKhlifi and @sash-a :fire:

mava/wrappers/jumanji.py
RuanJohn: I think that is okay, since the base Mava API is an extension of the Jumanji API. Thank you for the changes :raised_hands:

OmaymaMahjoub (approved):
Thanks @WiemKhlifi  and @sash-a  :fire:

PR #952, opened 2023-11-24
feat(lbf): generalise wrapper and logger

WiemKhlifi:
## What?
Based on the changes made in this **first PR** #948 for the **RwareWrapper**, we added a separate **LbfWrapper** as well as some changes to support the LBF environment. -> Merged them into one single wrapper 🤞 
## Why?
LbfWrapper is useful in this case to choose what reward will be used during training whether it is the shared reward or not.
## How?
* Create `LbfWrapper` and separate file systems for LBF -> changed this one general wrapper
* Edit the `logger_utils` file to log LBF tasks.
* Create a folder for different config files for LBF.


mava/configs/env/lbf.yaml
```diff
@@ -0,0 +1,11 @@
...
+defaults:
+  - _self_
+  - lbf_scenario: Foraging-2s-8x8-2p-2f-coop # [Foraging-2s-8x8-2p-2f-coop, Foraging-8x8-2p-2f-coop, Foraging-2s-10x10-3p-3f, Foraging-10x10-3p-3f, Foraging-15x15-4p-3f, Foraging-15x15-4p-5f]
+
+env_name: LevelBasedForaging-v0
+task_name: Foraging-2s-8x8-2p-2f-coop
+# Choose whether to use the use_idividual_reward = True OR aggregate_rewards=True.
+use_idividual_reward: False # If True, the network uses the list of different rewards given.
+aggregate_rewards : True  # If True, aggregates rewards across agents using the specified function.
+aggregate_rewards_with: sum # The function for aggregating rewards ("sum" or "mean").
```
RuanJohn, line 11: NIT: I think we could name this `reward_aggregation_function` ?

mava/configs/env/lbf_scenario/Foraging-15x15-4p-3f.yaml
```diff
@@ -0,0 +1,9 @@
+# The config of the Foraging-15x15-4p-3f scenario
+
+task_config_gen:
+  grid_size: 15
+  fov : 15
```
RuanJohn, line 5: A general description somewhere for what these attributes are would be really nice

mava/configs/env/lbf_scenario/Foraging-10x10-3p-3f.yaml
```diff
@@ -0,0 +1,9 @@
+# The config of the Foraging-10x10-3p-3f scenario
+
+task_config:
```
RuanJohn, line 3: A general comment on the config naming. I think it will be nicer to to remove `Foraging-` from all the LBF scenario names since all the scenario configs are already in the `lbf_scenario` folder.

mava/systems/ff_ippo_lbf.py
```diff
@@ -0,0 +1,638 @@
...
+        elapsed_time = time.time() - start_time
+        evaluator_output.episodes_info["steps_per_second"] = steps_per_rollout / elapsed_time
+        log(
+            metrics=evaluator_output,
+            t_env=steps_per_rollout * (i + 1),
+            absolute_metric=True,
+        )
+
+
+@hydra.main(config_path="../configs", config_name="default_ff_ippo.yaml", version_base="1.2")
```
RuanJohn, line 625: It will also be good if we have a default_ff_ippo_lbf config or something along those lines? Because this will probably not do what we want if the default ppo config points to rware.

mava/wrappers/jumanji.py
```diff
@@ -93,6 +59,84 @@ def step(
         return state, timestep
 
 
+class MultiAgentWrapper(Wrapper):
```
RuanJohn, line 62: I really like this wrapper!!

mava/wrappers/jumanji.py
```diff
@@ -93,6 +59,84 @@ def step(
...
+        self._num_agents = self._env.num_agents
+        # Mapping aggregation functions
+        self.aggregation_functions = {
+            "sum": jnp.sum,
+            "mean": jnp.mean,
+            # Add more functions as needed
+        }
+        self.aggregate_function = self.aggregation_functions.get(
+            aggregate_rewards_with, jnp.sum
+        )  # No error!
```
RuanJohn, line 93: Not sure what this comment means?

mava/wrappers/jumanji.py
```diff
@@ -93,6 +59,84 @@ def step(
...
+
+    def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
+        """Modify the timestep based on the specified reward handling strategy."""
+        observation = Observation(
+            agents_view=timestep.observation.agents_view,
+            action_mask=timestep.observation.action_mask,
+            step_count=jnp.repeat(timestep.observation.step_count, self._num_agents),
+        )
+
+        if self._use_individual_reward:
```
RuanJohn, line 110: Maybe just a comment at the top explaining that this will replace the timesteps accordingly? It is a bit hard to follow with all the returns.

RuanJohn (changes requested):
Super nice work! :fire: I really like the general env wrapper! Just some changes here and there.

mava/configs/env/lbf.yaml
WiemKhlifi: Agree!

mava/systems/ff_ippo_lbf.py
WiemKhlifi: I thought that we simply make changes in the default files but if this makes it easier to run experiments later I'll be happy to quickly add them!

mava/configs/env/lbf_scenario/Foraging-10x10-3p-3f.yaml
WiemKhlifi: Sure I'll change them it makes the name longer and it's redundant!

mava/configs/env/lbf_scenario/Foraging-15x15-4p-3f.yaml
WiemKhlifi: Sure! Also, should add the config and description for the case if we want to use the GridObserver for LBF? (in this case, the observation is a vector and VectorObserver is set as default).

mava/wrappers/jumanji.py
WiemKhlifi: Sorry it's a typo.

mava/wrappers/jumanji.py
WiemKhlifi: I added some details as comments and docs, should we add more or change something?

mava/wrappers/jumanji.py
sash-a: Unfortunately this is not going to continue to work. We *need* a wrapper per jumanji env so that we can correctly pack their obs into our `Observation`. 

For example connectors observation looks like this:
```python
class Observation
    grid
    action_mask
    step_count
```

Then many other jumanji envs have different values here. Only lbf and rware have the same obs names that is why this works now, but it won't continue to work, so we need to separate this per env.

mava/wrappers/jumanji.py
```diff
@@ -12,35 +12,87 @@
...
+        aggregate_rewards (bool): If True, aggregates rewards across agents.
+        reward_aggregation_function (str): The function for aggregating rewards ("sum" or "mean").
+    """
+
+    def __init__(
+        self,
+        env: Environment,
+        use_individual_reward: bool = False,
+        aggregate_rewards: bool = False,
+        reward_aggregation_function: str = "sum",
```
sash-a, line 43: I think we shouldn't have these options, we should decide if we want individual/aggregated rewards and use the option we want in each separate wrapper

mava/systems/ff_ippo_lbf.py
sash-a, line 1: Separate files for each env isn't ideal as then our number of files will explode. We want lbf/rware/future envs to share the same files. So we need to make a function e.g `make` that parses our env config and makes the correct env and then we can use that function in all the files.

mava/wrappers/jumanji.py
RuanJohn: I back the idea of a wrapper per env in that case.

mava/wrappers/jumanji.py
RuanJohn: If we are doing a different wrapper per env then hard setting the aggregation is fine.

mava/wrappers/jumanji.py
```diff
@@ -62,3 +55,87 @@ def observation_spec(self) -> specs.Spec[Observation]:
...
+    ):
+        super().__init__(env)
+        self._env: LevelBasedForaging
+        self._aggregate_rewards = aggregate_rewards
+        self.reward_aggregation_function = reward_aggregation_function
+        self._num_agents = self._env.num_agents
+        # Mapping aggregation functions
+        self.aggregation_functions: Dict[str, Callable] = {
+            "sum": jnp.sum,
```
RuanJohn, line 102: I agree with @sash-a's previous comment on this. I think we should just do what the benchmarking paper does and keep that as our default aggregation function. The only boolean we would then need is whether or not to aggregate the rewards over agents or not.

mava/configs/system/ff_ippo.yaml
```diff
@@ -1,6 +1,6 @@
 # --- Defaults FF-IPPO ---
 
-num_updates: 1000 # Number of updates
+num_updates: 5000 # Number of updates
```
RuanJohn, line 3: Please leave this to 1000 :)

mava/configs/env/rware.yaml
```diff
@@ -4,3 +4,4 @@ defaults:
   - rware_scenario: tiny-2ag # [tiny-2ag, tiny-4ag, tiny-4ag-easy, small-4ag]
 
 env_name: RobotWarehouse-v0
+task_name: tiny-2ag
```
RuanJohn, line 7: I think it is better to leave this in the child configs, otherwise we always have to set the scenario name and the task_name which could lead to accidental mismatches.

mava/wrappers/jumanji.py
WiemKhlifi: For the aggregation  I let the user choose whether to use the list of individual rewards (default setting) or  set the `aggregate_rewards=True` in lbf config file.

mava/configs/env/lbf.yaml
```diff
@@ -0,0 +1,11 @@
+# ---Environment Configs---
+defaults:
+  - _self_
+  - lbf_scenario: 2s-8x8-2p-2f-coop # [2s-8x8-2p-2f-coop, 8x8-2p-2f-coop, 2s-10x10-3p-3f, 10x10-3p-3f, 15x15-4p-3f, 15x15-4p-5f]
+# Further enviroment config details in "2s-8x8-2p-2f-coop" file.
+
+env_name: LevelBasedForaging-v0
+task_name: 2s-8x8-2p-2f-coop
```
RuanJohn, line 8: Similarly here, I think it is better to set the task names in the child configs to avoid logging mismatches.

mava/wrappers/jumanji.py
RuanJohn: This is good, but I don't think we should give an option for _what_ aggregation to use. We should just hardcode it in the wrapper to be the same as what other benchmarks do for each environment.

mava/wrappers/jumanji.py
WiemKhlifi: I'll change the aggregation to true the default  to use the sum of the individual rewards as the benchmarking paper did and make the other case optional (remove the mean function as well).

mava/wrappers/jumanji.py
```diff
@@ -62,3 +55,87 @@ def observation_spec(self) -> specs.Spec[Observation]:
...
+            "mean": jnp.mean,
+            # Add more functions as needed
+        }
+        self.aggregate_function: Callable = self.aggregation_functions.get(
+            reward_aggregation_function, jnp.sum
+        )
```
sash-a, line 108: Bit cleaner

```suggestion

        reward_aggregation_fn: Optional[Callable[[Array], Array] = jnp.sum,
    ):
        super().__init__(env)
        self._env: LevelBasedForaging
        self._num_agents = self._env.num_agents
        self._aggregate_rewards = reward_aggregation_fn is not None
        self.aggregate_function = reward_aggregation_fn
```

sash-a (commented):
Slightly cleaner way to do the reward aggregation :+1:

mava/wrappers/jumanji.py
WiemKhlifi: I set the `jnp.sum` as the only function we'll use as @RuanJohn mentioned, I hope the current changes are better now 🙏

sash-a: even better!

mava/configs/env/scenario/2s-8x8-2p-2f-coop.yaml
```diff
@@ -0,0 +1,15 @@
+# The config of the 2s-8x8-2p-2f-coop scenario with the VectorObserver set as default.
+task_name: 2s-8x8-2p-2f-coop
+
+task_config:
+  grid_size: 8 # size of the grid to generate.
+  fov: 2       # field of view of an agent.
+  num_agents: 2 # number of agents on the grid.
+  num_food: 2   # number of food in the environment.
+  max_agent_level: 2 # maximum level of the agents (inclusive).
+  force_coop: True   # force cooperation between agents.
```
RuanJohn, line 10: Bit cleaner
```suggestion
  grid_size: 8 # size of the grid to generate.
  fov: 2 # field of view of an agent.
  num_agents: 2 # number of agents on the grid.
  num_food: 2 # number of food in the environment.
  max_agent_level: 2 # maximum level of the agents (inclusive).
  force_coop: True # force cooperation between agents.
```

RuanJohn (commented):
One final detail here is that we need to update the requirements to point to a new branch we want to use as the base Jumanji branch so that we have LBF included. 
Do you have a suggestion for where this branch should live @sash-a?

mava/wrappers/jumanji.py
```diff
@@ -17,48 +17,107 @@
...
-        )
-        reward = jnp.repeat(timestep.reward, n_agents)
-        discount = jnp.repeat(timestep.discount, n_agents)
-        return timestep.replace(observation=observation, reward=reward, discount=discount)
+        """Modify the timestep for the selected enviroment."""
+        pass
 
     def reset(self, key: chex.PRNGKey) -> Tuple[State, TimeStep]:
-        """Reset the environment. Updates the step count."""
+        """Reset the environment and updates the timestep."""
```
sash-a, line 40:
```suggestion
        """Reset the environment."""
```

mava/wrappers/jumanji.py
```diff
@@ -17,48 +17,107 @@
...
         state, timestep = self._env.reset(key)
         return state, self.modify_timestep(timestep)
 
     def step(self, state: State, action: chex.Array) -> Tuple[State, TimeStep]:
-        """Step the environment. Updates the step count."""
+        """Step the environment and updates the timestep."""
```
sash-a, line 45:
```suggestion
        """Step the environment."""
```

mava/wrappers/jumanji.py
```diff
@@ -17,48 +17,107 @@
...
-            [0] * self._env.num_agents,
-            [self._env.time_limit] * self._env.num_agents,
+            [0] * self._num_agents,
+            [self.time_limit] * self._num_agents,
             "step_count",
         )
         return self._env.observation_spec().replace(step_count=step_count)
+
+
+class RWAREWrapper(MultiAgentWrapper):
```
sash-a, line 61: I just don't like how it looks with all caps, but I don't mind this too much
```suggestion
class RwareWrapper(MultiAgentWrapper):
```

mava/wrappers/jumanji.py
```diff
@@ -17,48 +17,107 @@
...
+            agents_view=timestep.observation.agents_view,
+            action_mask=timestep.observation.action_mask,
+            step_count=jnp.repeat(timestep.observation.step_count, self._num_agents),
+        )
+        reward = jnp.repeat(timestep.reward, self._num_agents)
+        discount = jnp.repeat(timestep.discount, self._num_agents)
+        return timestep.replace(observation=observation, reward=reward, discount=discount)
+
+
+class LBFWrapper(MultiAgentWrapper):
```
sash-a, line 79: Same here

```suggestion
class LbfWrapper(MultiAgentWrapper):
```

mava/wrappers/jumanji.py
```diff
@@ -17,48 +17,107 @@
...
+    """
+     Multi-agent wrapper for the Level-Based Foraging environment.
+
+    Args:
+        env (Environment): The base environment.
+        aggregate_rewards (bool): If True, aggregates rewards across agents.
+        reward_aggregation_function (str): The function for aggregating rewards ("sum" or "mean").
```
sash-a, line 86:
```suggestion
        use_individual_rewards (bool): If true each agent gets a separate reward, sum reward otherwise.
```

mava/wrappers/jumanji.py
```diff
@@ -17,48 +17,107 @@
...
+    """
+
+    def __init__(
+        self,
+        env: LevelBasedForaging,
+        use_individual_rewards: bool = False,
+    ):
```
sash-a, line 93: Nit: can this fit on a single line?

mava/wrappers/jumanji.py
```diff
@@ -17,48 +17,107 @@
...
 from jumanji.wrappers import Wrapper
 
 from mava.types import Observation, State
 
 
-class RwareWrapper(Wrapper):
-    """Multi-agent wrapper for the Robotic Warehouse environment."""
-
-    def __init__(self, env: RobotWarehouse):
+class MultiAgentWrapper(Wrapper):
```
sash-a, line 29: This is a great base class for our future environments :fire:

sash-a (changes requested):
Very minor nitpicks, otherwise I'm happy :rocket:

mava/wrappers/jumanji.py
RuanJohn: I also back not having all caps

mava/configs/env/scenario/15x15-4p-3f.yaml
```diff
@@ -0,0 +1,15 @@
...
+task_config:
+  grid_size: 15
+  fov: 15
+  num_agents: 4
+  num_food: 3
+  max_agent_level: 2
+  force_coop: False
+
+# GridObserver config
+grid_obs_config:
```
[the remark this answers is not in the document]
RuanJohn: They use the vector one in the paper yes. Since we don't have any conv nets yet in mava it might even be a good idea to remove the grid obs stuff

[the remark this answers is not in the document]
WiemKhlifi: Yes the default observer is the vector one so these configs will be unnecessary since we're only considering the vector observation, I'll remove them.

PR #954, opened 2023-11-27
feat: Model Checkpointing

callumtilbury:
## What?
Ability to checkpoint the `learner_state` using `orbax` as a base.

## Why?
Important for our research in multiple dimensions.

## How?
Built in the same way as for internal CityLearn code, wrapping `orbax`'s functionality.

## Extra
This PR adds simple checkpointing to each system, but we can remove that by default, if preferred. It does _not_ include any checkpoint reloading, but that can be done in the following way (example for Rec MAPPO with RWARE):
```py
loaded_state = Checkpointer(
        model_name="rec_mappo_rware",
        config=config,
        timestamp_override="20231127155319",
).restore_learner_state()

loaded_state = jax.device_put_replicated(loaded_state, jax.devices())

learner_state = learner_state._replace(
        params=Params._make(FrozenDict(loaded_state['params']).values()),
        hstates=HiddenStates._make(FrozenDict(loaded_state['hstates']).values())
)
```
But I do want to think of cleaner ways to reload—perhaps for a later PR?


OmaymaMahjoub:
> Open Q: should I add checkpointing options to config, or only later, when we find a need?

Thanks for the PR @callumtilbury, in my opinion, I think it will be better to make it optional and also make the `model_name` a config param

OmaymaMahjoub (changes requested):
If you can add to the .gitignore`checkpoints/`

mava/configs/logger/base_logger.yaml
```diff
@@ -9,3 +9,15 @@ use_neptune: True
...
+checkpointing:
+  save_model: True # Whether to save model checkpoints.
+  save_args:
+    save_interval_steps: 1 # Number of steps between saving checkpoints.
+    max_to_keep: 1 # Maximum number of checkpoints to keep.
+    keep_period: ~ # Don't delete any checkpoint where step % keep_period == 0
+
+  load_model: False # Whether to load model checkpoints.
+  load_args:
+    timestamp_override: "" # Timestamp override for loading checkpoint.
```
RuanJohn, line 23: This is not immediately apparent to me. Is possible to be a bit more descriptive with comment please?

mava/utils/checkpointing.py
```diff
@@ -0,0 +1,200 @@
...
+                If set, will not delete any checkpoint where
+                checkpoint_step % keep_period == 0. Defaults to None.
+        """
+
+        # Sharding info will be read from file, rather than from 'RestoreArgs'
+        warnings.filterwarnings(
+            action="ignore",
+            category=UserWarning,
+            message="Couldn't find sharding info under RestoreArgs",
+        )
```
RuanJohn, line 74: Just a question, what is this and why do we suppress this warning?

mava/utils/checkpointing.py
```diff
@@ -0,0 +1,200 @@
...
+
+        orbax_checkpointer = orbax.checkpoint.PyTreeCheckpointer()
+        timestamp_str = (
```
RuanJohn, line 77: Maybe can change this name to something like `checkpoint_uid` to make it a bit more descriptive?

mava/configs/logger/base_logger.yaml
RuanJohn: See the comment below on renaming this.

mava/utils/checkpointing.py
```diff
@@ -0,0 +1,200 @@
...
+import absl.logging as absl_logging
+import orbax.checkpoint
+from chex import Numeric
+from flax.core.frozen_dict import FrozenDict
+from jax.tree_util import tree_map
+from omegaconf import DictConfig
+
+from mava.types import HiddenStates, LearnerState, Params, RNNLearnerState
+
+CHECKPOINTER_VERSION = 0.1
```
RuanJohn, line 33: Very good to have versioning here! Can you please just add a comment on why.

mava/utils/checkpointing.py
```diff
@@ -0,0 +1,200 @@
...
+                Defaults to 0.0.
+
+        Returns:
+            bool: whether the saving was successful.
+        """
+        model_save_success: bool = self._manager.save(
+            step=timestep,
+            items={
+                "learner_state": unreplicated_learner_state,
+                "type": re.findall(
```
RuanJohn, line 135: Could you please add a comment about what this regex is doing?

mava/utils/checkpointing.py
```diff
@@ -0,0 +1,200 @@
...
+        self,
+        unreplicated_input_learner_state: Union[LearnerState, RNNLearnerState],
+        n: Optional[int] = None,
+        restore_params: bool = True,
+        restore_hstates: bool = True,
+    ) -> Union[LearnerState, RNNLearnerState]:
+        """Restore the learner state.
+
+        Args:
+            n (Optional[int], optional):
```
RuanJohn, line 156: I'm finding it difficult to find a nice name for this argument. But many something like `restart_env_timestep`?

RuanJohn (changes requested):
Really rally nice PR @callumtilbury and very easy to follow. Just a few questions and some suggestions around variable names.

mava/utils/checkpointing.py
callumtilbury: Done!

mava/utils/checkpointing.py
callumtilbury: Originally I was getting an unnecessary warning thrown about how I was saving a sharded array, but it's no longer there—so I can actually remove it! 🖖🏻

mava/utils/checkpointing.py
callumtilbury: Changed to `timestep` as it refers to the same thing as `timestep` in the `save()` method.

mava/utils/checkpointing.py
callumtilbury: Much better! Thanks.

mava/utils/checkpointing.py
callumtilbury: If we hit it again in the future, we can revisit :)

mava/configs/logger/base_logger.yaml
callumtilbury: Done :)

mava/utils/checkpointing.py
callumtilbury: Done!

mava/utils/checkpointing.py
RuanJohn: Sounds good, thanks man :)

callumtilbury: Oh wait, sorry! The warning is thrown when loading checkpoints. The full error message is
```
UserWarning: Couldn't find sharding info under RestoreArgs. Populating sharding info from sharding file. Please note restoration time will be slightly increased due to reading from file instead of directly from RestoreArgs.
```
So it uses the checkpointed file for sharding info, which is what we want (I believe), rather than providing it in the code. I think we can happily suppress the warning?

RuanJohn: Sounds good. Thanks :)

mava/utils/checkpointing.py
RuanJohn: Perfect

RuanJohn (approved):
Thanks @callumtilbury! Super nice! :raised_hands:

OmaymaMahjoub (approved):
Thanks  for the PR @callumtilbury :fire:

PR #955, opened 2023-11-29
feat: jaxmarl wrapper

sash-a:
## What?
Ability to support jaxmarl envs
## Why?
To experiment with more environments
## How?
With a wrapper

## Extra
Depends on choices made in #948 and #952 for how we're going to support new envs


mava/wrappers/jaxmarl.py
```diff
@@ -0,0 +1,236 @@
...
+        raise ValueError(f"Unsupported JaxMarl space: {space}")
+
+
+class JaxMarlState(NamedTuple):
+    """Wrapper around a JaxMarl state to provide necessary attributes for jumanji environments."""
+
+    state: ArrayTree
+    key: PRNGKey
+    step: int
+
```
OmaymaMahjoub, line 148: Maybe add this to types.py

mava/wrappers/jaxmarl.py
```diff
@@ -0,0 +1,236 @@
...
+    Note: currently this is done by checking the shape of the observation and action spaces
+    as gymnax/jaxmarl environments do not have a custom __eq__ for their specs.
+    """
+    agents = list(env.observation_spaces.keys())
+
+    main_agent_obs_shape = env.observation_space(agents[0]).shape
+    main_agent_act_shape = env.action_space(agents[0]).shape
+    # Cannot easily check low, high and n are the same, without being very messy.
+    # Unfortunately gymnax/jaxmarl doesn't have a custom __eq__ for their specs.
```
OmaymaMahjoub, line 91: Can you give it a try in an additional function else maybe add it as an issue

mava/wrappers/jaxmarl.py
```diff
@@ -0,0 +1,236 @@
...
+        homogenous_error = (
+            f"Mava only supports environments with homogeneous agents, "
+            f"but you tried to use {env} which is not homogeneous."
+        )
+        assert is_homogenous(env), homogenous_error
+
+        super().__init__(env)
+        self._env: MultiAgentEnv
+        self._timelimit = timelimit
+        self._action_shape = self.action_spec().shape
```
OmaymaMahjoub, line 164:
```suggestion
        self._action_shape = (self.action_spec().shape[0],  int(self.action_spec().num_values[0]))
```

requirements/requirements.txt
```diff
@@ -1,9 +1,10 @@
 colorama
 distrax @ git+https://github.com/google-deepmind/distrax  # distrax release doesn't support jax > 0.4.13
-flax==0.7.5
+flax~=0.7.5
```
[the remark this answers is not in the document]
sash-a: So @RuanJohn was worried that it will create a breaking change like they did previously. This `~=` means that we will stay in `0.7.x` so there shouldn't be a breaking change within here

mava/wrappers/jaxmarl.py
sash-a: I did try this earlier and the function just got very big and very messy :cry:

OmaymaMahjoub: If you can add an issue on mava then for it and we will look into it another time :pray:

OmaymaMahjoub (approved):
Thanks @sash-a :fire:

PR #959, opened 2023-12-06
Feat/sebulba ippo

OmaymaMahjoub:
## What?
Implement Sebulba architecture with feedforward IPPO on Rware.
## Why?
Integrate Sebulba's architecture due to its effectiveness in scenarios involving non-jitted/non-jax environments.
## How?
Enhance the existing [Cleanba](https://github.com/vwxyzjn/cleanba) code to support marl algorithm and ensure compatibility with Mava's key components, including logger, evaluator, etc.
## Extra
Action Item: 
   - Update the project's README to reflect these changes.
   - Support absolute metric
   - Add model checkpointer


mava/types.py
```diff
@@ -87,6 +86,18 @@ class PPOTransition(NamedTuple):
     info: Dict
 
 
+class SebulbaTransition(NamedTuple):
```
sash-a, line 89: We should try keep this to the same transition. If we're going to make a new transition for each system then those types should go in their own files

mava/types.py
```diff
@@ -118,6 +129,13 @@ class LearnerState(NamedTuple):
     timestep: TimeStep
 
 
+class SebulbaLearnerState(NamedTuple):
```
sash-a, line 132: Same here, can we not use the existing `LearnerState` it will make the mava API much cleaner because then everything will return a `LearnerState`/`RnnLearnerState`

mava/utils/jax.py
```diff
@@ -42,3 +40,26 @@ def merge_leading_dims(x: chex.Array, num_dims: chex.Numeric) -> chex.Array:
 
     new_shape = (np.prod(x.shape[:num_dims]),) + x.shape[num_dims:]
     return x.reshape(new_shape)
+
+
+def convert_data(
```
sash-a, line 45: Need a more explicit name
```suggestion
def add_minibatch_dim(
```

mava/utils/sebulba_tools.py
sash-a, line 1: I think just call it `sebulba.py` as we're already in the `utils` folder

mava/utils/sebulba_tools.py
```diff
@@ -0,0 +1,38 @@
...
+    os.environ["TF_XLA_FLAGS"] = "--xla_gpu_autotune_level=2 --xla_gpu_deterministic_reductions"
+    os.environ["TF_CUDNN DETERMINISTIC"] = "1"
+
+    os.environ["JAX_USE_PJRT_C_API_ON_TPU"] = ""
+    # Fix weird OOM https://github.com/google/jax/discussions/6332#discussioncomment-1279991
+    os.environ["XLA_PYTHON_CLIENT_MEM_FRACTION"] = "0.6"
+    os.environ["XLA_FLAGS"] = "--xla_cpu_multi_thread_eigen=false " "intra_op_parallelism_threads=1"
+    # Fix CUDNN non-determinisim; https://github.com/google/jax/issues/4823#issuecomment-952835771
+    os.environ["TF_XLA_FLAGS"] = "--xla_gpu_autotune_level=2 --xla_gpu_deterministic_reductions"
+    os.environ["TF_CUDNN DETERMINISTIC"] = "1"
```
sash-a, line 38: Same thing twice
```suggestion
```

mava/utils/sebulba_tools.py
```diff
@@ -0,0 +1,38 @@
...
+import os
+
+
+def configure_computation_environment() -> None:
+    """Configure the computation environment for JAX.
+
+    Note: Must be called before any JAX computation is run.
+    """
+    os.environ["JAX_USE_PJRT_C_API_ON_TPU"] = ""
+
```
sash-a, line 24: JAX_USE_PJRT_C_API_ON_TPU no longer has an effect (i.e. it always defaults to true).
```suggestion
```

mava/utils/sebulba_tools.py
```diff
@@ -0,0 +1,38 @@
...
+    # Fix weird OOM https://github.com/google/jax/discussions/6332#discussioncomment-1279991
+    os.environ["XLA_PYTHON_CLIENT_MEM_FRACTION"] = "0.6"
```
sash-a, line 26: I don't think we want this, it's fine to use more than 60% of GPU/TPU memory

mava/utils/sebulba_tools.py
```diff
@@ -0,0 +1,38 @@
...
+    os.environ["XLA_FLAGS"] = "--xla_cpu_multi_thread_eigen=false " "intra_op_parallelism_threads=1"
+    # Fix CUDNN non-determinisim; https://github.com/google/jax/issues/4823#issuecomment-952835771
+    os.environ["TF_XLA_FLAGS"] = "--xla_gpu_autotune_level=2 --xla_gpu_deterministic_reductions"
+    os.environ["TF_CUDNN DETERMINISTIC"] = "1"
```
sash-a, line 30: This is making the code more deterministic at the cost of speed, I think we shouldn't add this, but happy to hear thoughts

mava/wrappers/robot_warehouse.py
```diff
@@ -0,0 +1,227 @@
...
+                        add_agent_id=config.system.add_agent_id,
+                    )
+                    for _ in range(num_envs)
+                ]
+            )
+        envs.num_envs = num_envs
+        envs.is_vector_env = True
+        return envs
+
+    return thunk
```
sash-a, line 227: We'd never want to use `SyncVectorEnv` unless debugging and only one of these needs an inner method.

```suggestion
def make_env(
    map_name: str = "rware-tiny-2ag-v1",
    team_reward: bool = True,
    add_agent_id: bool = True,
) -> Callable[[], gym.Env]:
    """Create a function which creates a fully configured environment."""

    def thunk() -> gym.Env:
        env = GymWrapper(
            env=gym.make(map_name),
            team_reward=team_reward,
        )
        if add_agent_id:
            env = AgentIDWrapper(env)

        return env

    return thunk


def make_envs(num_envs: int, config: DictConfig) -> gym.vector.VectorEnv:
    envs = gym.vector.AsyncVectorEnv(
        [
            make_env(
                map_name="rware-tiny-2ag-v1",
                team_reward=config.system.use_team_reward,
                add_agent_id=config.system.add_agent_id,
            )
            for _ in range(num_envs)
        ]
    )

    envs.num_envs = num_envs
    envs.is_vector_env = True
    return envs
```

mava/wrappers/robot_warehouse.py
```diff
@@ -0,0 +1,227 @@
...
+        obs = np.concatenate([self.agent_ids, obs], axis=1)
+        return obs, info
+
+    def step(self, action: list) -> Tuple[np.ndarray, float, bool, bool, Dict]:
+        """Step the environment."""
+        obs, reward, terminated, truncated, info = self.env.step(action)
+        obs = np.concatenate([self.agent_ids, obs], axis=1)
+        return obs, reward, terminated, truncated, info
+
+
```
sash-a, line 175: Couple other comments here, it would be nice if we could integrate this with @WiemKhlifi new make function.

Do we need to set 
```python
       envs.num_envs = num_envs
       envs.is_vector_env = True
```
Do we use them? I think `num_envs` is set automatically anyways

pyproject.toml
```diff
@@ -77,4 +77,5 @@ ignore = [
     "W503",  # Line break before binary operator (not compatible with black).
     "B017",  # assertRaises(Exception): or pytest.raises(Exception) should be considered evil.
     "E203",  # black and flake8 disagree on whitespace before ':'.
+    "E402",  # Module level import not at top of file.
```
sash-a, line 80: I think just do a comment ignore, we still want this warning in general

mava/logger.py
```diff
@@ -36,10 +36,22 @@ def __call__(
         ...
 
 
-def get_logger_tools(logger: Logger) -> LogFn:  # noqa: CCR001
+class SebulbaLogFn(Protocol):
```
sash-a, line 39: I really don't think this is a good idea, can we please use the existing log_fn otherwise there will be a lot of code duplication

sash-a (changes requested):
Thanks for this @OmaymaMahjoub! Haven't looked at the actual sebulba file, but quite a few fundamental changes, keen to hear your thoughts.

OmaymaMahjoub:
Many changes happened to `deveop` since the last update of this PR that will result in a massive merge conflict, therefore we need to create a new branch and new PR for the sebulba ippo!

mava/logger.py
OmaymaMahjoub: I tired to address this issue in #977

PR #962, opened 2023-12-08
Feat/add time to json logger

RuanJohn:
## What?
Add elapsed wallclock time to the json files. This allows for richer plots down stream. 
## Extra
For now, the time is offset by the time of the first evaluation. This means that we won't take compilation time into and will have the first evaluation step at `t=0`. Alternatively, we could change this which would mean that the first time on the x-axis for the first evaluation step would be some non-zero value `k` and the gap between `t=0` and `t=k` will be larger than the gap between all other evaluation intervals.

### DIFF
--- a/mava/utils/logger_tools.py
+++ b/mava/utils/logger_tools.py
@@ -15,6 +15,7 @@
 import json
 import logging
 import os
+import time
 from datetime import datetime
 from typing import Dict, Optional
 
@@ -172,6 +173,7 @@ class JsonWriter:
         environment_name: str,
         seed: int,
     ):
+        self.start_time = time.time()
         self.path = path
         self.file_name = "metrics.json"
         self.run_data: Dict = {"absolute_metrics": {}}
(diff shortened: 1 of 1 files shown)


mava/utils/logger_tools.py
```diff
@@ -172,6 +173,7 @@ def __init__(
         environment_name: str,
         seed: int,
     ):
+        self.start_time = time.time()
```
callumtilbury, line 176: Will this not always be overwritten? We init the logger, and `self.start_time = time.time()`, but then when `evaluation_step == 0`, we set it again.

Perhaps it's better here to write something like,
```suggestion
        self.start_time = -1
```

That way, it will always be overwritten later, & if there's a bug, we can notice it easily? :)

callumtilbury (changes requested):
Thanks @RuanJohn! Just a minor suggestion.

OmaymaMahjoub (approved):
LGTM :fire:

PR #968, opened 2023-12-15
Update main with the latest Develop

RuanJohn:
## What?
Our main branch is massively outdated. This is just a quick PR to get it synced with the latest version of develop. 
In the new year, we will delete main and just use develop or we will move to using exclusively main.


OmaymaMahjoub (approved):
:fire: :fire: :fire:

arnupretorius (approved):
Thanks @RuanJohn!

PR #970, opened 2023-12-18
Feat: support SMAX

OmaymaMahjoub:
## What?
Support SMAX environment from [jaxmarl](https://github.com/FLAIROx/JaxMARL/tree/main)
## Why?
Add more environments that Mava supports
## How?
- Get action masking if exists
- Get global state instead of concatenate obs for centralized training.
- Get win rate during evaluation and log it to Neptune, json file, and cout in terminal
## Extra
TODO: add `27m_vs_30m` results to the readme


mava/wrappers/jaxmarl.py
```diff
@@ -157,16 +157,23 @@ def __init__(self, env: MultiAgentEnv, timelimit: int = 500):
 
         self.agents = list(self._env.observation_spaces.keys())
 
+        # Action mask fn
+        if hasattr(self._env, "get_avail_actions"):
+            self.get_avail_actions = self._get_avail_actions
+        else:
+            self.get_avail_actions = lambda _: jnp.ones(self._action_shape)
```
sash-a: `hasattr` is not something we want to do, do the jaxmarl envs not have a consistent way to get the action mask?

mava/wrappers/jaxmarl.py
```diff
@@ -157,16 +157,23 @@ def __init__(self, env: MultiAgentEnv, timelimit: int = 500):
...
 
+        extras = {"global_state": jnp.array(obs["world_state"])}
         obs = Observation(
             agents_view=batchify(obs, self.agents),
-            action_mask=jnp.ones(self._action_shape),
+            action_mask=self.get_avail_actions(state),
             step_count=jnp.zeros(self._env.num_agents, dtype=int),
         )
-        return JaxMarlState(state, key, 0), restart(obs, extras={}, shape=(self.num_agents,))
+        return JaxMarlState(state, key, 0), restart(obs, extras=extras, shape=(self.num_agents,))
```
sash-a: Lets not put global state in extras, we have and `ObservationGlobalState` so lets always use that for the global state and if there is no global state then we can just put a `jnp.zeros(1)` there. So then JaxMarl envs will always return this global state.

Only extra thing we will need to do is change the observation spec in this class a bit.

mava/wrappers/jaxmarl.py
```diff
@@ -176,6 +183,7 @@ def step(
         obs, env_state, reward, done, infos = self._env.step(
             step_key, state.state, unbatchify(action, self.agents)
         )
+        infos["global_state"] = jnp.array(obs["world_state"])
```
sash-a: Same here for global state

mava/wrappers/jaxmarl.py
sash-a: If we can't get rid of the `hasattr` this is my suggestion
```suggestion
        # True if the wrapped env has a function to get the action mask
        has_action_mask = hasattr(self._env, "get_avail_actions")
```

mava/wrappers/jaxmarl.py
```diff
@@ -226,3 +234,8 @@ def discount_spec(self) -> specs.BoundedArray:
         return specs.BoundedArray(
             shape=(self.num_agents,), dtype=float, minimum=0.0, maximum=1.0, name="discount"
         )
+
+    def _get_avail_actions(self, state: JaxMarlState) -> Array:
+        avail_actions = self._env.get_avail_actions(state)
+        action_mask = jnp.array(batchify(avail_actions, self.agents), dtype=jnp.float32)
+        return action_mask
```
```suggestion
    def action_mask(self, state: JaxMarlState) -> Array:
        if self.has_action_mask:
          avail_actions = self._env.get_avail_actions(state)
          mask = jnp.array(batchify(avail_actions, self.agents), dtype=jnp.float32)
        else:
            mask = jnp.ones(self._action_shape)

        return mask
```

mava/wrappers/jaxmarl.py
```diff
@@ -157,16 +157,23 @@ def __init__(self, env: MultiAgentEnv, timelimit: int = 500):
...
+
     def reset(self, key: PRNGKey) -> Tuple[JaxMarlState, TimeStep[Observation]]:
         key, reset_key = jax.random.split(key)
         obs, state = self._env.reset(reset_key)
 
+        extras = {"global_state": jnp.array(obs["world_state"])}
         obs = Observation(
             agents_view=batchify(obs, self.agents),
-            action_mask=jnp.ones(self._action_shape),
+            action_mask=self.get_avail_actions(state),
```
```suggestion
            action_mask=self.action_mask(state),
```

mava/wrappers/jaxmarl.py
```diff
@@ -157,16 +157,23 @@ def __init__(self, env: MultiAgentEnv, timelimit: int = 500):
...
+        if hasattr(self._env, "get_avail_actions"):
+            self.get_avail_actions = self._get_avail_actions
+        else:
+            self.get_avail_actions = lambda _: jnp.ones(self._action_shape)
+
     def reset(self, key: PRNGKey) -> Tuple[JaxMarlState, TimeStep[Observation]]:
         key, reset_key = jax.random.split(key)
         obs, state = self._env.reset(reset_key)
 
+        extras = {"global_state": jnp.array(obs["world_state"])}
```
sash-a: Use `ObservationGlobalState`
```suggestion
```

mava/wrappers/shared.py
```diff
@@ -142,40 +142,39 @@ class GlobalStateWrapper(Wrapper):
...
-        global_state = jnp.concatenate(timestep.observation.agents_view, axis=0)
-        global_state = jnp.tile(global_state, (self._env.num_agents, 1))
-
-        observation = ObservationGlobalState(
-            global_state=global_state,
-            agents_view=timestep.observation.agents_view,
-            action_mask=timestep.observation.action_mask,
-            step_count=timestep.observation.step_count,
-        )
+    def __init__(self, env: Environment, add_agent_id: bool = False) -> None:
```
sash-a: I think we should leave this class untouched and just use the `ObservationGlobalState` directly in the `JaxMarl` wrapper, let me know if you think that will cause issues?

mava/utils/logger_tools.py
```diff
@@ -140,10 +145,14 @@ def get_neptune_logger(cfg: Dict) -> neptune.Run:
 
 def get_experiment_path(config: Dict, logger_type: str) -> str:
     """Helper function to create the experiment path."""
+    task_name = (
+        config["env"]["scenario"]
+        if isinstance(config["env"]["scenario"], str)
+        else config["env"]["scenario"]["task_name"]
```
sash-a: Is it possible to unify this in the config so that we don't need this `if`?

mava/utils/logger_tools.py
```diff
@@ -77,11 +78,15 @@ def _setup_json(self, cfg: Dict) -> None:
             json_logs_path = os.path.join(
                 cfg["logger"]["base_exp_path"], "json", cfg["logger"]["kwargs"]["json_path"]
             )
-
+        task_name = (
+            cfg["env"]["scenario"]
+            if isinstance(cfg["env"]["scenario"], str)
+            else cfg["env"]["scenario"]["task_name"]
```
sash-a: Same here

mava/evaluator.py
```diff
@@ -262,7 +277,7 @@ def evaluator_setup(
     """Initialise evaluator_fn."""
     # Get available TPU cores.
     n_devices = len(jax.devices())
-
+    win_rate = config["env"]["env_name"] == "HeuristicEnemySMAX"
```
sash-a: Slightly widens this condition if we use the other envs in the future

```suggestion
    win_rate = isinstance(env, (SMAX, HeuristicEnemySMAX, LearnedPolicyEnemySMAX))
```

mava/logger.py
```diff
@@ -58,6 +58,11 @@ def log(
         if absolute_metric:
             prefix = "absolute/"
             episodes_info = metrics.episodes_info
+            # Add win rate to episodes_info in case it exists.
+            if "won_episode" in episodes_info:
+                win_rate = (
+                    jnp.sum(episodes_info["won_episode"]) / (logger.num_eval_episodes * 10)
+                ) * 100
```
sash-a: nit
```suggestion
                n_won_episodes = jnp.sum(episodes_info["won_episode"])
                n_episodes = logger.num_eval_episodes * 10
                win_rate = (n_won_episodes / n_episodes) * 100
```

sash-a (changes requested):
Awesome work @OmaymaMahjoub! 

Main thing I think we should do is use the `ObservationGlobalState` directly, that should reduce the amount of changes a lot!

mava/wrappers/jaxmarl.py
OmaymaMahjoub: unfortunately, based on what I found I don't think there is a standard way to have actions mask, for exampla smax uses `get_avail_actions`, hanabi uses `get_legal_moves`  and mabrax doesn't have the method as it is continuous actions

OmaymaMahjoub: shall we keep it `_get_avail_actions` and solve this issue another time when we support more envs from jaxmarl?

sash-a: Ye this really sucks, keep it for now then, I think we'll have to open an issue on their repo

mava/wrappers/shared.py
```diff
@@ -120,11 +120,14 @@ def observation_spec(
         agents_view = specs.Array(
             (self._env.num_agents, num_obs_features), jnp.int32, "agents_view"
         )
+
+        if hasattr(obs_spec, "global_state"):
+            state_size = obs_spec.global_state.shape[-1] + self._env.num_agents
+        else:
+            state_size = num_obs_features * self._env.num_agents + self._env.num_agents
```
sash-a: I think we can collapse this condition into `self.has_global_state` we must just make sure has_global_state is always true for `JaxMarlEnvs`. And then also use your way of calculating state size (with the `obs_spec`) as it's more future proof.


```python
if self.has_global_state:
    wrapped_state_shape = obs_spec.global_state.shape
    state_shape = (*wrapped_state_shape[:-1], wrapped_state_shape[-1] + self._env.num_agents)
    global_state = specs.Array(state_shape, ...)
    return obs_spec.replace(agents_view=agents_view, global_state=global_state)

return obs_spec.replace(agents_view=agents_view)
```

mava/wrappers/jaxmarl.py
```diff
@@ -163,36 +163,36 @@ def __init__(self, env: MultiAgentEnv, timelimit: int = 500):
...
-        extras = {"global_state": jnp.array(obs["world_state"])}
-        obs = Observation(
+        obs = ObservationGlobalState(
             agents_view=batchify(obs, self.agents),
             action_mask=self.get_avail_actions(state),
+            global_state=self.get_global_state(obs),
             step_count=jnp.zeros(self._env.num_agents, dtype=int),
         )
-        return JaxMarlState(state, key, 0), restart(obs, extras=extras, shape=(self.num_agents,))
+        return JaxMarlState(state, key, 0), restart(obs, extras={}, shape=(self.num_agents,))
```
sash-a: extras default is `{}`
```suggestion
        return JaxMarlState(state, key, 0), restart(obs, shape=(self.num_agents,))
```

mava/utils/make_env.py
```diff
@@ -100,8 +100,9 @@ def make_jaxmarl_env(env_name: str, config: Dict) -> Tuple[Environment, Environm
     env = JaxMarlWrapper(jaxmarl.make(env_name, **kwargs))
     eval_env = JaxMarlWrapper(jaxmarl.make(env_name, **kwargs))
 
-    env = add_optional_wrappers(env, config)
-    eval_env = add_optional_wrappers(eval_env, config)
+    if config["system"]["add_agent_id"]:
+        env = AgentIDWrapper(env, config["system"]["add_global_state"])
+        eval_env = AgentIDWrapper(eval_env, config["system"]["add_global_state"])
```
sash-a, line 108: If you like the changes suggested below then just make these both `True` and put a comment that JaxMarlEnvs always have global state

sash-a (changes requested):
Nice changes! Just a couple more minor things :pray:

mava/evaluator.py
OmaymaMahjoub: The issue here is that env is not an instance of SMAX... as it is a wrapped env therefore this `isinstance` will be very nested by mentioning `env._env._env...` depending on how many wrappers are added

mava/wrappers/jaxmarl.py
OmaymaMahjoub: The default is None which causes an error for the type change:
```
TypeError: Scanned function carry input and carry output must have the same pytree structure, but they differ:
  * the input carry component learner_state.timestep.extras is a <class 'NoneType'> but the corresponding component of the carry output is a <class 'dict'>, so their Python types differ

Revise the scanned function so that its output is a pair where the first element has the same pytree structure as the first argument.
```

mava/evaluator.py
```diff
@@ -86,6 +90,11 @@ def not_done(carry: Tuple) -> bool:
             "episode_return": final_state.return_,
             "episode_length": final_state.step_count_,
         }
+        # Log won episode if win rate is required.
+        if win_rate:
+            eval_metrics["won_episode"] = jnp.where(
+                jnp.all(final_state.timestep.reward >= 1.0), 1, 0
+            )
```
sash-a: Slightly cleaner

```suggestion
            eval_metrics["won_episode"] = jnp.all(final_state.timestep.reward >= 1.0).astype(int)
```

mava/evaluator.py
```diff
@@ -192,6 +202,11 @@ def not_done(carry: Tuple) -> bool:
...
```
sash-a: same here

```suggestion
            eval_metrics["won_episode"] = jnp.all(final_state.timestep.reward >= 1.0).astype(int)
```

mava/utils/logger_tools.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, cfg: Dict) -> None:
         self.should_log = bool(
             cfg["logger"]["use_json"] or cfg["logger"]["use_tf"] or cfg["logger"]["use_neptune"]
         )
+        self.num_eval_episodes = cfg["arch"]["num_eval_episodes"]
```
sash-a, line 54: Why is this here? Doesn't seem like we use it

sash-a (dismissed):
Minor style change otherwise I'm super happy with this, great job!

mava/logger.py
```diff
@@ -58,6 +58,11 @@ def log(
         if absolute_metric:
             prefix = "absolute/"
             episodes_info = metrics.episodes_info
+            # Add win rate to episodes_info in case it exists.
+            if "won_episode" in episodes_info:
+                n_won_episodes = jnp.sum(episodes_info["won_episode"])
+                n_episodes = logger.num_eval_episodes * 10
```
callumtilbury, line 68: Could we please explain why *10 here?

mava/evaluator.py
callumtilbury: +1

mava/evaluator.py
```diff
@@ -262,7 +277,7 @@ def evaluator_setup(
     """Initialise evaluator_fn."""
     # Get available TPU cores.
     n_devices = len(jax.devices())
-
+    win_rate = config["env"]["env_name"] in ["HeuristicEnemySMAX", "LearnedPolicyEnemySMAX"]
```
callumtilbury: Small explanation here might be helpful, especially because the variable may be seen as "the win rate" itself.

mava/evaluator.py
```diff
@@ -275,28 +290,27 @@ def evaluator_setup(
...
         vmapped_eval_network_apply_fn = jax.vmap(
             network.apply,
             in_axes=(None, 0),
         )
-        evaluator = get_ff_evaluator_fn(
-            eval_env,
-            vmapped_eval_network_apply_fn,
-            config,
-        )
+        evaluator = get_ff_evaluator_fn(eval_env, vmapped_eval_network_apply_fn, config, win_rate)
```
callumtilbury: Perhaps same formatting as the others
```suggestion
        evaluator = get_ff_evaluator_fn(
            eval_env,
            vmapped_eval_network_apply_fn,
            config,
            win_rate,
        )
```

mava/utils/logger_tools.py
callumtilbury: It seems to be used here as a public member variable:
https://github.com/instadeepai/Mava/blob/771283f2327718248203d4b64728c3f77453e0f0/mava/logger.py#L64

mava/wrappers/jaxmarl.py
```diff
@@ -142,7 +142,7 @@ def jaxmarl_space_to_jumanji_spec(space: jaxmarl_spaces.Space) -> specs.Spec:
 class JaxMarlWrapper(Wrapper):
     """Wraps a JaxMarl environment so that its API is compatible with jumaji environments."""
```
callumtilbury:
```suggestion
    """Wraps a JaxMarl environment so that its API is compatible with Jumanji environments."""
```

mava/utils/logger_tools.py
OmaymaMahjoub: Because we need it to calculate the win rate in logger.py: `win_rate = (n_won_episodes / logger.num_eval_episodes) * 100`

mava/utils/make_env.py
```diff
@@ -94,14 +94,17 @@ def make_jaxmarl_env(env_name: str, config: Dict) -> Tuple[Environment, Environm
 
     kwargs = config["env"]["kwargs"]
     if "smax" in env_name.lower():
-        kwargs["scenario"] = map_name_to_scenario(config["env"]["scenario"])
+        kwargs["scenario"] = map_name_to_scenario(config["env"]["scenario"]["task_name"])
 
     # Placeholder for creating JAXMARL environment.
```
callumtilbury: Is this still just a 'placeholder'?

sash-a (dismissed):
:rocket:

mava/wrappers/jaxmarl.py
```diff
@@ -226,3 +262,16 @@ def discount_spec(self) -> specs.BoundedArray:
...
             shape=(self.num_agents,), dtype=float, minimum=0.0, maximum=1.0, name="discount"
         )
+
+    def action_mask(self, state: JaxMarlState) -> Array:
+        """Get action mask for each agent."""
+        if self.has_action_mask:
+            avail_actions = self._env.get_avail_actions(state)
+            mask = jnp.array(batchify(avail_actions, self.agents), dtype=jnp.float32)
+        else:
+            mask = jnp.ones(self._action_shape)
```
callumtilbury:
```suggestion
            mask = jnp.ones(self._action_shape, dtype=jnp.float32)
```
Just to be explicit + consistent with the line above :)

callumtilbury (commented):
A few minor points, but otherwise looks good :) Thanks @OmaymaMahjoub!

mava/logger.py
callumtilbury: (As discussed privately 👍🏻 )

mava/evaluator.py
callumtilbury: Or rename to e.g. `include_win_rate`?

mava/utils/make_env.py
callumtilbury: Sorry, this is pointing to the wrong location. See line 99 below 👍🏻

OmaymaMahjoub: I updated the comment to be more informative

mava/evaluator.py
```diff
@@ -35,7 +35,7 @@ def get_ff_evaluator_fn(
     env: Environment,
     apply_fn: ActorApply,
     config: dict,
-    win_rate: bool = False,
+    include_win_rate: bool = False,
```
sash-a: I like being more explicit here, sorry one final thing: 
```suggestion
    log_win_rate: bool = False,
```

mava/evaluator.py
```diff
@@ -91,7 +91,7 @@ def not_done(carry: Tuple) -> bool:
             "episode_length": final_state.step_count_,
         }
         # Log won episode if win rate is required.
-        if win_rate:
+        if include_win_rate:
```
sash-a:
```suggestion
        if log_win_rate:
```

mava/evaluator.py
```diff
@@ -131,7 +131,7 @@ def get_rnn_evaluator_fn(
     apply_fn: RecActorApply,
     config: dict,
     scanned_rnn: nn.Module,
-    win_rate: bool = False,
+    include_win_rate: bool = False,
```
sash-a:
```suggestion
    log_win_rate: bool = False,
```

mava/evaluator.py
```diff
@@ -201,7 +201,7 @@ def not_done(carry: Tuple) -> bool:
             "episode_length": final_state.step_count_,
         }
         # Log won episode if win rate is required.
-        if win_rate:
+        if include_win_rate:
```
sash-a:
```suggestion
        if log_win_rate:
```

mava/evaluator.py
```diff
@@ -273,40 +273,40 @@ def evaluator_setup(
     """Initialise evaluator_fn."""
     # Get available TPU cores.
     n_devices = len(jax.devices())
-    win_rate = config["env"]["env_name"] in ["HeuristicEnemySMAX", "LearnedPolicyEnemySMAX"]
+    # Check if win rate is required for evaluation.
+    include_win_rate = config["env"]["env_name"] in ["HeuristicEnemySMAX", "LearnedPolicyEnemySMAX"]
```
sash-a:
```suggestion
    log_win_rate = config["env"]["env_name"] in ["HeuristicEnemySMAX", "LearnedPolicyEnemySMAX"]
```

mava/evaluator.py
```diff
@@ -273,40 +273,40 @@ def evaluator_setup(
...
         )
-        evaluator = get_ff_evaluator_fn(eval_env, vmapped_eval_network_apply_fn, config, win_rate)
+        evaluator = get_ff_evaluator_fn(eval_env, vmapped_eval_apply_fn, config, include_win_rate)
         absolute_metric_evaluator = get_ff_evaluator_fn(
             eval_env,
-            vmapped_eval_network_apply_fn,
+            vmapped_eval_apply_fn,
             config,
-            win_rate,
+            include_win_rate,
```
sash-a:
```suggestion
            log_win_rate,
```

mava/evaluator.py
```diff
@@ -273,40 +273,40 @@ def evaluator_setup(
...
         )
         absolute_metric_evaluator = get_rnn_evaluator_fn(
             eval_env,
-            vmapped_eval_network_apply_fn,
+            vmapped_eval_apply_fn,
             config,
             scanned_rnn,
-            win_rate,
+            include_win_rate,
```
sash-a:
```suggestion
            log_win_rate,
```

mava/evaluator.py
```diff
@@ -273,40 +273,40 @@ def evaluator_setup(
...
             network.apply, in_axes=(None, 1, (2, None)), out_axes=(1, 2)
         )
         evaluator = get_rnn_evaluator_fn(
             eval_env,
-            vmapped_eval_network_apply_fn,
+            vmapped_eval_apply_fn,
             config,
             scanned_rnn,
-            win_rate,
+            include_win_rate,
```
sash-a:
```suggestion
            log_win_rate,
```

sash-a (changes requested):
Really small issue I have with the new variable name, sorry! After this I'm happy to :ship: it

sash-a (approved):
:rocket:

callumtilbury (approved):
🚗

PR #972, opened 2023-12-21
Enhancement: Make a `network.py` file

OmaymaMahjoub:
## What?
Make a shared file for the networks.
## Why?
To avoid redundancy across the single files systems, we made a shared file containing the networks. This approach ensures clarity and is easy to maintain.
## How?
Develop the `get_networks` function to enable user selection between feedforward or recurrent networks, and to specify whether the training is centralized or independent.


mava/networks.py
```diff
@@ -0,0 +1,250 @@
...
+                kernel_init=orthogonal(np.sqrt(2)),
+                bias_init=constant(0.0),
+            )(x)
+            if self.use_layer_norm:
+                x = nn.LayerNorm(use_scale=False)(x)
+            x = self.activation_fn(x)
+        return x
+
+
+class FF_Actor(nn.Module):
```
RuanJohn: In general, I think it would be better to have these names follow standard class naming conventions. Maybe `FFActor` or even more explicit like `FeedForwardActor`?

mava/networks.py
```diff
@@ -0,0 +1,250 @@
...
+
+from mava.types import (
+    Observation,
+    ObservationGlobalState,
+    RNNGlobalObservation,
+    RNNObservation,
+)
+
+
+class Torso(nn.Module):
```
RuanJohn: Can we call this `MLPTorso` please? Since later on we will have CNN and potentially Transformer torsos as well.

mava/networks.py
RuanJohn: Similarly elsewhere about the class naming.

mava/networks.py
```diff
@@ -0,0 +1,250 @@
...
+        actor_policy = distrax.Categorical(logits=masked_logits)
+
+        return actor_policy
+
+
+class FF_Critic(nn.Module):
+    """Feedforward Critic Network."""
+
+    torso: nn.Module
+    centralized_critic: bool = False
```
RuanJohn: NIT: can we please use British English? ie. `centralized` -> `centralised`

RuanJohn (dismissed):
Thank you for this @OmaymaMahjoub! I have a few minor comments

mava/configs/system/ff_ippo.yaml
```diff
@@ -20,3 +20,14 @@ clip_eps: 0.2 # Clipping value for PPO updates and value function.
...
+# ---Networks---
+actor_network:
+  layer_sizes: [128, 128]
+  use_layer_norm: True
+  activation: "relu"
+
+critic_network:
+  layer_sizes: [128, 128]
+  use_layer_norm: False
+  activation: "relu"
```
sash-a: Seeing as this is repeated, we should have a `config/networks/` folder and then we can specify different networks in there. So one file for mlp, another for recurrent another for transformer etc

mava/systems/ff_ippo.py
sash-a, line 1: So much cleaner :heart_eyes:

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+    layer_sizes: Sequence[int]
+    activation: str = "relu"
+    use_layer_norm: bool = False
+
+    def setup(self) -> None:
+        """Set up the activation function."""
+        if self.activation == "relu":
+            self.activation_fn = nn.relu
+        elif self.activation == "tanh":
+            self.activation_fn = nn.tanh
```
sash-a: I think this is cleaner because then we don't have to parse the string, we can just pass in the function directly

```suggestion
    activation: Callbale[[Array], Array] = nn.relu
    use_layer_norm: bool = False
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+
+    @nn.compact
+    def __call__(self, observation: Observation) -> distrax.Categorical:
+        """Forward pass."""
+        x = observation.agents_view
+
+        x = self.torso(x)
+        actor_output = nn.Dense(
+            self.num_actions, kernel_init=orthogonal(0.01), bias_init=constant(0.0)
+        )(x)
```
sash-a: Had a quick look and by default bias_init is zeros, so *I think* we don't need that part. While we're here though why do we do the `kernel_init`? I assume the default wouldn't make a huge difference?

```suggestion
        actor_output = nn.Dense(self.num_actions, kernel_init=orthogonal(0.01))(x)
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+            # Get global state in the case of a centralised critic.
+            observation = observation.global_state
+        else:
+            # Get single agent view in the case of a decentralised critic.
+            observation = observation.agents_view
+
+        critic_output = self.torso(observation)
+        critic_output = nn.Dense(1, kernel_init=orthogonal(1.0), bias_init=constant(0.0))(
+            critic_output
+        )
```
sash-a: Same here for bias

```suggestion
        critic_output = nn.Dense(1, kernel_init=orthogonal(1.0))(critic_output)
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+        """Forward pass."""
+        observation, done = observation_done
+
+        policy_embedding = self.pre_torso(observation.agents_view)
+        policy_rnn_in = (policy_embedding, done)
+        policy_hidden_state, policy_embedding = ScannedRNN()(policy_hidden_state, policy_rnn_in)
+        actor_output = self.post_torso(policy_embedding)
+        actor_output = nn.Dense(
+            self.action_dim, kernel_init=orthogonal(0.01), bias_init=constant(0.0)
+        )(actor_output)
```
sash-a: Same here for bias

```suggestion
        actor_output = nn.Dense(self.action_dim, kernel_init=orthogonal(0.01))(actor_output)
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+
+        masked_logits = jnp.where(
+            observation.action_mask,
+            actor_output,
+            jnp.finfo(jnp.float32).min,
+        )
+
+        actor_policy = distrax.Categorical(logits=masked_logits)
+
+        return actor_policy
```
sash-a: nit

```suggestion
        return distrax.Categorical(logits=masked_logits)
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+            # Get single agent view in the case of a decentralised critic.
+            observation = observation.agents_view
+
+        critic_embedding = self.pre_torso(observation)
+        critic_rnn_in = (critic_embedding, done)
+        critic_hidden_state, critic_embedding = ScannedRNN()(critic_hidden_state, critic_rnn_in)
+        critic_output = self.post_torso(critic_embedding)
+        critic_output = nn.Dense(1, kernel_init=orthogonal(1.0), bias_init=constant(0.0))(
+            critic_output
+        )
```
sash-a: Same here for bias

```suggestion
        critic_output = nn.Dense(1, kernel_init=orthogonal(1.0))(critic_output)
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+) -> Union[Tuple[FeedForwardActor, FeedForwardCritic], Tuple[RecurrentActor, RecurrentCritic]]:
+    """Get the networks."""
+
+    def create_torso(network_key: str, layer_size_key: str) -> MLPTorso:
+        """Helper function to create a torso object from the config."""
+        return MLPTorso(
+            layer_sizes=config["system"][network_key][layer_size_key],
+            activation=config["system"][network_key]["activation"],
+            use_layer_norm=config["system"][network_key]["use_layer_norm"],
+        )
```
sash-a: Can we add this for parsing activation, keeps it decoupled and adds ability to more easily add new activation functions

```suggestion
def parse_activation_fn(activation_fn_name: str):
    activation_fns = {"relu": nn.relu, "tanh": nn.tanh}
    return activation_fns[activation_fn_name]

def get_networks(
    config: Dict, network: str, centralised_critic: bool = False
) -> Union[Tuple[FeedForwardActor, FeedForwardCritic], Tuple[RecurrentActor, RecurrentCritic]]:
    """Get the networks."""

    def create_torso(network_key: str, layer_size_key: str) -> MLPTorso:
        """Helper function to create a torso object from the config."""
        activation_fn = parse_activation_fn(config["system"][network_key]["activation"])
            
        return MLPTorso(
            layer_sizes=config["system"][network_key][layer_size_key],
            activation=activation_fn,
            use_layer_norm=config["system"][network_key]["use_layer_norm"],
        )
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+        critic_hidden_state, critic_embedding = ScannedRNN()(critic_hidden_state, critic_rnn_in)
+        critic_output = self.post_torso(critic_embedding)
+        critic_output = nn.Dense(1, kernel_init=orthogonal(1.0), bias_init=constant(0.0))(
+            critic_output
+        )
+
+        return critic_hidden_state, jnp.squeeze(critic_output, axis=-1)
+
+
+def get_networks(
```
sash-a: Nit: can we change this to `make` because I see this becoming something like an env registry but for networks. So then we could use it like this:

```
from mava import networks
networks.make(...)
```

But I'm very open to opinions on this.

sash-a (changes requested):
Very nice and so much cleaner! Very minor comments on bias/initialization, activation functions and the method name, otherwise I think this is almost good to go!

mava/networks.py
WiemKhlifi: Or maybe to avoid setting it in all the calls of MLPTorso and to make it more flexible to control changes only  from the yaml file, we can change it to this:

```suggestion
    activation_fns = {"relu": nn.relu, "tanh": nn.tanh} # Add the necessary activation functions

    def setup(self) -> None:
        """Set up the activation function."""
        self.activation_fn = self.activation_fns[self.activation]

``` 
UPDATE: Also to avoid adding an extra function (parser) 😅

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+    torso: nn.Module
+    num_actions: Sequence[int]
+
+    @nn.compact
+    def __call__(self, observation: Observation) -> distrax.Categorical:
+        """Forward pass."""
+        x = observation.agents_view
+
+        x = self.torso(x)
+        actor_output = nn.Dense(
```
WiemKhlifi: Can we please change the `actor_output `to `actor_logits`?

```suggestion
actor_logits = nn.Dense(
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+            pre_torso=create_torso("actor_network", "pre_torso_layer_sizes"),
+            post_torso=create_torso("actor_network", "post_torso_layer_sizes"),
+        )
+        critic = RecurrentCritic(
+            pre_torso=create_torso("critic_network", "pre_torso_layer_sizes"),
+            post_torso=create_torso("critic_network", "post_torso_layer_sizes"),
+            centralised_critic=centralised_critic,
+        )
+    else:
+        raise ValueError(f"Network {config['system']['network']} not supported.")
```
WiemKhlifi: Small edits:
```suggestion
raise ValueError(f"The network '{network}' is not supported.")
```

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+        x = observation.agents_view
+
+        x = self.torso(x)
+        actor_output = nn.Dense(
+            self.num_actions, kernel_init=orthogonal(0.01), bias_init=constant(0.0)
+        )(x)
+
+        masked_logits = jnp.where(
+            observation.action_mask,
+            actor_output,
```
WiemKhlifi: Same : `actor_output `-> `actor_logits`

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+            if not isinstance(observation, ObservationGlobalState):
+                raise ValueError("Global state must be provided to the centralised critic.")
+            # Get global state in the case of a centralised critic.
+            observation = observation.global_state
+        else:
+            # Get single agent view in the case of a decentralised critic.
+            observation = observation.agents_view
+
+        critic_embedding = self.pre_torso(observation)
+        critic_rnn_in = (critic_embedding, done)
```
WiemKhlifi: NIT: `critic_rnn_in `-> `critic_rnn_input`

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+    def __call__(
+        self,
+        policy_hidden_state: chex.Array,
+        observation_done: RNNObservation,
+    ) -> Tuple[chex.Array, distrax.Categorical]:
+        """Forward pass."""
+        observation, done = observation_done
+
+        policy_embedding = self.pre_torso(observation.agents_view)
+        policy_rnn_in = (policy_embedding, done)
```
WiemKhlifi: NIT: `policy_rnn_in` -> `policy_rnn_input`

WiemKhlifi (changes requested):
Thank you Omayma for these changes 🔥 
I have only small changes to make names clearer 😅

mava/networks.py
```diff
@@ -0,0 +1,253 @@
...
+        if self.centralised_critic:
+            if not isinstance(observation, ObservationGlobalState):
+                raise ValueError("Global state must be provided to the centralised critic.")
+            # Get global state in the case of a centralised critic.
+            observation = observation.global_state
+        else:
+            # Get single agent view in the case of a decentralised critic.
+            observation = observation.agents_view
+
+        critic_output = self.torso(observation)
```
WiemKhlifi, line 94: `observation` -> `critic_input`:
```suggestion
            critic_input = observation.global_state
        else:
            # Get single agent view in the case of a decentralised critic.
            critic_input = observation.agents_view
    critic_output = self.torso(critic_input)
```

mava/configs/system/ff_ippo.yaml
OmaymaMahjoub: yeah! This is a good idea to scale in the future :raised_hands:

mava/networks.py
OmaymaMahjoub: I am convinced more with @WiemKhlifi solution, any additional thoughts on that @sash-a?

mava/networks.py
OmaymaMahjoub: I just moved the current network that we have without any investigation related to the net setup, I prefer we keep it with orthogonal initialization for now and investigate it because it is important for stability based on what I know and it was mentioned in the literature of PPO. Maybe we can make it as a feature :woman_shrugging:

mava/networks.py
OmaymaMahjoub: oh I didn't see the `parse_activation_fn` I think it is quite the same approach but just to make things nit I will edit based on @sash-a  suggestion :sweat_smile:

mava/networks.py
OmaymaMahjoub: I think it is a nice idea, just I wanted to double check your thoughts about keeping the other networks classes import as usual for example keeping `from mava.networks import ScannedRNN`  and not use `from mava import networks ... networks.ScannedRNN`

mava/networks.py
WiemKhlifi: Think that mine is shorter and lazier but sure pick what you prefer the most 😅

mava/networks.py
OmaymaMahjoub: I think keeping it as observation because critic_input will be confused with critic_rnn_input

mava/networks.py
sash-a: Ye I don't mind either :shrug: 

I think mine is slightly more future proof so that when we make `CnnTorso`/`TransformerTorso` we don't have to redo the function, but we can cross that bridge later if you like Wiem's more

mava/networks.py
sash-a: Cool, happy to keep the init, was just wondering if you knew why we did it

mava/networks.py
sash-a: hmmm that is up to you, I don't mind either way

mava/networks.py
WiemKhlifi: No yours is more convincing 😅 and better for future to avoid redundancy in defining the activation_fn attribute each time 🙌

mava/networks.py
WiemKhlifi: I think the second way is better to use `network.X` for better readability and maybe in future to avoid naming conflicts ( for e.g. now we have `make()` for envs and for networks)

OmaymaMahjoub: the issue with the second one we will have `networks.` used many times for unnecessary reasons

sash-a (approved):
:shipit:

mava/networks.py
WiemKhlifi: Humm true, so let's keep it as is now 🙌

WiemKhlifi (approved):
⛵

PR #973, opened 2023-12-22
Enhancement: change config to `DictConfig` instead of `Dict`

OmaymaMahjoub:
## What?
Implement config that supports Hydra syntax.
## Why?
To simplify code and improve readability, replacing cumbersome dictionary-based config access in Mava systems with Hydra's default syntax.
## How?
Change configuration references from config["key"]["subkey"] to config.key.subkey, enhancing code clarity and maintainability.## Extra
This PR is based on  #972 , therefore to review this PR we need to review  #972 first!
Close #949

PR #984, opened 2024-01-10
fix: neptune tags need to be a list

sash-a:
## What?
During #973 the type of config.logger.kwargs.neptune_tag was changed from a python list to an OmegaConf list and neptune was not happy with this
## Why?
So we can log to neptune


OmaymaMahjoub (approved):
Thanks for the fix @sash-a!

WiemKhlifi (approved):
Nice catch @sash-a 🚀

PR #985, opened 2024-01-11
fix: allow dynamic config setting, as done in the non-advanced systems

callumtilbury:
Trivial change, mirroring that which is done in the simple systems.


sash-a (approved):

OmaymaMahjoub (approved):
:fire:

PR #986, opened 2024-01-11
Fix: checkpointer

OmaymaMahjoub:
## What?
Fix the checkpointer to accept `dictconfig` and load non `FrozenDict` params
## Why?
Currently, when trying to add metadata config we will have an error as the config are not dict, besides an error when loading checkpointed params as the params type is not FrozenDict in the systems.
## How?
- [x] Change metadata to dict in case of DictConfig
- [x] Load params as FrozenDict in case the learner_state.params of type FrozenDict
- [x] Unreplicate params based on the devices and `update_batch_size` during storing

### DIFF
--- a/mava/advanced_usage/ff_ippo_store_experience.py
+++ b/mava/advanced_usage/ff_ippo_store_experience.py
@@ -27,7 +27,6 @@ import numpy as np
 import optax
 from colorama import Fore, Style
 from flashbax.vault import Vault
-from flax import jax_utils
 from flax.core.frozen_dict import FrozenDict
 from flax.linen.initializers import constant, orthogonal
 from jumanji.env import Environment
@@ -49,7 +48,7 @@ from mava.types import (
     PPOTransition,
 )
 from mava.utils.checkpointing import Checkpointer
-from mava.utils.jax import merge_leading_dims
+from mava.utils.jax import merge_leading_dims, unreplicate_learner_state
 from mava.utils.make_env import make
 
 StoreExpLearnerFn = Callable[[MavaState], Tuple[ExperimentOutput[MavaState], PPOTransition]]
@@ -449,6 +448,19 @@ def learner_setup(
     critic_params = critic_network.init(key_p, init_x)
     critic_opt_state = critic_optim.init(critic_params)
 
+    # Load model from checkpoint if specified.
+    if config.logger.checkpointing.load_model:
+        loaded_checkpoint = Checkpointer(
+            model_name=config.logger.system_name,
+            **config.logger.checkpointing.load_args,  # Other checkpoint args
+        )
+        # Restore the learner state from the checkpoint
+        restored_params, _ = loaded_checkpoint.restore_learner_state(
+            input_params=Params(actor_params, critic_params)
+        )
+        # Update the params
+        actor_params, critic_params = restored_params.actor_params, restored_params.critic_params
+
     # Vmap network apply function over number of agents.
     vmapped_actor_network_apply_fn = jax.vmap(
         actor_network.apply,
@@ -564,18 +576,6 @@ def run_experiment(_config: DictConfig) -> None:  # noqa: CCR001
             model_name=config.logger.system_name,
             **config.logger.checkpointing.save_args,  # Checkpoint args
         )
-
-    if config.logger.checkpointing.load_model:
-        loaded_checkpoint = Checkpointer(
-            model_name=config.logger.system_name,
-            **config.logger.checkpointing.load_args,  # Other checkpoint args
-        )
-        # Restore the learner state from the checkpoint
-        learner_state_reloaded = loaded_checkpoint.restore_learner_state(
-            unreplicated_input_learner_state=jax_utils.unreplicate(learner_state)
-        )
-        # Overwrite learner state with reloaded state, and replicate across devices.
-        learner_state = jax.device_put_replicated(learner_state_reloaded, jax.devices())
 
     dummy_flashbax_transition = {
         "done": jnp.zeros((config.system.num_agents,), dtype=bool),
@@ -706,7 +706,7 @@ def run_experiment(_config: DictConfig) -> None:  # noqa: CCR001
             # Save checkpoint of learner state
             checkpointer.save(
                 timestep=steps_per_rollout * (i + 1),
-                unreplicated_learner_state=jax_utils.unreplicate(learner_output.learner_state),
+                unreplicated_learner_state=unreplicate_learner_state(learner_output.learner_state),
                 episode_return=episode_return,
             )
 
--- a/mava/advanced_usage/ff_ippo_store_experience.py
+++ b/mava/advanced_usage/ff_ippo_store_experience.py
@@ -455,7 +455,7 @@ def learner_setup(
             **config.logger.checkpointing.load_args,  # Other checkpoint args
         )
         # Restore the learner state from the checkpoint
-        restored_params, _ = loaded_checkpoint.restore_learner_state(
+        restored_params, _ = loaded_checkpoint.restore_params(
             input_params=Params(actor_params, critic_params)
         )
         # Update the params
--- a/mava/utils/checkpointing.py
+++ b/mava/utils/checkpointing.py
@@ -159,7 +159,6 @@ class Checkpointer:
         timestep: Optional[int] = None,
         restore_params: bool = True,
         restore_hstates: bool = True,
-        convert_to_frozen_dict: bool = False,
     ) -> Union[LearnerState, RNNLearnerState]:
         """Restore the learner state.
 
@@ -169,8 +168,6 @@ class Checkpointer:
                 Defaults to None, in which case the latest step will be used.
             restore_params (bool, optional): Whether to restore the params.
             restore_hstates (bool, optional): Whether to restore the hidden states.
-            convert_to_frozen_dict (bool, optional): Whether to convert the params and hstates
-            to FrozenDicts.
 
         Returns:
             Union[LearnerState, RNNLearnerState]: the restored learner state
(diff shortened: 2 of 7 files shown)


callumtilbury (commented):
This is brilliant, @OmaymaMahjoub! The proposed changes (primarily _where_ we reload the params in the code) make the checkpointing much cleaner—thank you.

A few ideas:
- We can now remove the following block:
https://github.com/instadeepai/Mava/blob/15b2106c1a48a689bba2ba19e94ed060f3d1fbe0/mava/utils/checkpointing.py#L142-L145
We now reconstruct the learner state manually, which means we don't have to worry about saving the type of our state (FF or RNN). I like this, as the alternative always felt a bit hacky :)

- We probably shouldn't call our method `restore_learner_state` anymore, since it returns params only (i.e. not opt_state etc.) I think this is good, but let's rename to match? I feel that `restore_params` is fine (even though `hstates` are also restored sometimes), but open to other ideas!

- Along with above, I'm wondering if we should even be checkpointing the entire learner state (`unreplicated_learner_state`), instead of just the params/hstates?
    - We shouldn't: because we're only loading the params anyway, so why save unnecessary stuff like opt_state?
    - We should: because in the future we might want to load the opt_state again, or just have more info at hand. Thoughts?

Otherwise, great stuff! Thanks agian @OmaymaMahjoub 🚀

OmaymaMahjoub:
> This is brilliant, @OmaymaMahjoub! The proposed changes (primarily _where_ we reload the params in the code) make the checkpointing much cleaner—thank you.
> 
> A few ideas:
> 
> * We can now remove the following block:
>   https://github.com/instadeepai/Mava/blob/15b2106c1a48a689bba2ba19e94ed060f3d1fbe0/mava/utils/checkpointing.py#L142-L145
>   
>   We now reconstruct the learner state manually, which means we don't have to worry about saving the type of our state (FF or RNN). I like this, as the alternative always felt a bit hacky :)
> * We probably shouldn't call our method `restore_learner_state` anymore, since it returns params only (i.e. not opt_state etc.) I think this is good, but let's rename to match? I feel that `restore_params` is fine (even though `hstates` are also restored sometimes), but open to other ideas!
> * Along with above, I'm wondering if we should even be checkpointing the entire learner state (`unreplicated_learner_state`), instead of just the params/hstates?
>   
>   * We shouldn't: because we're only loading the params anyway, so why save unnecessary stuff like opt_state?
>   * We should: because in the future we might want to load the opt_state again, or just have more info at hand. Thoughts?
> 
> Otherwise, great stuff! Thanks agian @OmaymaMahjoub 🚀

Thanks  @callumtilbury for the review! I addressed the first two points. For the third point yeah I kept the storing of learning_state as I thought maybe in the future it will be helpful however keen to hear the others' thoughts on that!

mava/utils/jax.py
```diff
@@ -42,3 +47,18 @@ def merge_leading_dims(x: chex.Array, num_dims: chex.Numeric) -> chex.Array:
...
+    """Unreplicates a learner state.
+
+    This function takes a learner state, and removes the axes associated with device replication
+    and the `update batch size`. The unreplication process is essential to store the learner state
+    in a checkpoint.
+
+    Note:
+        The function internally uses `jax_utils.unreplicate` twice to remove the necessary axes.
+    """
+    return jax_utils.unreplicate(jax_utils.unreplicate(learner_state))  # type: ignore
```
callumtilbury: Perhaps we can replace this with a single tree map? Not a bottleneck in speed, but just a little cleaner :)
```suggestion
    return jax.tree_map(lambda x: x[0][0], learner_state)
```

Confirmed that these are equivalent:
```
chex.assert_trees_all_equal(
    jax.tree_map(lambda x: x[0][0], learner_state),
    jax_utils.unreplicate(jax_utils.unreplicate(learner_state)),
)
```

mava/utils/jax.py
```diff
@@ -42,3 +47,18 @@ def merge_leading_dims(x: chex.Array, num_dims: chex.Numeric) -> chex.Array:
...
+    learner_state: Union[LearnerState, RNNLearnerState]
+) -> Union[LearnerState, RNNLearnerState]:
+    """Unreplicates a learner state.
+
+    This function takes a learner state, and removes the axes associated with device replication
+    and the `update batch size`. The unreplication process is essential to store the learner state
+    in a checkpoint.
+
+    Note:
+        The function internally uses `jax_utils.unreplicate` twice to remove the necessary axes.
```
callumtilbury: To update, if we do change below.

callumtilbury (commented):
Very happy with this, thanks @OmaymaMahjoub!

Re: the third point from before—I've thought about it & looked at your changes, and I think saving the entire `learner_state` is a good idea. We have a method now called `restore_params`, so perhaps in the future we can add other methods, e.g. `restore_opt_state`, both restoring from the same underlying checkpoint. So this is great 💯 

One minor suggestion for the 2 x unreplicate to use a single tree map. Otherwise, I am happy!

mava/utils/jax.py
callumtilbury: Actually, I have another recommendation—give me a sec :)

callumtilbury:
```suggestion
def unreplicate_learner_state(
    learner_state: Union[LearnerState, RNNLearnerState],
    unreplicate_depth: int = 2,
) -> Union[LearnerState, RNNLearnerState]:
    """Unreplicates a learner state.

    This function takes a learner state and removes some number of axes, associated with parameter 
    duplication. This is typically one axis for device replication, and one for the `update batch size`.
    """
    return jax.tree_map(lambda x: x[(0,) * unreplicate_depth], learner_state)
```

callumtilbury: see below :)

OmaymaMahjoub: much cleaner thanks for the suggestion

OmaymaMahjoub: I couldn't commit suggestion as the code was outdated so I tried to make co-author commit but it didn't work as it should :sweat_smile: https://github.com/instadeepai/Mava/pull/986/commits/5d426dc259318077a24f88ad76bdcac606b873a8

callumtilbury (dismissed):
Happy! 🚀

mava/advanced_usage/ff_ippo_store_experience.py
```diff
@@ -449,6 +448,19 @@ def learner_setup(
     critic_params = critic_network.init(key_p, init_x)
     critic_opt_state = critic_optim.init(critic_params)
 
+    # Load model from checkpoint if specified.
+    if config.logger.checkpointing.load_model:
```
sash-a, line 452: Why is checkpointing inside the logger config? I think we can give it its own config? Or make a common base system config with checkpointing inside it?

mava/utils/checkpointing.py
```diff
@@ -182,25 +175,21 @@ def restore_learner_state(
...
-        if restore_hstates and restored_learner_state_type == RNNLearnerState:
-            new_learner_state = typing.cast(RNNLearnerState, new_learner_state)  # for mypy
-            new_learner_state = new_learner_state._replace(
-                hstates=HiddenStates(**FrozenDict(restored_learner_state_raw["hstates"])),
-            )
+        # Restore params
+        if isinstance(input_params.actor_params, FrozenDict):
+            restored_params = Params(**FrozenDict(restored_learner_state_raw["params"]))
+        else:
+            restored_params = Params(**restored_learner_state_raw["params"])
```
sash-a, line 185: What's the difference here? Why would we store two different cases?

sash-a (changes requested):
Happy with this, just one question and then a config change :smile:

mava/advanced_usage/ff_ippo_store_experience.py
OmaymaMahjoub: I did it in the logger_setup in order to not broadcast the init params and then we re-do that with checkpointed ones

mava/utils/checkpointing.py
callumtilbury: Previously, it was necessary to unpack the params into a FrozenDict, and we used to _only_ do:
```
Params(**FrozenDict(restored_learner_state_raw["params"]))
```
But @OmaymaMahjoub found that recently, the params are coming in as dicts. We don't really understand the change (perhaps a new flax version?), but I believe we're checking for both possibilities in case of a change later.

callumtilbury: Though perhaps it is better to figure out what is happening, and choose one approach 🤔

OmaymaMahjoub: From my side I think we can keep it as a sanity check

sash-a: Ok then can we just document this, basically put exact what you've written above - flax now returns dicts and we're being careful?

mava/advanced_usage/ff_ippo_store_experience.py
sash-a: No but the config itself, why does checkpointing live inside logging? If I want to configure checkpointing I shouldn't have to go into the logging folder :thinking:

OmaymaMahjoub: oh I see your point :sweat_smile: do you think it will be better to have separate one?

OmaymaMahjoub: @callumtilbury your thoughts on that? As you are the person who add it to the logger maybe you have a reason behind that

callumtilbury: Yeah, I recall having a discussion about it, but unsure why we settled on that location. Can you remember @RuanJohn? Happy for it to go elsewhere, if we're all good with that :)

mava/utils/checkpointing.py
OmaymaMahjoub: the issue is that we are not 100% sure the reason was flax but I will try to downgrade it to the old version and see if it was the reason behind this

mava/advanced_usage/ff_ippo_store_experience.py
sash-a: Chatted to @RuanJohn, we're happy to leave as is :smile:

mava/utils/checkpointing.py
OmaymaMahjoub: Updates yeah flax 0.6.11 was having the networks as FrozenDict

OmaymaMahjoub: I updated the docs, feel free to resolve the conversation once you double check it

sash-a (dismissed):
:ship:

mava/advanced_usage/ff_ippo_store_experience.py
OmaymaMahjoub: sounds great :)

mava/utils/checkpointing.py
```diff
@@ -182,25 +175,24 @@ def restore_learner_state(
...
-
-        if restore_hstates and restored_learner_state_type == RNNLearnerState:
-            new_learner_state = typing.cast(RNNLearnerState, new_learner_state)  # for mypy
-            new_learner_state = new_learner_state._replace(
-                hstates=HiddenStates(**FrozenDict(restored_learner_state_raw["hstates"])),
-            )
+        # Check the type of `input_params` for compatibility.
+        # This is a sanity check to ensure correct handling of parameter types.
+        # In Flax 0.6.11, parameters were typically of the `FrozenDict` type,
+        # but in cu versions, a regular dictionary is used.
```
callumtilbury:
```suggestion
        # but in current versions, a regular dictionary is used.
```
Is this correct?

callumtilbury (dismissed):
One typo check, otherwise let's goooo 📦

mava/utils/checkpointing.py
OmaymaMahjoub: either current version or later versions

callumtilbury (dismissed):
💯

callumtilbury (approved):

sash-a (approved):

PR #987, opened 2024-01-12
feat: remove all systems'defaults and add a common default file

WiemKhlifi:
## What?
Integrate all system defaults and the system's logger defaults into a single `defaults.yaml` file.

## Why?
This helps simplify the config's folder and avoids filling it with multiple default files for each new system.

## How?
- Create one file containing all the common elements for all files and use it as default for all systems.
- Initialise the config based on this default file then add all the specific elements for each system from their `system.yaml` files as well as for the `network.yaml` file.

**Note:**  There is another approach (shorter one) mentioned [here](https://github.com/instadeepai/Mava/blob/ee4ce31d59ce20b3d8cae3bb71c1de5562ee3471/mava/configs/defaults.yaml#L4) and [here](https://github.com/instadeepai/Mava/blob/ee4ce31d59ce20b3d8cae3bb71c1de5562ee3471/mava/configs/system/ff_ippo.yaml#L3).


mava/advanced_usage/ff_ippo_store_experience.py
```diff
@@ -749,8 +749,16 @@ def hydra_entry_point(cfg: DictConfig) -> None:
...
+    # Load and merge the system and network config files with the original one
+    system_config_path = f"{hydra.utils.get_original_cwd()}/mava/configs/system/ff_ippo.yaml"
+    system_config = OmegaConf.load(system_config_path)
+    cfg.system = OmegaConf.merge(cfg.system, system_config)
+
+    network_config_path = (
+        f"{hydra.utils.get_original_cwd()}/mava/configs/network/{cfg.system.network_name}.yaml"
+    )
+    network_config = OmegaConf.load(network_config_path)
+    cfg["network"] = network_config
```
sash-a: I'm not a fan of this I think it's too much friction if we want to add configs in the future or if a name gets changed, but I do think the goal of this PR is good and we can do a lot of it with normal hydra.

sash-a (changes requested):
Searching for configs with strings is not a good idea, we should just let hydra take care of this and structure our configs a bit better.

I like that you've removed the per system log files, I think we should just have a single logging config. But I don't like that you've merged this config with the base the config, base config should still import this and it should be accessible through `cfg.logging.X`

I don't like that you've deleted the `default_ff_ippo.yaml`, etc. But I do like that you've started to move out some of the common parameters from the PPO files. What would be really nice here is add num_updates, seed and total_timesteps into arch.

If we wanted we could also make a `default_ppo.yaml` and have all the ppo configs inherit from this, not sure what your thoughts are on this?

WiemKhlifi:
@sash-a I agree I don't like it either but we can avoid that if we only change the system name in the default before running the `system.py` file. I wouldn't say I like these two paths and the extra effort, but they're better than the old way.

So, based on what you said,  I made a few changes based on that: add the num_updates, seed and total_timesteps to `arch.yaml`, and separate the logging params in the `logging.yaml` file. 

But what remains unchanged is our use of different default configs for each system where the only difference is the system name (the network can be included as part of the default in the system.yaml`).

Thus, I made only one default for all systems and we only need to change the system name before running the system.py file OR just run this command and it will do the job:
```python
python sytem_name.py system=system_name
```

mava/advanced_usage/ff_ippo_store_experience.py
WiemKhlifi: This code has been removed in the latest update.

WiemKhlifi:
This will be taken care of after finishing our measure sets. An issue will be opened for this matter 😄

WiemKhlifi:
This will be taken care of after finishing our measure sets. An issue will be opened for this matter 😄

PR #989, opened 2024-01-15
feat: simplify logger

sash-a:
## What?
Loggers were getting a bit complicated so I've simplified them.

## Why?
Keep code complexity as low as possible.

## How?

With the refactor I wanted to achieve two things:
* Allow for logging of arbitrary dictionaries so that future metrics are easy to add.
* Make the loggers composable - see `MultiLogger`.

Big change is experiment output now looks like this:

```
class ExperimentOutput(NamedTuple, Generic[MavaState]):
    """Experiment output."""

    learner_state: MavaState
    episode_metrics: Dict[str, chex.Array]
    train_metrics: Dict[str, chex.Array]
```

So our train metrics are a dict and no longer explicit - `value_loss`, `actor_loss` etc.

**Note**: I removed `mava/logger.py` and put it's contents in `mava/utils/logger.py`. I think this is better because I kept getting confused as to where to look for logging stuff, but open to opinions on this :smile:


sash-a:
todo: check this works with MarlEval

mava/evaluator.py
```diff
@@ -127,7 +128,7 @@ def evaluator_fn(trained_params: FrozenDict, key: chex.PRNGKey) -> ExperimentOut
             axis_name="eval_batch",
         )(trained_params, eval_state)
 
-        return ExperimentOutput(episodes_info=eval_metrics, learner_state=eval_state)
+        return ExperimentOutput(eval_state, eval_metrics, train_metrics={})
```
callumtilbury: Could we use named arguments please?
```suggestion
        return ExperimentOutput(
            learner_state=eval_state,
            episode_metrics=eval_metrics,
            train_metrics={},
        )
```

mava/evaluator.py
```diff
@@ -261,10 +262,7 @@ def evaluator_fn(
             axis_name="eval_batch",
         )(trained_params, eval_state)
 
-        return ExperimentOutput(
-            episodes_info=eval_metrics,
-            learner_state=eval_state,
-        )
+        return ExperimentOutput(eval_state, eval_metrics, train_metrics={})
```
callumtilbury: as above
```suggestion
        return ExperimentOutput(
            learner_state=eval_state,
            episode_metrics=eval_metrics,
            train_metrics={},
        )
```

mava/systems/ff_ippo.py
```diff
@@ -517,10 +514,12 @@ def run_experiment(_config: DictConfig) -> None:
...
         elapsed_time = time.time() - start_time
-        learner_output.episodes_info["steps_per_second"] = steps_per_rollout / elapsed_time
-        logger.log_trainer_metrics(
-            experiment_output=learner_output, t_env=steps_per_rollout * (i + 1)
-        )
+        t = steps_per_rollout * (eval_step + 1)
+        learner_output.episode_metrics["steps_per_second"] = steps_per_rollout / elapsed_time
+        learner_output.episode_metrics["timestep"] = t
+
+        logger.log(learner_output.episode_metrics, t, eval_step, LogEvent.ACT)
```
callumtilbury, line 522: How do you feel about named arguments here (and all the `logger.log` calls) too? I realise it can get messy, but I am a big fan in almost all situations.

callumtilbury (dismissed):
Fantastic work! This is great.

Minor request for named args in fn calls, but either way, I am happy for us to 🚢!

mava/systems/ff_ippo.py
callumtilbury: hmm, but looking at it again, perhaps it is too messy.

callumtilbury (dismissed):

mava/systems/ff_ippo.py
sash-a: Agreed, happy with it in most cases, especially when args aren't clear, but in this case it's too messy to be worth it. Others I am happy with :smile:

RuanJohn (changes requested):
Just a general request before we merge this PR, can you please run an experiment and log to JSON and Neptune? Please share a link on the PR to the Neptune logs and could you please share the logged JSON file offline? Ideally can you please have 2 algorithms with ~5 seeds each on two RWARE scenarios. The the runs don't need to be very long maybe just 1M steps each. I just want to make sure everything still works well with MARL-eval.

sash-a:
[Neptune runs](https://app.neptune.ai/o/InstaDeep/org/Mava/runs/compare?viewId=standard-view&detailsTab=charts&shortId=MAV-23439&dash=charts&query=((%60sys%2Ftags%60%3AstringSet%20CONTAINS%20%22simplify-logger%22))&sortBy=%5B%22sys%2Fcreation_time%22%5D&sortFieldType=%5B%22datetime%22%5D&sortFieldAggregationMode=%5B%22auto%22%5D&sortDirection=%5B%22descending%22%5D&suggestionsEnabled=true&lbViewUnpacked=true&compare=EwZgLCCcA0CMcPiRig)

RuanJohn (approved):
Super nice improvement @sash-a :raised_hands: 
Can confirm, Neptune logs are looking good and the JSON files work with marl-eval. 
:ship:

OmaymaMahjoub (approved):
Huge effort! Thanks @sash-a :fire:

PR #991, opened 2024-01-17
feat: add time_limit to jumanji env config

WiemKhlifi:
## What?
Add the time_limit attribute to Jumanji's environments.
## Why?
This helps control the `time_limit` and avoid using the default one.
## How?
Extract this attribute from config and feed it to the make function as `jumanji.make(...., **config.env.kwargs)`.


sash-a (approved):
Really useful addition, thanks for this! :smile:

callumtilbury (approved):
Nice!

PR #992, opened 2024-01-18
fix: use new vault api

callumtilbury:
## What?
Use the new (proposed) Vault API, with some minor changes.

## Why?
For it to work :^)

## How?
Simple changes.

## Extra
n/a


sash-a (dismissed):
LGTM :shield:

WiemKhlifi (changes requested):
The changes LGTM but I have a few suggestions unrelated to these and I can't add comments there 😅 
1- Update the network: This folder wasn't updated I guess to use the actor and critic from the network.py ( I can help with that).
2- This line was repeated twice I guess it's the default merge's fault:
```python
# Allow dynamic attributes.
  OmegaConf.set_struct(cfg, False)
```

callumtilbury:
Well spotted, thanks @WiemKhlifi!

sash-a (approved):
Nice spot @WiemKhlifi

WiemKhlifi (approved):
Thanks guys this LGTM 🥳

PR #994, opened 2024-01-18
feat: stop accessing LogEnvState variables directly

sash-a:
## What?
Stop accessing `LogEnvState` variables directly as this forces us to always make the `LogEnvWrapper` the outermost wrapper - so we can access it's variables through `state.episode_return` instead of `state.env_state.episode_return` and if we have an unknown number of wrappers we can't know how deep in the `env_state` chain we have to go. So instead pass the variables we care about through `timestep.extras`.

Also changed some file names to make stuff more clear :smile: 
Also renamed the `LogEnvState` to bring it more inline with the gym wrapper that does similar things.

## Why?
Allows for vault wrapper to also not be the outermost wrapper if we start passing the state info through `timestep.extras`


callumtilbury (approved):
Great stuff! Thanks @sash-a! 🚀

callumtilbury (approved):
Great stuff! Thanks @sash-a! 🚀

mava/wrappers/episode_metrics.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class RecordEpisodeMetrics(Wrapper):
+    """Record the episode returns and lengths."""
+
+    def reset(self, key: chex.PRNGKey) -> Tuple[RecordEpisodeMetricsState, TimeStep]:
+        """Reset the environment."""
+        state, timestep = self._env.reset(key)
+        state = RecordEpisodeMetricsState(state, 0.0, 0, 0.0, 0)
+        timestep.extras["episode_metrics"] = {"episode_return": 0.0, "episode_length": 0}
```
callumtilbury, line 50: Sorry, I missed this first time around! I'm hitting an error:
```
  File "/Users/callum/Mava/mava/systems/ff_ippo.py", line 584, in hydra_entry_point
    run_experiment(cfg)
  File "/Users/callum/Mava/mava/systems/ff_ippo.py", line 457, in run_experiment
    learn, actor_network, learner_state = learner_setup(
  File "/Users/callum/Mava/mava/systems/ff_ippo.py", line 404, in learner_setup
    env_states, timesteps = jax.vmap(env.reset, in_axes=(0))(
  File "/Users/callum/Mava/mava/wrappers/episode_metrics.py", line 50, in reset
    timestep.extras["episode_metrics"] = {"episode_return": 0.0, "episode_length": 0}
TypeError: 'NoneType' object does not support item assignment
```

And I think we ought to write:
```suggestion
        if timestep.extras is None:
            timestep.extras = {}
        timestep.extras["episode_metrics"] = {"episode_return": 0.0, "episode_length": 0}
```

callumtilbury (changes requested):

mava/wrappers/episode_metrics.py
callumtilbury: Solved by updating Jumanji version :)

callumtilbury (dismissed):

mava/wrappers/marl.py
OmaymaMahjoub, line 1: Maybe `observation_wrappers.py` is better as a name for this file

mava/wrappers/episode_metrics.py
OmaymaMahjoub, line 1: If you can check the possibility of using this during evaluation, it will make evaluation much easier to read and maybe having same type of output as the trainer :thinking:

OmaymaMahjoub (changes requested):
Thanks @sash-a! Happy to approve the PR once the comments are addressed!

mava/wrappers/episode_metrics.py
sash-a: Yes, so I did look into this, wanted to use the `TrainState` also. Only issue is that the evaluator currently takes in `EvalState`, `params` and an `apply_fn` which I think makes it very nice and general. If we use `TrainState` then we'd use the params in the train state and so it would be a bit less general, but it is something worth looking into. Maybe I can make an issue for this as it's quite a big change?

mava/wrappers/marl.py
sash-a: Nice, much better name :smile:

sash-a: Called it `observation.py` as we're already in the `wrappers` folder and the rest of the files don't have the _wrapper part, hope that's ok :smile:

mava/wrappers/episode_metrics.py
OmaymaMahjoub: yeah if you can make an issue related to that as I am currently not satisfied with our evaluator code, it must be there another better way to handle things within it :sweat_smile:

OmaymaMahjoub (dismissed):
Thanks @sash-a :fire:

RuanJohn (dismissed):
LGTM :ship:

OmaymaMahjoub (changes requested):
If you can double check the quickstart notebook working properly else happy to approve :raised_hands:

OmaymaMahjoub:
> If you can double check the quickstart notebook working properly else happy to approve 🙌

I tested the notebook and it raises some errors, the main changes should be fixing the imports, getting the episode return from timesteps.extras, and using `RecordEpisodeMetricsState` instead of `LogWrapper`

RuanJohn (dismissed):
:spider:

OmaymaMahjoub (approved):
I fixed the quickstart notebook :sweat_smile:, thanks @sash-a for this PR :fire:

RuanJohn (approved):
Thank you @sash-a :raised_hands:

PR #995, opened 2024-01-24
fix: bug fixes + add missing kwargs

WiemKhlifi:
## What?
Set the appropriate type for `episode_return` in the `Checkpointer` class, fix the `ExperimentOutput` type in the quickstart notebook, and add missing kwargs in the `jumanji_make_env` function.
## Why?
The changes made have resolved the bugs identified.


examples/Quickstart.ipynb
OmaymaMahjoub, line 1: @WiemKhlifi  double check the rware version that you are using, as I can see from the rendering and results that you are using the old version of rware before the fix that was made in jumanji.

mava/utils/make_env.py
```diff
@@ -72,10 +72,9 @@ def make_jumanji_env(
     wrapper = _jumanji_registry[env_name]["wrapper"]
 
     # Create envs.
-    env = jumanji.make(env_name, generator=generator)
-    env = wrapper(env)
+    env = jumanji.make(env_name, generator=generator, **config.env.kwargs)
```
OmaymaMahjoub, line 75: Nice catch!

OmaymaMahjoub (changes requested):
Thanks @WiemKhlifi, if you can rerun the notebook with the correct rware version and from there happy to approve the PR

WiemKhlifi:
All checks are done 🙌

mava/utils/logger.py
```diff
@@ -393,18 +393,19 @@ def write(
...
-            step_metrics = {"step_count": timestep, "elapsed_time": current_time - self.start_time}
-            step_metrics.update(metrics)  # type: ignore
+        if logging_prefix == "absolute":
+            self.run_data["absolute_metrics"].update(metrics)
+
+        elif logging_prefix == "evaluator":
+            step_metrics = {
+                "step_count": timestep,
+                "elapsed_time": current_time - self.start_time,
+            } | metrics
```
sash-a, line 403: I definitely change this in #994. Can we get that in first? And then make the change on top of that?

sash-a (commented):
Nice, I like this just need to wait for #994 I think

mava/utils/logger.py
sash-a: Scratch that I was thinking of the other logging PR, so never mind this. Will approve :smile:

sash-a (approved):
:fire_engine: :fire:

OmaymaMahjoub (approved):
Thanks @WiemKhlifi  LGTM!

PR #997, opened 2024-01-29
Feat/store json to neptune

RuanJohn:
## What?
Add functionality to upload json data as zipped files to Neptune.
## Why?
This allows for storing json data from many distributed experiments runs in a central place. GIven a Neptune tag the json files can then be downloaded, joined, aggregated and plotted with tooling from `MARL-eval`
## How?
Add a `zip_and_upload_json` method to the Neptune logger. 
## Extra
Simplified the path where json files are stored locally.


mava/utils/logger.py
```diff
@@ -281,6 +302,18 @@ def _make_multi_logger(cfg: DictConfig) -> BaseLogger:
     loggers: List[BaseLogger] = []
     unique_token = datetime.now().strftime("%Y%m%d%H%M%S")
 
+    if (
+        cfg.logger.use_neptune
+        and cfg.logger.use_json
+        and cfg.logger.kwargs.upload_json_data
+        and cfg.logger.kwargs.json_path
+    ):
+        raise ValueError(
```
OmaymaMahjoub, line 311: Maybe instead of raising an error, we raise a warning and change the json_path to empty str?

RuanJohn: I'm not super sure about this because it could be that someone is doing a big run on a some local hardware where they just want the single json file at the end. In which case it makes more sense to not upload json files instead of storing many.

OmaymaMahjoub: I see! yeah this is reasonable

RuanJohn: I also made the error message a bit more descriptive :)

mava/utils/logger.py
```diff
@@ -281,6 +302,20 @@ def _make_multi_logger(cfg: DictConfig) -> BaseLogger:
...
+            "Cannot upload json data to Neptune when `json_path` is set in the base logger config. "
+            "This is because each subsequent run will create a larger json file which will use "
+            "unnecessary storage. Either set `upload_json_data` if you don't want to upload your "
+            "json data but store a large file locally or remove `json_path` from the base logger "
+            "config."
```
sash-a: Improved the wording a little bit :smile: 

```suggestion
            "unnecessary storage. Either set `upload_json_data: false` if you don't want to "
            "upload your json data but store a large file locally or set `json_path: ~` in "
            "the base logger config."
```

OmaymaMahjoub (dismissed):
LGTM! Thanks @RuanJohn :fire:

mava/utils/logger.py
```diff
@@ -295,11 +330,7 @@ def _make_multi_logger(cfg: DictConfig) -> BaseLogger:
 
 def get_logger_path(config: DictConfig, logger_type: str) -> str:
     """Helper function to create the experiment path."""
-    return (
-        f"{logger_type}/{config.logger.system_name}/{config.env.env_name}/"
-        + f"{config.env.scenario.task_name}"
-        + f"/envs_{config.arch.num_envs}/seed_{config.system.seed}"
-    )
+    return f"{logger_type}/{config.logger.system_name}"
```
sash-a, line 333: I feel like removing all these things will make it very likely to always have a long json file or will it create a new one each time?

sash-a (changes requested):
Just wording for the error and then some clarification, otherwise I'm happy :smile:

mava/utils/logger.py
RuanJohn: In the `JsonLogger` class init we add a unique token to the json file name so it should always be unique. I just simplified this because it was a relic of old Mava experiments we did when we were still using sacred.

sash-a (approved):
:rugby_football: :

OmaymaMahjoub (approved):
:medal_sports:

PR #998, opened 2024-01-31
Feat/learning rate decay

RuanJohn:
## What?
Add the option to linearly decay the actor and critic learning rates during training. 
## Why?
Based PPO implementation details blog [here](https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/). 
## How?
Simple utils file that creates a linear learning rate decay scheduler and passes that to the optimisers. 
## Extra
Default behaviour (constant learning rates) are retained by setting `decay_learning_rates: False` in the system configs.

### DIFF
--- a/mava/systems/ff_ippo.py
+++ b/mava/systems/ff_ippo.py
@@ -44,13 +44,10 @@ from mava.types import (
 )
 from mava.utils import make_env as environments
 from mava.utils.checkpointing import Checkpointer
-from mava.utils.jax import (
-    make_learning_rate_schedule,
-    merge_leading_dims,
-    unreplicate_learner_state,
-)
+from mava.utils.jax import merge_leading_dims, unreplicate_learner_state
 from mava.utils.logger import LogEvent, MavaLogger
 from mava.utils.total_timestep_checker import check_total_timesteps
+from mava.utils.training import make_learning_rate_schedule
 
 
 def get_learner_fn(
--- a/mava/systems/ff_ippo.py
+++ b/mava/systems/ff_ippo.py
@@ -47,7 +47,7 @@ from mava.utils.checkpointing import Checkpointer
 from mava.utils.jax import merge_leading_dims, unreplicate_learner_state
 from mava.utils.logger import LogEvent, MavaLogger
 from mava.utils.total_timestep_checker import check_total_timesteps
-from mava.utils.training import make_learning_rate_schedule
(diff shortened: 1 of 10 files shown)


WiemKhlifi (dismissed):
LGTM 🛥️

mava/systems/ff_ippo.py
```diff
@@ -355,13 +356,25 @@ def learner_setup(
     actor_network, critic_network = networks.make(
         config=config, network="feedforward", centralised_critic=False
     )
+
+    actor_lr = (
+        make_learning_rate_schedule(config.system.actor_lr, config)
```
OmaymaMahjoub: how about instead of making lr scheduler you get_lr directly: by that you save the if statements to be inside the function

RuanJohn: Super nice idea! What do you think of the changes?

mava/systems/ff_ippo.py
```diff
@@ -357,16 +357,8 @@ def learner_setup(
...
-        make_learning_rate_schedule(config.system.actor_lr, config)
-        if config.system.decay_learning_rates
-        else config.system.actor_lr
-    )
-    critic_lr = (
-        make_learning_rate_schedule(config.system.critic_lr, config)
-        if config.system.decay_learning_rates
-        else config.system.critic_lr
-    )
+    actor_lr = make_learning_rate(config.system.actor_lr, config)
```
OmaymaMahjoub, line 360: isn't sending the config enough to retrieve the actor_lr?

mava/systems/ff_ippo.py
OmaymaMahjoub: Yes! Thanks for the changes :)

mava/systems/ff_ippo.py
RuanJohn: It is difficult to do, since we would then have to pass in an arg to say that we want to make a scheduler for the actor or the critic. So I felt that passing in an initial learning rate keeps things as general as possible.

OmaymaMahjoub: Oh I get it now! Thanks for the explanation :pray:

WiemKhlifi (approved):
Thank you for these changes @RuanJohn 🥳 🙌

OmaymaMahjoub (approved):
Thanks @RuanJohn!

PR #999, opened 2024-02-02
Feat: Continuous actions networks and training utils

WiemKhlifi:
## What?
In this PR, we add continuous networks and training functions to transform the log probability and the actions produced. It also modifies configs and the JaxMARL wrapper accordingly, since mabrax doesn't provide a global state. 

## Why?
These changes add support for environments with continuous actions.

## How?
* Add continuous networks that return actor_mean and actor_log_std.
* Use utils functions to sample action and log_prob from the distribution and transform them.
* Add a global state to mabrax by simply concatenating agent observations.
## Extra
This PR will be followed by a full PR that will add the various continuous systems as well as their configs and evaluator changes.


mava/configs/env/mabrax.yaml
```diff
@@ -0,0 +1,16 @@
...
+# Possible envs: [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
+env_name: ant_4x2
+scenario:
+  task_name: ant_4x2  # In this case: env_name = task_name
+
+kwargs:
+  episode_length: 1000
+  homogenisation_method: max # [None, max, concat] if the dims of observations and actions are not homegenous across agents.
+  # Homogeneous tasks: ant_4x2, walker2d_2x3 -> "homogenisation_method: None"
+  auto_reset: True
```
sash-a: If it makes no difference we shouldn't even expose this option, otherwise we should set it to false, otherwise it might become confusing

```suggestion
```

mava/configs/env/mabrax.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+actions_type: continuous # The type of actions used in this env (Note: use the mlp_continuous.yaml in this case).
```
sash-a, line 15: Cool! :fire:

mava/networks.py
```diff
@@ -217,22 +260,23 @@ def create_torso(network_key: str, layer_size_key: str) -> MLPTorso:
             use_layer_norm=config.network[network_key].use_layer_norm,
         )
 
+    actions_type = config.env.actions_type if hasattr(config.env, "actions_type") else "discrete"
```
sash-a: I think we should force all configs to have the `action_type` variable, so existing ones should get discrete

mava/networks.py
```diff
@@ -217,22 +260,23 @@ def create_torso(network_key: str, layer_size_key: str) -> MLPTorso:
...
     if network == "feedforward":
-        actor = FeedForwardActor(
+        actor = _networks[actions_type]["ff_actor"](
```
sash-a: Unfortunate that these functions take different args, would be nice to just go

```
actor = _networks[actions_type]["ff_actor"](...)
critic = _networks["ff_critic"](...)
```

and get rid of all the `if`s but I'm not sure there's a clean way to do this :cry:

mava/utils/training.py
```diff
@@ -56,3 +59,74 @@ def make_learning_rate(init_lr: float, config: DictConfig) -> Union[float, Calla
         return make_learning_rate_schedule(init_lr, config)
     else:
         return init_lr
+
+
+def select_action_ppo(
```
sash-a:
```suggestion
def select_action_cont_ppo(
```

mava/utils/training.py
```diff
@@ -56,3 +59,74 @@ def make_learning_rate(init_lr: float, config: DictConfig) -> Union[float, Calla
...
+
+
+def transform_actions_log(
+    env_name: str, raw_action: Array, log_prob: Array
+) -> Union[Array, Tuple[Array, Array]]:
+    """Transform action and log_prob values"""
+
+    # IF action in [-N, N] and N != 1 ELSE [-1, 1] and N == 1
+    n = 0.4 if env_name == "humanoid_9|8" else 1
+    action = n * jnp.tanh(raw_action)
```
sash-a: This is very hard to read and doesn't support future envs, we should pass in action scale and bias to this function and pipe it through all the other functions. See how I do it [here](https://github.com/instadeepai/Mava/blob/feat/jitting-jax-sac/mava/systems/ff_isac.py#L203)

mava/utils/training.py
```diff
@@ -56,3 +59,74 @@ def make_learning_rate(init_lr: float, config: DictConfig) -> Union[float, Calla
...
+    """Get the log probability and entropy of a given actor output and traj_batch action."""
+
+    actor_mean, actor_log_std = actor_output
+    actor_policy = distrax.MultivariateNormalDiag(actor_mean, jnp.exp(actor_log_std))
+
+    log_prob = actor_policy.log_prob(traj_action)
+    entropy = actor_policy.entropy().mean()
+
+    _, log_prob = transform_actions_log(env_name, traj_action, log_prob)
+    return log_prob, entropy
```
sash-a: naming and separating 

```suggestion
def get_logprob_entropy(
    mean: Array,
    log_std: Array,
    action: Array,
    env_name: str,
) -> Tuple[Array, Array]:
    """Get the log probability and entropy of a given mean, log_std and action."""

    actor_policy = distrax.MultivariateNormalDiag(mean, log_std)

    log_prob = actor_policy.log_prob(action)
    entropy = actor_policy.entropy().mean()

    _, log_prob = transform_actions_log(env_name, action, log_prob)
    return log_prob, entropy
```

mava/wrappers/jaxmarl.py
```diff
@@ -208,7 +218,7 @@ def step(
             reward=batchify(reward, self.agents),
             discount=1.0 - batchify(done, self.agents),
             observation=obs,
-            extras=infos,
+            # extras=infos,
```
```suggestion
```

mava/utils/training.py
```diff
@@ -56,3 +59,74 @@ def make_learning_rate(init_lr: float, config: DictConfig) -> Union[float, Calla
...
+    """Transform action and log_prob values"""
+
+    # IF action in [-N, N] and N != 1 ELSE [-1, 1] and N == 1
+    n = 0.4 if env_name == "humanoid_9|8" else 1
+    action = n * jnp.tanh(raw_action)
+    if log_prob is None:
+        return action  # during evaluation.
+
+    # Note: jnp.log(derivative of action equation)
+    log_prob -= jnp.sum(jnp.log(n * (1 - jnp.tanh(raw_action) ** 2) + 1e-6), axis=-1)
```
sash-a: This might not be the best formulation of the transformation. I've been experimenting with 

```
jnp.sum(2 * (jnp.log(2) - unbound_action - jax.nn.softplus(-2 * unbound_action)), axis=-1)
```

Which might be a more numerically stable version. Can you test this and see if it works for you.

mava/utils/training.py
```diff
@@ -56,3 +59,74 @@ def make_learning_rate(init_lr: float, config: DictConfig) -> Union[float, Calla
...
+    #     hstate, actor_mean, actor_log_std = actor_output
+    # else:
+    #     actor_mean, actor_log_std = actor_output
+
+    actor_policy = distrax.MultivariateNormalDiag(actor_output[-2], jnp.exp(actor_output[-1]))
+
+    raw_action = actor_policy.sample(seed=key)
+    action = transform_actions_log(env_name, raw_action, None)
+
+    return (actor_output[0], action) if network == "recurrent" else action
```
sash-a: I don't like the if in the return and the `[-1]`/`[-2]` it's quit hard to read, can you rather make a rec and ff version of this fn? Also can you make it clear that this is for continuous systems. eg `def select_action_eval_cont_ff`

mava/networks.py
```diff
@@ -98,6 +99,23 @@ def __call__(self, observation: Union[Observation, ObservationGlobalState]) -> c
         return jnp.squeeze(critic_output, axis=-1)
 
 
+class ContinuousFFActor(nn.Module):
```
sash-a, line 153: Might be nice to have a log_std that's just a network param and see how that works? It's how a lot of [other PPO implementations](https://github.com/FLAIROx/JaxMARL/blob/main/baselines/IPPO/ippo_ff_mabrax.py#L42) do it. Could you test this and add it as a separate type of network if it works well?

sash-a (changes requested):
Thanks for this Wiem! Looking good

WiemKhlifi:
We're closing this PR since we'll adopt new ways to initialise the action's distribution in the continuous case with non-vmapped networks and avoid creating different systems files.

PR #1003, opened 2024-02-06
fix: use correct rware timelimit

callumtilbury:
## What?
Incorrect default time limit for RWARE in #991 – should be 500.

## Why?
NB to have the correct default value.

## How?
–

## Extra
–


RuanJohn (approved):
Thank you @callumtilbury  :100:

OmaymaMahjoub (approved):
LGTM! Thanks @callumtilbury :fire:

PR #1004, opened 2024-02-06
fix: use correct fbx location

callumtilbury:
Point to correct fbx location. Temporary until new PyPI release, unless we always want to point to fbx's repo directly.


WiemKhlifi (approved):
Thanks @callumtilbury for fixing this 🥳 🙏

RuanJohn (approved):

PR #1005, opened 2024-02-06
feat: use pypi's version for fbx

callumtilbury:
## What?
Sorry, another change here: the new fbx location works, but it would be better to point to PyPI's version, which now has been updated.

## Why?
Better to point to the official pip version :)

## How?
Update requirements file.

## Extra
—


RuanJohn (approved):
Thanks @callumtilbury :raised_hands:

WiemKhlifi (approved):
Cool 🚀  Thanks @callumtilbury 🙏

PR #1007, opened 2024-02-07
feat: change how networks are initialised by making use of hydra

EdanToledo:
## What?
Change how networks are created to utilise hydra to instantiate torsos and action heads. 
## Why?
Makes it extremely easy to: a) change the network to be suitable for discrete or continuous environments b) change the type of critic head used i.e. a scalar head or a distributional head as this is all controlled in configs c) allow multi-runs and hyperparameter sweeps to easily tune not only the network configs but also the network types and structures etc. 

All in all it makes it easier and more general to construct and change networks as well as parameterise them efficiently.
## How?
Utilise hydras instantiate abilities. 
## Extra
I have added a CNN torso and config. Although i have not used them as there isn't an environment that has a 3D observation, I know it works/the config is set up correctly as i use it in a mava fork for single agent envs.


mava/networks.py
```diff
@@ -51,19 +52,38 @@ def __call__(self, observation: chex.Array) -> chex.Array:
...
+        return x.reshape((x.shape[0], -1))
+
+
+class DiscreteActionHead(nn.Module):
+    """Discrete Action Head"""
+
+    action_dim: int
+
+    @nn.compact
+    def __call__(self, x: chex.Array, observation: Observation) -> distrax.Categorical:
```
sash-a: Why pass in x and observation here? I think we should always pass in observation

mava/networks.py
```diff
@@ -130,7 +168,7 @@ def initialize_carry(batch_size: int, hidden_size: int) -> chex.Array:
 class RecurrentActor(nn.Module):
     """Recurrent Actor Network."""
 
-    action_dim: Sequence[int]
+    action_head: nn.Module
```
sash-a: Nit: can we keep the ordering the same as ff actor, first torso's then head, should be less confusing in future

sash-a (changes requested):
Really awesome changes, thanks @EdanToledo two super minor comments :smile:

mava/networks.py
EdanToledo: So x is for the hidden layer output of a torso - the only reason observation is passed in is for the action mask and/or if somebody wanted to use additional information from the observation.

mava/networks.py
EdanToledo: agreed

EdanToledo:
@RuanJohn is going to take over this PR. Thanks for the help.

mava/networks.py
sash-a: Could we name it better/add docs to make this clearer as it wasn't clear at first glance

mava/networks.py
RuanJohn: Fixed this :+1:

mava/networks.py
RuanJohn: I replaced `observation` -> `observation.action_mask` to make things clearer. The action head's input name was also updated. What do you think of this @sash-a?

mava/networks.py
```diff
@@ -37,9 +35,12 @@ class MLPTorso(nn.Module):
     """MLP torso."""
 
     layer_sizes: Sequence[int]
-    activation_fn: Callable[[chex.Array], chex.Array] = nn.relu
+    activation: str = "relu"
     use_layer_norm: bool = False
 
+    def setup(self) -> None:
```
sash-a, line 41: Nice

mava/networks.py
```diff
@@ -80,13 +84,13 @@ class DiscreteActionHead(nn.Module):
     action_dim: int
 
     @nn.compact
-    def __call__(self, x: chex.Array, observation: Observation) -> distrax.Categorical:
+    def __call__(self, x: chex.Array, action_mask: chex.Array) -> distrax.Categorical:
```
sash-a: As discussed offline: 

```suggestion
    def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distrax.Categorical:
```

sash-a (changes requested):
Really nice, just one thing that we discussed offline

mava/networks.py
RuanJohn: Just added a bit more detail than your suggestion - but very much along the same lines

mava/configs/network/cnn.yaml
```diff
@@ -0,0 +1,21 @@
+# ---MLP Networks---
```
EdanToledo:
```suggestion
# ---CNN Networks---
```

mava/networks.py
```diff
@@ -107,13 +107,13 @@ class FeedForwardActor(nn.Module):
...
 
-        x = self.torso(x)
+        obs_embedding = self.torso(observation.agents_view)
 
-        x = self.action_head(x, observation.action_mask)
-
-        return x
+        # We pass both the observation embedding and the observation object to the action head
+        # since the observation object contains the action mask and other potentially useful
+        # information.
```
sash-a: Nit: can you add this to the docs of the `__call__` of `DiscreteActionHead`

RuanJohn: Done :call_me_hand:

sash-a (approved):
Lekker :1st_place_medal:

OmaymaMahjoub (approved):
LGTM!

PR #1010, opened 2024-02-08
fix: cast timestep to int

### LINKED ISSUE #1009: [BUG] Crash due to floating point instead of int
### Describe the bug
The timestep variable $t$ is sometimes a float and other times an int. It depends on the configs. 

Super simple fix: 

wherever you see : 

`t = steps_per_rollout * (eval_step + 1)`

make it:

`t = int(steps_per_rollout * (eval_step + 1))`

### Expected behavior
A crash not happening.

callumtilbury:
## What?
Cast timestep to int before logging.

## Why?
Fixes #1009.

## How?
Cast to int.

## Extra
—


callumtilbury:
@EdanToledo can you check that this is ok? Thanks pal!

EdanToledo (approved):

OmaymaMahjoub (approved):
Thanks @callumtilbury :fire:

PR #1013, opened 2024-02-12
feat: independent sac

sash-a:
## What?
Feed Forward Independent Soft Actor Critic implementation

## Extra
Will make a smaller PR for the changes around SAC so that the only thing I'm changing in this PR is adding SAC. Coming soon...


mava/systems/sac/types.py
OmaymaMahjoub, line 1: For consistency, if we will have types for sac we need to separate the ppo types from the default file types.py and move them to ppo folder in systems

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+
+    def replicate(x: Any) -> Any:
+        """First replicate the update batch dim then put on devices."""
+        x = jax.tree_map(lambda y: jnp.broadcast_to(y, (cfg.system.update_batch_size, *y.shape)), x)
+        return jax.device_put_replicated(x, devices)
+
+    env, eval_env = environments.make(cfg)
+
+    n_agents = env.action_spec().shape[0]
+    action_dim = env.action_spec().shape[1]
```
WiemKhlifi: Small NIT: We can now access these from env 🙌 
``` suggestion
    n_agents = env.num_agents
    action_dim = env.action_dim
```

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+    acts = env.action_spec().generate_value()  # all agents actions
+    act_single_batched = acts[0][jnp.newaxis, ...]  # batch single agent action
+    obs = env.observation_spec().generate_value()
+    obs_single_batched = jax.tree_map(lambda x: x[0][jnp.newaxis, ...], obs)
+
+    # Making actor network
+    actor_torso = hydra.utils.instantiate(cfg.network.actor_network.pre_torso)
+    actor_action_head = hydra.utils.instantiate(
+        cfg.network.action_head, action_dim=env.action_dim, independent_std=False
+    )
```
WiemKhlifi:
```suggestion
    actor_action_head = hydra.utils.instantiate(
        cfg.network.action_head, action_dim=action_dim, independent_std=False
    )
```

mava/configs/system/sac/ff_isac.yaml
```diff
@@ -0,0 +1,33 @@
+# --- Defaults FF-ISAC ---
+seed: 1
+n_envs: 16 # number of environments to run in parallel (vmapped)
+
+# --- Agent observations ---
+add_agent_id: False
```
OmaymaMahjoub, line 5: Unrelated to this PR, shall we have this in env files rather than having it here?

mava/systems/sac/types.py
```diff
@@ -0,0 +1,73 @@
...
+BufferState: TypeAlias = TrajectoryBufferState[Transition]
+
+
+class LearnerState(NamedTuple):
+    obs: Array
+    env_state: State
+    buffer_state: BufferState
+    params: SacParams
+    opt_states: OptStates
+    t: Array
```
OmaymaMahjoub: for consistency if we can make it here `timestep` instead of `t`

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+
+    return pmaped_explore, pmaped_updated_step
+
+
+def run_experiment(cfg: DictConfig) -> float:
+    # Add runtime variables to config
+    cfg.arch.n_devices = len(jax.devices())
+
+    # Number of env steps before evaluating/logging.
+    steps_per_rollout = int(cfg.system.total_timesteps // cfg.arch.num_evaluation)
```
OmaymaMahjoub, line 489: If the user specifies the number of updates, setting up timesteps becomes optional. Consequently, to ensure consistency, especially when conducting sweeps, it's advisable to adopt the same approach so that parameters aiming to achieve the same objective will be uniformly defined.
```
    # Calculate total timesteps.
    config = check_total_timesteps(config)
    steps_per_rollout = (
        n_devices
        * config.system.num_updates_per_eval
        * config.system.rollout_length
        * config.system.update_batch_size
        * config.arch.num_envs
    )
```

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+
+    key, actor_key, q1_key, q2_key, q1_target_key, q2_target_key = jax.random.split(key, 6)
+
+    acts = env.action_spec().generate_value()  # all agents actions
+    act_single_batched = acts[0][jnp.newaxis, ...]  # batch single agent action
+    obs = env.observation_spec().generate_value()
+    obs_single_batched = jax.tree_map(lambda x: x[0][jnp.newaxis, ...], obs)
+
+    # Making actor network
+    actor_torso = hydra.utils.instantiate(cfg.network.actor_network.pre_torso)
```
lbeyers, line 109: (Newbie eyes comment) This hydra way of instantiating a network is new to me - how does it know where to look, and does it return the exact same thing as the usual way of instantiating a network?

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+        action = jax.random.uniform(explore_key, full_action_shape)
+        next_obs, env_state, buffer_state, metrics = step(action, obs, env_state, buffer_state)
+
+        t += cfg.system.n_envs
+        learner_state = carry._replace(
+            obs=next_obs, env_state=env_state, buffer_state=buffer_state, t=t, key=key
+        )
+        return learner_state, metrics
+
+    scanned_update = lambda state: lax.scan(update_epoch, state, None, length=cfg.system.epochs)
```
lbeyers: Not a big thing, I initially found it difficult to understand from the name "update_epoch" what the function was doing. I used the word "train" when I adapted the code, so that the scan makes the system "train" for a number of epochs.

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+        rewards = timestep.reward
+        terms = ~timestep.discount.astype(bool)
+        infos = timestep.extras
+
+        real_next_obs = infos["real_next_obs"]
+
+        transition = Transition(obs, action, rewards, terms, real_next_obs)
+        buffer_state = rb.add(buffer_state, transition)
+
+        return next_obs, env_state, buffer_state, infos["episode_metrics"]
```
lbeyers, line 252: I really like how general this is.

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+
+        # Sample and learn
+        learn_state = (buffer_state, params, opt_states, t, learn_key)
+        (buffer_state, params, opt_states, _, _), losses = scanned_update(learn_state)
+
+        t += cfg.system.n_envs * cfg.system.rollout_length
+        return (
+            LearnerState(next_obs, env_state, buffer_state, params, opt_states, t, key),
+            (metrics, losses),
+        )
```
lbeyers, line 458: I like how visible the overall structure here is - the name might suggest, though, that this updates the "step" function defined earlier. The docstrings go a long way in providing clarification.

mava/configs/system/sac/ff_isac.yaml
```diff
@@ -0,0 +1,33 @@
+# --- Defaults FF-ISAC ---
+seed: 1
+n_envs: 16 # number of environments to run in parallel (vmapped)
```
OmaymaMahjoub: This should be defined in the arch file called `num_envs`

mava/configs/system/sac/ff_isac.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+# --- Agent observations ---
+add_agent_id: False
+
+# --- RL hyperparameters ---
+# step related
+total_timesteps: 3e8
```
OmaymaMahjoub: To maintain consistency in the logic of defining experiments with PPO, it would be advantageous to introduce a parameter similar to num_updates as found in PPO.
```
total_timesteps: ~ # Set the total environment steps.
# If unspecified, it's derived from num_updates; otherwise, num_updates adjusts based on this value.
num_updates: 1000 # Number of updates
```

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+import time
+from pprint import pprint
```
OmaymaMahjoub:
```suggestion
from rich.pretty import pprint
```

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+    Tuple[Environment, Environment],
+    Networks,
+    Optimisers,
+    TrajectoryBuffer,
+    LearnerState,
+    Array,
+    MavaLogger,
+    chex.PRNGKey,
+]:
+    logger = MavaLogger(cfg)
```
OmaymaMahjoub, line 86: nit: add a comment to explain that init is to define the net...

OmaymaMahjoub (changes requested):
Thanks @sash-a keen to see isac part of Mava :fire:  The system training fine however it would be great if we could also work on aligning it closely with PPO systems for better consistency.

mava/configs/system/sac/ff_isac.yaml
RuanJohn: Totally agree on this. This gives better control on how many timesteps and evaluations get done. 
I general we should document to the total_timesteps a bit better.

mava/configs/system/sac/ff_isac.yaml
```diff
@@ -0,0 +1,33 @@
...
+policy_frequency: 4  # the frequency of training policy - every n q learning steps the policy is trained once
+
+# sizes
+buffer_size: 500000  # size of the replay buffer. Note: total size is this * num_devices
+batch_size: 32
+
+# learning rates
+policy_lr: 3e-4  # the learning rate of the policy network optimizer
+q_lr: 1e-3  # the learning rate of the Q network network optimizer
+alpha_lr: 13e-4  # the learning rate of the alpha optimizer
```
RuanJohn: Is this 13 intended here?

mava/configs/system/sac/ff_isac.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+# --- RL hyperparameters ---
+# step related
+total_timesteps: 3e8
+explore_steps: 5000  # number of steps to take with random actions at the start of training
+update_batch_size: 1 # number of vectorised gradient updates per device.
+
+rollout_length: 1 # number of environment steps per vectorised environment.
+epochs: 1 # number of learn epochs per training data batch.
+policy_frequency: 4  # the frequency of training policy - every n q learning steps the policy is trained once
```
RuanJohn: I don't quite follow this?

mava/networks.py
```diff
@@ -125,10 +125,15 @@ class ContinuousActionHead(nn.Module):
 
     action_dim: int
     min_scale: float = 1e-3
+    independent_std: bool = True  # whether or not the log_std is independent of the observation.
```
RuanJohn, line 128: Just putting the question here. Does this get set to false by default for SAC?

mava/networks.py
```diff
@@ -166,8 +177,8 @@ def __call__(self, observation: Observation) -> tfd.Distribution:
         return self.action_head(obs_embedding, observation)
 
 
-class FeedForwardCritic(nn.Module):
-    """Feedforward Critic Network."""
+class FeedForwardValueNet(nn.Module):
```
RuanJohn, line 177: I like this generalisation!

mava/systems/sac/types.py
RuanJohn: This is already done for the PPO systems.

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+from mava.networks import FeedForwardActor as Actor
+from mava.networks import FeedForwardQNet as QNetwork
+from mava.systems.sac.types import (
+    BufferState,
+    LearnerState,
+    Metrics,
+    Networks,
+    Optimisers,
+    OptStates,
+    Qs,
```
RuanJohn: Or something along these lines? `Qs` just seems a bit too limited. 
```suggestion
    QVals,
```

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,565 @@
...
+    QsAndTarget,
```
RuanJohn:
```suggestion
    QValsAndTarget,
```

mava/configs/system/sac/ff_isac.yaml
sash-a: Ye maybe it should be in the env file, I'm honestly not sure. I feel like some envs are better with agent_ids and some without, but also the same systems, I think some might do better with/without agent IDs

mava/systems/sac/types.py
sash-a: This is actually t as in the current timestep counter. I'll put a comment to make that more clear :+1:

mava/systems/sac/ff_isac.py
sash-a: Check in the config, there's a `_target_` key that allows hydra to know where to look to instantiate each config item.

mava/systems/sac/ff_isac.py
sash-a: @OmaymaMahjoub and @RuanJohn what do we think about changing the name to `train_epoch` I kept update in most of these function names to stay in line with PPO

mava/systems/sac/ff_isac.py
sash-a: Good point, not sure what else to call it though, it's like the main algorithm step :thinking:

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,598 @@
...
+    rb = fbx.make_item_buffer(
+        max_length=cfg.system.buffer_size,
+        min_length=cfg.system.explore_steps,
+        sample_batch_size=cfg.system.batch_size,
+        add_batches=True,
+    )
+    buffer_state = replicate(rb.init(init_transition))
+
+    nns = (actor, q)
+    opts = (actor_opt, q_opt, alpha_opt)
```
RuanJohn: Can we please be a bit more descriptive with these variable names?

mava/systems/sac/ff_isac.py
RuanJohn: Maybe we can add it to the docs somewhere? Can you make a comment or a commit on the [docs PR](https://github.com/instadeepai/Mava/pull/1052) about this please @lbeyers?

mava/systems/sac/ff_isac.py
sash-a: Just a comment on this I found while checking things, since we have a much lower `rollout_length` in SAC and update more frequently because it's an off policy algorithm, we have to `num_updates` *much* higher, to ~600_000

mava/configs/system/sac/ff_isac.yaml
sash-a: @liamclarkza maybe something to specifically add to the docs, maybe in the sharp bits section is that you can set either of these properties

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,598 @@
...
+    first_keys = first_keys.reshape((cfg.arch.n_devices, cfg.system.update_batch_size, -1))
+
+    env_state, first_timestep = jax.pmap(  # devices
+        jax.vmap(  # update_batch_size
+            jax.vmap(env.reset),  # num_envs
+            axis_name="batch",
+        ),
+        axis_name="device",
+    )(reset_keys)
+    first_obs = first_timestep.observation
```
RuanJohn, line 195: Quite a NIT question: 

But in the PPO systems we do this: 
```
    # Initialise environment states and timesteps: across devices and batches.
    key, *env_keys = jax.random.split(
        key, n_devices * config.system.update_batch_size * config.arch.num_envs + 1
    )
    env_states, timesteps = jax.vmap(env.reset, in_axes=(0))(
        jnp.stack(env_keys),
    )
    reshape_states = lambda x: x.reshape(
        (n_devices, config.system.update_batch_size, config.arch.num_envs) + x.shape[1:]
    )
    # (devices, update batch size, num_envs, ...)
    env_states = jax.tree_map(reshape_states, env_states)
    timesteps = jax.tree_map(reshape_states, timesteps)
```

Is there a reason you are manually `pmap`ing here? Since the train function gets `pmap`ed later on.

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,598 @@
...
+        rb: The replay buffer.
+        target_entropy: The target entropy.
+
+    Returns:
+        Tuple of (explore_fn, update_fn).
+        Explore function is used for initial exploration with random actions.
+        Update function is the main learning function, it both acts and learns.
+    """
+    actor, q = nns
+    actor_opt, q_opt, alpha_opt = opts
```
RuanJohn: Some more descriptive variable names please :pray:

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,598 @@
...
+    explore_steps = cfg.system.explore_steps // cfg.arch.num_envs
+    pmaped_explore = jax.pmap(
+        jax.vmap(
+            lambda state: lax.scan(explore, state, None, length=explore_steps),
+            axis_name="batch",
+        ),
+        axis_name="device",
+        donate_argnums=0,
+    )
+    pmaped_updated_step = jax.pmap(
```
RuanJohn:
```suggestion
    pmaped_update_step = jax.pmap(
```

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,598 @@
...
+        jax.vmap(
+            lambda state: lax.scan(update_step, state, None, length=cfg.system.scan_steps),
+            axis_name="batch",
+        ),
+        axis_name="device",
+        donate_argnums=0,
+    )
+
+    return pmaped_explore, pmaped_updated_step
```
RuanJohn:
```suggestion
    return pmaped_explore, pmaped_update_step
```

RuanJohn (commented):
Really super work here @sash-a!

A first pass through and just a few minor suggestions and comments from me. Will do another pass a bit later today.

mava/configs/system/sac/ff_isac.yaml
sash-a: no, my bad

mava/networks.py
sash-a: Ye it does :+1:

mava/configs/system/sac/ff_isac.yaml
sash-a: It's a DDPG thing, but useful in SAC also, you delay policy updates. So you'll do 4 Q updates and then you update the policy 4 times

mava/systems/sac/ff_isac.py
sash-a: Ye I pmap here because then all the timespteps are on the correct devices at the start of training, but you could also do it without pmapping, I think it makes it a bit more clear honestly

mava/configs/system/sac/ff_isac.yaml
RuanJohn: Ah makes sense! Can we then call it something along these lines please? :pray: 
```suggestion
policy_update_frequency: 4  # the frequency of training policy - every `n` q network learning steps the policy network is updated once
```

mava/systems/sac/ff_isac.py
RuanJohn: Okay, thanks for clarifying - let's keep it like this in that case then.

mava/systems/sac/types.py
```diff
@@ -0,0 +1,73 @@
...
+from mava.types import Observation, ObservationGlobalState, State
+
+Metrics: TypeAlias = Dict[str, Array]
+Networks: TypeAlias = Tuple[nn.Module, nn.Module]
+Optimisers: TypeAlias = Tuple[
+    optax.GradientTransformation, optax.GradientTransformation, optax.GradientTransformation
+]
+
+
+class Qs(NamedTuple):
```
RuanJohn: These would need to update as well given then changes to naming in the system file.

mava/systems/sac/types.py
```diff
@@ -0,0 +1,73 @@
...
+    q1: FrozenVariableDict
+    q2: FrozenVariableDict
+
+
+class QsAndTarget(NamedTuple):
```
RuanJohn: Similar here and elsewhere as well.

mava/configs/system/sac/ff_isac.yaml
sash-a: Did it slightly differently, but similar

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,600 @@
...
+    actor_torso = hydra.utils.instantiate(cfg.network.actor_network.pre_torso)
+    actor_action_head = hydra.utils.instantiate(
+        cfg.network.action_head, action_dim=env.action_dim, independent_std=False
+    )
+    actor = Actor(actor_torso, actor_action_head)
+    actor_params = actor.init(actor_key, obs_single_batched)
+
+    # Making Q networks
+    critic_torso = hydra.utils.instantiate(cfg.network.critic_network.pre_torso)
+    q = QNetwork(critic_torso)
```
RuanJohn: I just goes a bit more with the standard in the PPO files where we have `actor_network` and `critic_network` instead of just `actor` and `critic`. 
```suggestion
    q_network = QNetwork(critic_torso)
```

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,600 @@
...
+    act_single_batched = acts[0][jnp.newaxis, ...]  # batch single agent action
+    obs = env.observation_spec().generate_value()
+    obs_single_batched = jax.tree_map(lambda x: x[0][jnp.newaxis, ...], obs)
+
+    # Making actor network
+    actor_torso = hydra.utils.instantiate(cfg.network.actor_network.pre_torso)
+    actor_action_head = hydra.utils.instantiate(
+        cfg.network.action_head, action_dim=env.action_dim, independent_std=False
+    )
+    actor = Actor(actor_torso, actor_action_head)
```
RuanJohn: Similarly,
```suggestion
    actor_network = Actor(actor_torso, actor_action_head)
```

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,600 @@
...
+        return loss, loss_info
+
+    def actor_loss_fn(
+        actor_params: FrozenVariableDict, obs: Array, alpha: Array, q_params: Qs, key: chex.PRNGKey
+    ) -> Array:
+        pi = actor.apply(actor_params, obs)
+        action = pi.sample(seed=key)
+        log_prob = pi.log_prob(action)
+
+        qf1_pi = q.apply(q_params.q1, obs, action)
```
RuanJohn: This is not quite intuitive to read? What is `qf1_pi`?

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,600 @@
...
+
+        loss_info = {"actor_loss": actor_loss, "alpha_loss": alpha_loss}
+        return params, opt_states, loss_info
+
+    # Act/learn loops:
+    def update_epoch(
+        carry: Tuple[BufferState, SacParams, OptStates, int, chex.PRNGKey], _: Any
+    ) -> Tuple[Tuple[BufferState, SacParams, OptStates, int, chex.PRNGKey], Metrics]:
+        """Update the Q function and optionally policy/alpha with TD3 delayed update."""
+        buffer_state, params, opt_states, t, key = carry
```
RuanJohn, line 383: A general remark, but can we say `timestep` instead of `t` please?

mava/systems/sac/ff_isac.py
RuanJohn: Perhaps only `train` is fine? I think avoiding the word epoch here could be beneficial to avoid confusion with PPO epochs.

mava/systems/sac/ff_isac.py
RuanJohn: I think `update_step` is okay in this case but open to hearing other people's thoughts? @OmaymaMahjoub, @lbeyers ?

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,600 @@
...
+    # Add runtime variables to config
+    cfg.arch.n_devices = len(jax.devices())
+    cfg = check_total_timesteps(cfg)
+
+    # Number of env steps before evaluating/logging.
+    steps_per_rollout = int(cfg.system.total_timesteps // cfg.arch.num_evaluation)
+    # Multiplier for a single env/learn step in an anakin system
+    anakin_steps = cfg.arch.n_devices * cfg.system.update_batch_size
+    # Number of env steps in one anakin style update.
+    anakin_act_steps = anakin_steps * cfg.arch.num_envs * cfg.system.rollout_length
```
RuanJohn, line 493: If we can remove the word anakin here and name slightly differently please?

RuanJohn (commented):
Just a few other minor comments. Thanks @sash-a :fire:

mava/systems/sac/ff_isac.py
sash-a: Had a similar chat with @OmaymaMahjoub, if we do timestep here it could be confusing as one might think it's the object, I think `t` is more clear that it's a counter.

mava/systems/sac/ff_isac.py
sash-a: Ah I quite liked this, what else would you call it. I went with anakin because it seems to encapsulate that notion of _steps over distributed devices_, maybe `distributed_steps`? Open to other names if you have any :smile:

mava/systems/sac/ff_isac.py
sash-a: ah this is a relic from cleanrl I think. It's Q-function 1's policy, I'll update :smile:

sash-a: Ah this is a relic from clean RL, it's the Q value from Q function 1 from the action produced by policy pi, I've updated it :smile:

mava/systems/sac/ff_isac.py
sash-a: I'm happy with `update_step`, we could call it like `main_step` or something but I think `update_step` works ok here

mava/configs/logger/ff_isac.yaml
```diff
@@ -0,0 +1,9 @@
+defaults:
+    - base_logger
+
+system_name: ff_isac
+
+# --- Neptune options ---
+kwargs:
+  neptune_project: Instadeep/Mava
+  neptune_tag: [ff-isac, mabrax]
```
RuanJohn: Nit, can we remove all the Neptune overrides from here please. The logger tags etc should all be getting set in `base_logger.yaml` for now similarly to other systems.

mava/configs/system/sac/ff_isac.yaml
RuanJohn: I think we should still keep it at the system level for now since it is a system feature which we require since we are doing parameter sharing.

mava/networks.py
RuanJohn, line 177: Nit: 
One more detail here, can we please rename `RecurrentCritic` -> `RecurrentValueNet` as well for consistency?

mava/systems/sac/ff_isac.py
RuanJohn: Makes total sense.

mava/systems/sac/ff_isac.py
```diff
@@ -0,0 +1,607 @@
...
+    ) -> Tuple[Tuple[BufferState, SacParams, OptStates, int, chex.PRNGKey], Metrics]:
+        """Update the Q function and optionally policy/alpha with TD3 delayed update."""
+        buffer_state, params, opt_states, t, key = carry
+        key, buff_key, q_key, actor_key = jax.random.split(key, 4)
+
+        # sample
+        data = rb.sample(buffer_state, buff_key).experience
+
+        # learn
+        # todo: do this in a loop and get rid of the cond
```
RuanJohn: Discussed offline, but please remove the `todo` comment here and make an issue instead please.

mava/systems/sac/ff_isac.py
RuanJohn: Ja, I don't have a better name off the cuff. Let's keep it as is for now. We can probably think about a better name when we have Sebulba versions of everything.

RuanJohn (commented):
Thank you @sash-a! Vert close! Just some very small final comments from me.

mava/systems/sac/ff_isac.py
lbeyers: On it!

mava/systems/sac/ff_isac.py
lbeyers: Yeah, it's not a big thing!

mava/configs/system/sac/ff_isac.yaml
```diff
@@ -0,0 +1,35 @@
...
+seed: 1
+
+# --- Agent observations ---
+add_agent_id: False
+
+# --- RL hyperparameters ---
+# step related
+total_timesteps: 2e7 # Set the total environment steps.
+# If unspecified, it's derived from num_updates; otherwise, num_updates adjusts based on this value.
+num_updates: 625000 # Number of updates
```
WiemKhlifi: If we want to use a **large** `num_updates`, shouldn't we ignore the `total_timesteps` setting? Or just set one of them 🤔

mava/networks.py
```diff
@@ -142,7 +147,13 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
             tfd.Independent: Independent transformed distribution.
         """
         loc = self.mean(obs_embedding)
-        scale = jax.nn.softplus(self.log_std) + self.min_scale
+
+        if self.independent_std:
+            scale = jax.nn.softplus(self.log_std) + self.min_scale
+        else:
+            scale = self.log_std(obs_embedding)
+            scale = jax.nn.softplus(scale) + self.min_scale
```
WiemKhlifi: NIT:
``` suggestion
        if self.independent_std:
            scale = self.log_std
        else:
            scale = self.log_std(obs_embedding)
      scale = jax.nn.softplus(scale) + self.min_scale
``` 

Shorter but not sure if readable:
``` python
     scale = self.log_std if self.independent_std else self.log_std(obs_embedding)
     scale = jax.nn.softplus(scale) + self.min_scale

```

mava/networks.py
```diff
@@ -166,8 +177,8 @@ def __call__(self, observation: Observation) -> tfd.Distribution:
         return self.action_head(obs_embedding, observation)
 
 
-class FeedForwardCritic(nn.Module):
-    """Feedforward Critic Network."""
+class FeedForwardValueNet(nn.Module):
+    """Feedforward Value Network. Returns the value of an observation."""
```
WiemKhlifi, line 178: NIT:
``` suggestion
    """Feedforward Value Network. Returns a value estimate of the given observation."""
``

File: mava/systems/sac/ff_isac.py (hunk 5 above)
Comment (OmaymaMahjoub) [reply mava/systems/sac/ff_isac.py:489]: Thanks for clarity and the changes @sash-a

File: mava/networks.py (hunk 35 above)
Comment (sash-a) [reply mava/networks.py]: ^ like this!

File: mava/configs/system/sac/ff_isac.yaml [hunk 37]
```diff
@@ -0,0 +1,35 @@
...
+total_timesteps: ~ # Set the total environment steps.
+# If unspecified, it's derived from num_updates; otherwise, num_updates adjusts based on this value.
+num_updates: 625000 # Number of updates
+explore_steps: 5000  # number of steps to take with random actions at the start of training
+update_batch_size: 1 # number of vectorised gradient updates per device.
+
+rollout_length: 2 # number of environment steps per vectorised environment.
+epochs: 32 # number of learn epochs per training data batch.
+policy_update_frequency: 4  # the frequency of training policy -
+# every `policy_frequency` q network learning steps the policy network is trained `policy_frequency` times
```
RuanJohn: Just a small fix needed for this comment please.

OmaymaMahjoub (approved):
Thanks @sash-a :fire:

RuanJohn (approved):
:ship: :fire:

PR #1015, opened 2024-02-12
chore: minor evaluator clean up

sash-a:
## What?
Minor clean up to the evaluator. 

Removed the creation of trained params and eval keys from the eval setup method as it's not really the responsibility of that method, it should just make the evaluation methods.

Also created the `unreplicate_batch_dim` method to easily unreplicate our `update_batch_size` dim.

Finally removed the unecessary vmap inside the evaluator as flax will automatically vmap for us

### DIFF
--- a/mava/advanced_usage/ff_ippo_store_experience.py
+++ b/mava/advanced_usage/ff_ippo_store_experience.py
@@ -472,7 +472,6 @@ def run_experiment(_config: DictConfig) -> None:  # noqa: CCR001
     config = copy.deepcopy(_config)
     logger = MavaLogger(config)
 
-    # Calculate total timesteps.
     n_devices = len(jax.devices())
 
     # Create the enviroments for train and eval.
--- a/mava/advanced_usage/ff_ippo_store_experience.py
+++ b/mava/advanced_usage/ff_ippo_store_experience.py
@@ -31,7 +31,7 @@ from optax._src.base import OptState
 from rich.pretty import pprint
 
 from mava import networks
-from mava.evaluator import evaluator_setup
+from mava.evaluator import make_eval_fns
 from mava.networks import FeedForwardActor as Actor
 from mava.types import (
     ActorApply,
@@ -488,7 +488,7 @@ def run_experiment(_config: DictConfig) -> None:  # noqa: CCR001
     trained_params = unreplicate_batch_dim(learner_state.params.actor_params)
     eval_keys = jax.random.split(key, n_devices)
     eval_keys = eval_keys.reshape(n_devices, -1)
-    evaluator, absolute_metric_evaluator = evaluator_setup(eval_env, actor_network, config)
+    evaluator, absolute_metric_evaluator = make_eval_fns(eval_env, actor_network, config)
 
     config.system.num_updates_per_eval = config.system.num_updates // config.arch.num_evaluation
     steps_per_rollout = (
--- a/mava/advanced_usage/ff_ippo_store_experience.py
+++ b/mava/advanced_usage/ff_ippo_store_experience.py
@@ -485,9 +485,7 @@ def run_experiment(_config: DictConfig) -> None:  # noqa: CCR001
     learn, actor_network, learner_state = learner_setup(env, (key, key_p), config)
 
     # Setup evaluator.
-    trained_params = unreplicate_batch_dim(learner_state.params.actor_params)
-    eval_keys = jax.random.split(key, n_devices)
-    eval_keys = eval_keys.reshape(n_devices, -1)
+    eval_keys = jax.random.split(key_e, n_devices)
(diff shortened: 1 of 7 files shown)


mava/systems/ff_ippo.py
```diff
@@ -464,16 +471,12 @@ def run_experiment(_config: DictConfig) -> None:
...
     # Setup evaluator.
-    evaluator, absolute_metric_evaluator, (trained_params, eval_keys) = evaluator_setup(
-        eval_env=eval_env,
-        key_e=key_e,
-        network=actor_network,
-        params=learner_state.params.actor_params,
-        config=config,
-    )
+    # Broadcast trained params to cores and split keys for each core.
+    trained_params = unreplicate_batch_dim(learner_state.params.actor_params)
```
OmaymaMahjoub: I don't think it is necessary to have trained params initialized before the for loop as we will get the first needed trained_params in line 533

mava/utils/jax.py
```diff
@@ -59,3 +59,9 @@ def unreplicate_learner_state(
     and one for the `update batch size`.
     """
     return jax.tree_map(lambda x: x[(0,) * unreplicate_depth], learner_state)  # type: ignore
+
+
+def unreplicate_batch_dim(x: Any) -> Any:
+    """Unreplicated just the batch dim.
```
OmaymaMahjoub: maybe specify that it is the update_batch_dim and getting the element 0 is fine as the params are replicated (broadcasted) over that batch

mava/systems/ff_ippo.py
OmaymaMahjoub: The same for all the other systems

OmaymaMahjoub (changes requested):
Thanks @sash-a  for cleaning the evaluator :fire:  Just some minor comments!

mava/advanced_usage/ff_ippo_store_experience.py
```diff
@@ -468,6 +472,9 @@ def run_experiment(_config: DictConfig) -> None:  # noqa: CCR001
     config = copy.deepcopy(_config)
     logger = MavaLogger(config)
 
+    # Calculate total timesteps.
```
RuanJohn: We are only getting the number of devices here, is this intended? Same for all systems.

mava/evaluator.py
```diff
@@ -311,25 +307,12 @@ def evaluator_setup(
             10,
         )
     else:
-        vmapped_eval_apply_fn = jax.vmap(
-            network.apply,
-            in_axes=(None, 0),
-        )
-        evaluator = get_ff_evaluator_fn(eval_env, vmapped_eval_apply_fn, config, log_win_rate)
+        evaluator = get_ff_evaluator_fn(eval_env, network.apply, config, log_win_rate)
```
RuanJohn, line 310: I really like moving to network.apply and not having to vmap here. I just want us to be super sure that the networks are behaving as expected, ie. broadcasting over the agents. 
Did you run any performance tests on this @sash-a?

RuanJohn (commented):
Overall looks very nice @sash-a! :clap: 
Just some small questions.

mava/evaluator.py
OmaymaMahjoub: Nice point @RuanJohn , It may behave the same (@sash-a if you can test it and make sure this is true) and if this is the case something really important to check is steps per seconds

mava/systems/ff_ippo.py
```diff
@@ -450,12 +454,15 @@ def run_experiment(_config: DictConfig) -> None:
     """Runs experiment."""
     config = copy.deepcopy(_config)
 
+    # Calculate total timesteps.
```
OmaymaMahjoub: the comment is wrong here this is getting the number of devices and this comment should be before line 477 (same for other files)

mava/advanced_usage/ff_ippo_store_experience.py
OmaymaMahjoub: yes because we need it before we initialize the evaluator as it is necessary for `eval_keys = jax.random.split(key_e, n_devices)`

sash-a: Ah my bad, copied the comment as well as the `n_devices` when I moved it

mava/evaluator.py
sash-a: I remember testing this ages ago and it was the same, also I do this in SAC and it works fine. Although I'm actually a bit worried about it behaving incorrectly with CNNs 🤔

RuanJohn: I think a few simple ppo tests would be great for this if possible just to be sure please :pray:

sash-a: Testing...

OmaymaMahjoub (approved):
LGTM!

RuanJohn (approved):
:ship:

PR #1016, opened 2024-02-12
fix: agent id obs spec dtype

sash-a:
## What?
Agent ID wrapper assumed that the obs dtype was int (discrete) this is no longer the case as we support continuous envs.

1 liner to fix this

### DIFF
--- a/mava/wrappers/observation.py
+++ b/mava/wrappers/observation.py
@@ -86,15 +86,15 @@ class AgentIDWrapper(Wrapper):
         agents_view = specs.Array((self._env.num_agents, num_obs_features), dtype, "agents_view")
 
         if self.has_global_state:
-            state_spec = obs_spec.global_state
-            state_shape = state_spec.shape
-            state_dtype = state_spec.dtype
+            wrapped_state_spec = obs_spec.global_state
+            wrapped_state_shape = wrapped_state_spec.shape
+            wrapped_state_dtype = wrapped_state_spec.dtype
 
-            state_shape = (
-                *state_shape[:-1],
-                state_shape[-1] + self._env.num_agents,
+            wrapped_state_shape = (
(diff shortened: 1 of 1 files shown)


RuanJohn (dismissed):
Good catch! Thanks @sash-a :fire:

OmaymaMahjoub (approved):
nice catch :eyes:

OmaymaMahjoub (changes requested):
Can you do the same for `global_state = specs.Array(state_shape, jnp.int32, "global_state")` (line 95)

sash-a:
> Can you do the same for `global_state = specs.Array(state_shape, jnp.int32, "global_state")` (line 95)

Great catch @OmaymaMahjoub, fixed that one and another one in the `GlobalStateWrapper`

OmaymaMahjoub (dismissed):
:fire:

mava/wrappers/observation.py
```diff
@@ -81,18 +81,17 @@ def observation_spec(
...
+        agents_view = specs.Array((self._env.num_agents, num_obs_features), dtype, "agents_view")
 
         if self.has_global_state:
             wrapped_state_shape = obs_spec.global_state.shape
             state_shape = (
                 *wrapped_state_shape[:-1],
                 wrapped_state_shape[-1] + self._env.num_agents,
             )
-            global_state = specs.Array(state_shape, jnp.int32, "global_state")
+            global_state = specs.Array(state_shape, dtype, "global_state")
```
RuanJohn: Quite a nit question, but can't there be weird cases where a global state has a different dtype than the agent's view?

RuanJohn (approved):
LGTM. Thanks @sash-a!

OmaymaMahjoub (approved):
:fire:

PR #1018, opened 2024-02-13
feat: add dones to episode metrics

sash-a:
## What?
Improves the `RecordEpisodeMetrics` wrapper by adding dones, which allows us to pull out the final reward from the episode. Thus instead of logging the mean of many final episode returns we now log the mean of just the last episode return.

## Why?
More accurate representation of system performance.

### DIFF
--- a/mava/wrappers/episode_metrics.py
+++ b/mava/wrappers/episode_metrics.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING, Tuple
+from typing import TYPE_CHECKING, Dict, Tuple
 
+import jax
 import chex
 import jax.numpy as jnp
 from jumanji.types import TimeStep
@@ -32,6 +33,7 @@ class RecordEpisodeMetricsState:
     """State of the `LogWrapper`."""
 
     env_state: State
+    key: chex.PRNGKey
     # Temporary variables to keep track of the episode return and length.
     running_count_episode_return: chex.Numeric
     running_count_episode_length: chex.Numeric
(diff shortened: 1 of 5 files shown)


mava/wrappers/episode_metrics.py
```diff
@@ -72,13 +79,31 @@ def step(
...
             episode_length=episode_length_info,
         )
         return state, timestep
+
+def get_final_step_metrics(metrics: Dict[str, chex.Array]) -> Dict[str, chex.Array]:
+    """Get the metrics for the final step of an episode.
+
+    Note: this is not a jittable method.
+    """
+    is_final_ep = metrics.pop("final_step")
```
liamclarkza: Perhaps rename "final_step" to "is_terminal_step" or something similar for clarity?

mava/wrappers/episode_metrics.py
```diff
@@ -72,13 +79,31 @@ def step(
...
             running_count_episode_length=new_episode_length * not_done,
             episode_return=episode_return_info,
             episode_length=episode_length_info,
         )
         return state, timestep
+
+def get_final_step_metrics(metrics: Dict[str, chex.Array]) -> Dict[str, chex.Array]:
+    """Get the metrics for the final step of an episode.
+
+    Note: this is not a jittable method.
```
RuanJohn: Please explain why this method is returning variable length arrays, since the logger requires an array for computing aggregated scores.

sash-a: Updated, resolve if you're happy with the description :smile:

liamclarkza (dismissed):
All looking good.

mava/wrappers/episode_metrics.py
```diff
@@ -72,13 +79,33 @@ def step(
...
         )
         return state, timestep
+
+
+def get_final_step_metrics(metrics: Dict[str, chex.Array]) -> Dict[str, chex.Array]:
+    """Get the metrics for the final step of an episode.
+
+    Note: this is not a jittable method. We need to return variable length arrays, since
+    we don't know how many episodes have been run.
+    """
```
RuanJohn:
```suggestion
    """Get the metrics for the final step of an episode.

    Note: this is not a jittable method. We need to return variable length arrays, since
    we don't know how many episodes have been run. This is done since the logger
    expects arrays for computing summary statistics on the episode metrics.
    """
```

RuanJohn (approved):
LGTM

liamclarkza (approved):
👍

PR #1000, opened 2024-02-02
Feat: Full support for continuous actions case

WiemKhlifi:
## What?
This PR adds full support for continuous actions in all Mava PPO systems. 

## Why?
Add support for environments with continuous actions.

## How?
* Add continuous PPO systems using the networks and utils functions introduced in PR #999.
* Add all systems and networks config.
* Modify the evaluator to include transformations made on selected actions by the actor_policy.

## Extra
This PR is the continuation of this #999 so it should be reviewed first 🙌

PR #1019, opened 2024-02-13
feat: Edit JaxMARL wrapper 

WiemKhlifi:
## What?
Edit the JaxMARL wrapper to support the mabrax environment. 
## Why?
These changes serve to support mabrax and use the brax's observation as the global state for the mappo systems. These changes are useful for the continuous actions PR #1000.

### DIFF
--- a/mava/configs/env/lbf.yaml
+++ b/mava/configs/env/lbf.yaml
@@ -4,7 +4,8 @@ defaults:
   - scenario: 2s-8x8-2p-2f-coop # [2s-8x8-2p-2f-coop, 8x8-2p-2f-coop, 2s-10x10-3p-3f, 10x10-3p-3f, 15x15-4p-3f, 15x15-4p-5f]
 # Further environment config details in "2s-8x8-2p-2f-coop" file.
 
-env_name: LevelBasedForaging-v0
+env_name: LevelBasedForaging
+
 # Choose whether to aggregate the list of individual rewards and use the team reward (default setting) OR use_individual_rewards=True.
 use_individual_rewards: False  # If True, use the list of individual rewards.
 
--- a/mava/configs/env/lbf.yaml
+++ b/mava/configs/env/lbf.yaml
@@ -4,7 +4,7 @@ defaults:
   - scenario: 2s-8x8-2p-2f-coop # [2s-8x8-2p-2f-coop, 8x8-2p-2f-coop, 2s-10x10-3p-3f, 10x10-3p-3f, 15x15-4p-3f, 15x15-4p-5f]
 # Further environment config details in "2s-8x8-2p-2f-coop" file.
 
-env_name: LevelBasedForaging
+env_name: LevelBasedForaging # Used for logging purposes.
 
 # Choose whether to aggregate the list of individual rewards and use the team reward (default setting) OR use_individual_rewards=True.
 use_individual_rewards: False  # If True, use the list of individual rewards.
--- a/mava/configs/env/mabrax.yaml
+++ b/mava/configs/env/mabrax.yaml
@@ -2,7 +2,7 @@
 defaults:
   - _self_
 
-env_name: MaBrax
+env_name: MaBrax  # Used for logging purposes and selection of the corresponding wrapper.
 scenario:
   name: ant_4x2   # [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
 
--- a/mava/configs/env/mabrax.yaml
+++ b/mava/configs/env/mabrax.yaml
@@ -2,7 +2,7 @@
 defaults:
   - _self_
 
-env_name: Mabrax
+env_name: MaBrax
 scenario:
   name: ant_4x2   # [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
 
--- a/mava/configs/env/mabrax.yaml
+++ b/mava/configs/env/mabrax.yaml
@@ -3,8 +3,9 @@ defaults:
   - _self_
 
 # Possible envs: [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
-env_name: ant_4x2
+env_name: MABRAX
 scenario:
+  name: ant_4x2
   task_name: ant_4x2  # In this case: env_name = task_name
 
 kwargs:
@@ -12,5 +13,3 @@ kwargs:
   homogenisation_method: max #Default is None. [max, concat] if the dims of observations and actions are not homegenous across agents.
   # Note: "max" adds one-hot for the agent's ID so the system.add_agent_id should be False. In this case, the agent's ID will be
   # added to the global_state within the wrapper.
-
-env_wrapper: MabraxWrapper
--- a/mava/configs/env/mabrax.yaml
+++ b/mava/configs/env/mabrax.yaml
@@ -1,0 +1,16 @@
+# ---Environment Configs---
+defaults:
+  - _self_
+
+# Possible envs: [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
+env_name: ant_4x2
+scenario:
+  task_name: ant_4x2  # In this case: env_name = task_name
+
+kwargs:
+  episode_length: 1000
+  homogenisation_method: max #Default is None. [max, concat] if the dims of observations and actions are not homegenous across agents.
+  # Note: "max" adds one-hot for the agent's ID so the system.add_agent_id should be False. In this case, the agent's ID will be
+  # added to the global_state within the wrapper.
+
+env_wrapper: MabraxWrapper
--- a/mava/configs/env/mabrax.yaml
+++ b/mava/configs/env/mabrax.yaml
@@ -1,15 +1,15 @@
-# ---Environment Configs---
+# --- Environment Configs---
 defaults:
   - _self_
 
-# Possible envs: [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
-env_name: MABRAX
+env_name: Mabrax
 scenario:
-  name: ant_4x2
-  task_name: ant_4x2  # In this case: env_name = task_name
+  name: ant_4x2   # [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
 
 kwargs:
   episode_length: 1000
-  homogenisation_method: max #Default is None. [max, concat] if the dims of observations and actions are not homegenous across agents.
-  # Note: "max" adds one-hot for the agent's ID so the system.add_agent_id should be False. In this case, the agent's ID will be
-  # added to the global_state within the wrapper.
+  homogenisation_method: max # Default is None. Options: [max, concat] if dimensions of observations and actions are not homogeneous across agents,
+  # "max" adds one-hot for the agent's ID. In this case, system.add_agent_id should be set to False.
+  # Note: In this case, the agent's ID will be added to the global state within the wrapper if `add_agent_ids_to_state` is True.
+
+add_agent_ids_to_state: True # Adds the agent IDs to the global state.
--- a/mava/configs/env/smax.yaml
+++ b/mava/configs/env/smax.yaml
@@ -2,8 +2,8 @@
 defaults:
   - _self_
 
-env_name: SMAX
-# Possible scenarios:
+env_name: Smax
+# Possible tasks:
 # 2s3z | 3s5z | 5m_vs_6m | 10m_vs_11m | 27m_vs_30m | 3s5z_vs_3s6z | 3s_vs_5z | 6h_vs_8z | smacv2_5_units | smacv2_10_units | smacv2_20_units
 scenario:
   name: HeuristicEnemySMAX
--- a/mava/configs/env/smax.yaml
+++ b/mava/configs/env/smax.yaml
@@ -2,7 +2,7 @@
 defaults:
   - _self_
 
-env_name: Smax
+env_name: Smax  # Used for logging purposes and selection of the corresponding wrapper.
 # Possible tasks:
 # 2s3z | 3s5z | 5m_vs_6m | 10m_vs_11m | 27m_vs_30m | 3s5z_vs_3s6z | 3s_vs_5z | 6h_vs_8z | smacv2_5_units | smacv2_10_units | smacv2_20_units
 scenario:
@@ -17,3 +17,5 @@ kwargs:
   see_enemy_actions: True # Whether to enable enemy vision. If True, the enemy will be able to see the actions of the agent.
   walls_cause_death: True # Whether to kill the agent if it collides with a wall.
   attack_mode: "closest" # "closest" or "random". Whether to attack the closest enemy or a random enemy.
+
+add_agent_ids_to_state: False # This should still be False for now.
--- a/mava/configs/env/smax.yaml
+++ b/mava/configs/env/smax.yaml
@@ -18,4 +18,4 @@ kwargs:
   walls_cause_death: True # Whether to kill the agent if it collides with a wall.
   attack_mode: "closest" # "closest" or "random". Whether to attack the closest enemy or a random enemy.
 
-add_agent_ids_to_state: False # This should still be False for now.
+add_agent_ids_to_state: False # Currently has no effect.
--- a/mava/configs/env/smax.yaml
+++ b/mava/configs/env/smax.yaml
@@ -2,15 +2,14 @@
 defaults:
   - _self_
 
-env_name: HeuristicEnemySMAX
+env_name: SMAX
 # Possible scenarios:
 # 2s3z | 3s5z | 5m_vs_6m | 10m_vs_11m | 27m_vs_30m | 3s5z_vs_3s6z | 3s_vs_5z | 6h_vs_8z | smacv2_5_units | smacv2_10_units | smacv2_20_units
 scenario:
+  name: HeuristicEnemySMAX
   task_name: 2s3z
 
 kwargs:
   see_enemy_actions: True # Whether to enable enemy vision. If True, the enemy will be able to see the actions of the agent.
   walls_cause_death: True # Whether to kill the agent if it collides with a wall.
   attack_mode: "closest" # "closest" or "random". Whether to attack the closest enemy or a random enemy.
-
-env_wrapper: SmaxWrapper
--- a/mava/configs/env/smax.yaml
+++ b/mava/configs/env/smax.yaml
@@ -12,3 +12,5 @@ kwargs:
   see_enemy_actions: True # Whether to enable enemy vision. If True, the enemy will be able to see the actions of the agent.
   walls_cause_death: True # Whether to kill the agent if it collides with a wall.
   attack_mode: "closest" # "closest" or "random". Whether to attack the closest enemy or a random enemy.
+
+env_wrapper: SmaxWrapper
--- a/mava/utils/make_env.py
+++ b/mava/utils/make_env.py
@@ -114,11 +114,8 @@ def make_jaxmarl_env(
 
     # Create jaxmarl envs.
     wrapper_name = config.env.env_name
-    add_agent_ids_to_state = (
-        config.env.add_agent_ids_to_state
-        if hasattr(config.env, "add_agent_ids_to_state")
-        else False
-    )
+    add_agent_ids_to_state = config.env.add_agent_ids_to_state
+
     env = _jaxmarl_wrappers[wrapper_name](
         jaxmarl.make(env_name, **kwargs), add_global_state, add_agent_ids_to_state
     )
--- a/mava/utils/make_env.py
+++ b/mava/utils/make_env.py
@@ -114,8 +114,17 @@ def make_jaxmarl_env(
 
     # Create jaxmarl envs.
     wrapper_name = config.env.env_name
-    env = _jaxmarl_wrappers[wrapper_name](jaxmarl.make(env_name, **kwargs), add_global_state)
-    eval_env = _jaxmarl_wrappers[wrapper_name](jaxmarl.make(env_name, **kwargs), add_global_state)
+    add_agent_ids_to_state = (
+        config.env.add_agent_ids_to_state
+        if hasattr(config.env, "add_agent_ids_to_state")
+        else False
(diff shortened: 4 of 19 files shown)


mava/wrappers/jaxmarl.py
```diff
@@ -287,6 +304,16 @@ def action_mask(self, state: JaxMarlState) -> Array:
             mask = jnp.ones(self._action_shape, dtype=jnp.float32)
         return mask
 
-    def get_global_state(self, obs: Dict[str, Array]) -> Array:
+    def get_global_state(self, state: Array, obs: Dict[str, Array], agents_view: Array) -> Array:
         """Get global state from observation and copy it for each agent."""
-        return jnp.tile(jnp.array(obs["world_state"]), (self._env.num_agents, 1))
+        if hasattr(obs, "world_state"):
+            global_state = jnp.tile(jnp.array(obs["world_state"]), (self._env.num_agents, 1))
+        elif isinstance(self._env, MABraxEnv):
```
RuanJohn: I wonder if we shouldn't follow a similar approach to our Jumanji support and have `MABrax` have its own wrapper? Because the isinstance check is not the nicest :sweat_smile: 
Keen to hear your thoughts on this @sash-a?

sash-a: Ye totally agreed! Was actually talking to @liamclarkza about it this morning.

What I'd recommend:
* Have a `JaxMarlBaseWrapper` which keeps all the common functionality
* Have a `SmaxWrapper` and an `MaBraxWrapper` which extends `JaxMarlBaseWrapper` and implements the `_action_shape` and `get_global_shape` methods differently, I think that's all we'd need to do differently :thinking:

WiemKhlifi: I didn't think about that it's better for future use as well (if we aim to test other envs) 😅 I'll make the necessary changes then 🙌

mava/configs/env/mabrax.yaml
```diff
@@ -0,0 +1,16 @@
+# ---Environment Configs---
+defaults:
+  - _self_
+
+# Possible envs: [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
+env_name: ant_4x2
```
RuanJohn: Can we please standardise this to follow what we do in all the other environments as well where `env_name: MaBrax` and then we initialise different scenarios with `scenario.task_name`. It will also make the json logs much easier to handle later on.

mava/wrappers/jaxmarl.py
```diff
@@ -169,63 +171,55 @@ def __init__(self, env: MultiAgentEnv, has_global_state: bool = False, timelimit
         super().__init__(env)
         self._env: MultiAgentEnv
         self._timelimit = timelimit
+        self.agents = self._env.agents
+        self.num_agents = self._env.num_agents
+        self.has_global_state = has_global_state
+        self.state_size = state_size
```
sash-a: Slightly more standard to make this an abstract property, then you can implement the `state_size` property in the child classes, just check that jax is ok with this :thinking: 

```suggestion

    @property
    @abtractmethod
    def state_size(self):
        ...
```

mava/wrappers/jaxmarl.py
```diff
@@ -297,23 +281,55 @@ def discount_spec(self) -> specs.BoundedArray:
...
+        self.action_shape = self.action_spec().shape
+
+    def action_mask(self, state: JaxMarlState) -> Array:
+        """Get action mask for each agent."""
+        return jnp.ones(self.action_shape, dtype=jnp.float32)
+
+    def get_global_state(self, env_state: Array, obs: Dict[str, Array]) -> Array:
+        """Get global state from observation and copy it for each agent."""
+        # Use the global state of brax.
+        global_state = jnp.tile(env_state.obs, (self.num_agents, 1))
```
sash-a: It would be great to see if we can get the underlying envs obs to put here, I'll have a quick look into this :thinking:

mava/wrappers/jaxmarl.py
```diff
@@ -297,23 +281,55 @@ def discount_spec(self) -> specs.BoundedArray:
...
-            avail_actions = self._env.get_avail_actions(state)
-            mask = jnp.array(batchify(avail_actions, self.agents), dtype=jnp.float32)
-        else:
-            mask = jnp.ones(self._action_shape, dtype=jnp.float32)
-        return mask
+        pass
+
+    def get_global_state(self, env_state: Array, obs: Dict[str, Array]) -> Array:
+        """Get global state from observation for each agent."""
+        pass
```
sash-a: make this and action_mask abstract methods (also I just prefer `...` over `pass`, but that's up to you :smile: )

```suggestion
    @abstractmethod
    def get_global_state(self, env_state: Array, obs: Dict[str, Array]) -> Array:
        """Get global state from observation for each agent."""
        ...
```

mava/wrappers/jaxmarl.py
```diff
@@ -241,29 +235,19 @@ def step(
...
         single_agent_action_space = self._env.action_space(self.agents[0])
-        # we can't mask continuous actions, so just return a shape of 0 for this
-        n_actions = (
-            single_agent_action_space.n
-            if _is_discrete(self._env.action_space(self._env.agents[0]))
-            else 0
-        )
+
+        # we can't mask continuous actions, so we return a mask of shape 0.
+        n_actions = single_agent_action_space.n if _is_discrete(single_agent_action_space) else 0
```
sash-a: Could you also make this an abstract property, then we don't need to do this if as we know smax is discrete and mabrax is continuous

sash-a (changes requested):
Looking really good! Just make some methods abstract and make `state_size` an abstract property :smile: 

Also would be great if we could get the obs of the underlying env for mabrax global state as this is what mamujoco does

mava/wrappers/jaxmarl.py
sash-a: Also looks like this causes a size mismatch for mabrax as you have a action_mask shape of 0 in the spec but you return a different shape for the action mask in mabrax, so making it a property should help avoid this

mava/wrappers/jaxmarl.py
WiemKhlifi: Done 😅

mava/wrappers/jaxmarl.py
WiemKhlifi: Done² 😅

mava/configs/env/mabrax.yaml
WiemKhlifi: Done ² 🙌

mava/configs/env/smax.yaml
```diff
@@ -2,10 +2,11 @@
 defaults:
   - _self_
 
-env_name: HeuristicEnemySMAX
+env_name: SMAX
```
sash-a: can we make these names lower case

mava/configs/env/mabrax.yaml
```diff
@@ -0,0 +1,15 @@
+# ---Environment Configs---
+defaults:
+  - _self_
+
+# Possible envs: [ant_4x2, halfcheetah_6x1, hopper_3x1, humanoid_9|8, walker2d_2x3]
+env_name: MABRAX
+scenario:
+  name: ant_4x2
+  task_name: ant_4x2  # In this case: env_name = task_name
```
sash-a: If we don't use task_name lets leave it out of this config

mava/wrappers/jaxmarl.py
```diff
@@ -223,30 +227,26 @@ def step(
...
         )
         step_count = specs.BoundedArray(
-            (self._env.num_agents,), jnp.int32, 0, self._timelimit, "step_count"
+            (self.num_agents,), jnp.int32, 0, self._timelimit, "step_count"
         )
 
         if self.has_global_state:
             global_state = specs.Array(
-                (self._env.num_agents, self._env.state_size),
+                (self.num_agents, self.state_size),
```
sash-a, line 261: Can you fix this dtype below, it's not int32 but rather `agents_view.dtype`

mava/wrappers/jaxmarl.py
```diff
@@ -271,22 +271,76 @@ def action_spec(self) -> specs.Spec:
...
+
+
+class MabraxWrapper(JaxMarlWrapper):
+    """Wrraper for the Mabrax environment."""
+
+    def __init__(self, env: MABraxEnv, has_global_state: bool = False, timelimit: int = 1000):
+        super().__init__(env, has_global_state, timelimit)
+        self.action_shape = self.action_spec().shape
+
+    def state_size(self) -> chex.Array:
```
sash-a, line 362: won't work if it's not a property

```suggestion
    @property
    def state_size(self) -> chex.Array:
```

mava/wrappers/jaxmarl.py
```diff
@@ -271,22 +271,76 @@ def action_spec(self) -> specs.Spec:
...
+        ...
+
+
+class SmaxWrapper(JaxMarlWrapper):
+    """WWrapper for SMAX environment"""
+
+    def __init__(self, env: MultiAgentEnv, has_global_state: bool = False, timelimit: int = 500):
+        super().__init__(env, has_global_state, timelimit)
+
+    def state_size(self) -> chex.Array:
```
sash-a, line 329: won't work if it's not a property

```suggestion
    @property
    def state_size(self) -> chex.Array:
```

mava/wrappers/jaxmarl.py
```diff
@@ -271,22 +271,76 @@ def action_spec(self) -> specs.Spec:
...
+    def get_global_state(self, env_state: BraxState, obs: Dict[str, Array]) -> Array:
+        """Get global state from observation and copy it for each agent."""
+        # Use the global state of brax.
+        global_state = jnp.tile(env_state.obs, (self.num_agents, 1))
+
+        #  In this case, add_agent_id=False so the agent's ID must be added to the global state.
+        if self._env.homogenisation_method == "max":
+            agent_ids = jnp.eye(self.num_agents)
+            global_state = jnp.tile(env_state.obs, (self.num_agents, 1))
+            global_state = jnp.concatenate([agent_ids, global_state], axis=-1)
```
sash-a: I don't think we ever need to add agent_ids to global state as all critics would get the same state, so no need to differentiate?

Also @RuanJohn @OmaymaMahjoub do we want' multiple copies of the global state per agent? To me it seems like something you'd only pass this to one critic, at least this is how I'm doing it in MA-SAC :thinking:

sash-a (changes requested):
Couple more things and one question for @RuanJohn and @OmaymaMahjoub

mava/wrappers/jaxmarl.py
sash-a: Ok chatted to @RuanJohn about this and it seems like it was done for convenience and it would be a mission to change it in the PPO systems, so I'll work around this for SAC.

As for adding agent ids to the global state, can we comment here that we've found that having IDs in the global state is usually good, but can you also add an option for this so it's easier to experiment with? `if self._env.homogenisation_method == "max" and add_agent_ids:`

mava/configs/env/mabrax.yaml
```diff
@@ -0,0 +1,15 @@
+# --- Environment Configs---
+defaults:
+  - _self_
+
+env_name: Mabrax
```
sash-a:
```suggestion
env_name: MaBrax
```

mava/utils/make_env.py
```diff
@@ -42,6 +43,8 @@
     "LevelBasedForaging-v0": {"generator": LbfRandomGenerator, "wrapper": LbfWrapper},
 }
 
+_jaxmarl_wrappers = {"Smax": SmaxWrapper, "Mabrax": MabraxWrapper}
```
sash-a:
```suggestion
_jaxmarl_wrappers = {"Smax": SmaxWrapper, "MaBrax": MabraxWrapper}
```

mava/wrappers/jaxmarl.py
WiemKhlifi: Done ✅

mava/configs/env/mabrax.yaml
WiemKhlifi: Removed ✔️

mava/wrappers/jaxmarl.py
```diff
@@ -155,9 +158,25 @@ def jaxmarl_space_to_jumanji_spec(space: jaxmarl_spaces.Space) -> specs.Spec:
...
+        timelimit: int,
+        add_agent_ids_to_state: bool = False,
+    ) -> None:
+        """
+        Initialize the JaxMarlWrapper.
+
+        Args:
+        - env (MultiAgentEnv): The JaxMarl environment to wrap.
+        - has_global_state (bool): Whether the environment has global state.
+        - timelimit (int): The time limit for each episode.
```
```suggestion
        - timelimit (int): The time limit for each episode.
        - add_agent_ids_to_state (bool): whether to add the agent ids to the global state.
```

mava/wrappers/jaxmarl.py
```diff
@@ -271,22 +283,111 @@ def action_spec(self) -> specs.Spec:
...
+    def action_mask(self, state: JaxMarlState) -> Array:
+        """Get action mask for each agent."""
+        return jnp.ones((self.n_actions), dtype=jnp.float32)
+
+    def get_global_state(self, env_state: BraxState, obs: Dict[str, Array]) -> Array:
+        """Get global state from observation and copy it for each agent."""
+        # Use the global state of brax.
+        global_state = jnp.tile(env_state.obs, (self.num_agents, 1))
+
+        #  Including IDs in the global state can be generally beneficial.
```
sash-a: nit

```suggestion
        # Including IDs in the global state can be generally beneficial.
```

mava/wrappers/jaxmarl.py
```diff
@@ -271,22 +283,111 @@ def action_spec(self) -> specs.Spec:
...
+    def n_actions(self) -> chex.Array:
+        "Get the number of actions for each agent."
+        single_agent_action_space = self._env.action_space(self.agents[0])
+        return single_agent_action_space.n
 
-    def get_global_state(self, obs: Dict[str, Array]) -> Array:
+    def action_mask(self, state: JaxMarlState) -> Array:
+        """Get action mask for each agent."""
+        avail_actions = self._env.get_avail_actions(state)
+        return jnp.array(batchify(avail_actions, self.agents), dtype=jnp.float32)
```
sash-a: In the spec its a bool

```suggestion
        return jnp.array(batchify(avail_actions, self.agents), dtype=bool)
```

mava/wrappers/jaxmarl.py
```diff
@@ -271,22 +283,111 @@ def action_spec(self) -> specs.Spec:
...
+        )
+
+    @property
+    def n_actions(self) -> chex.Array:
+        "Get the number of actions for each agent."
+        return self.action_spec().shape[0]
+
+    def action_mask(self, state: JaxMarlState) -> Array:
+        """Get action mask for each agent."""
+        return jnp.ones((self.n_actions), dtype=jnp.float32)
```
sash-a, line 378: Same here

```suggestion
        return jnp.ones((self.n_actions), dtype=bool)
```

sash-a (changes requested):
lgtm except for the couple small changes I recommended

mava/wrappers/jaxmarl.py
```diff
@@ -246,6 +222,30 @@ def step(
...
 
+    def _create_observation(
+        self,
+        obs: Dict[str, Array],
+        env_state: State,
+        jaxmarl_state: Optional[JaxMarlState] = None,
+        reset: bool = False,
+    ) -> Union[Observation, ObservationGlobalState]:
+        """
+        Create an observation from the raw observation and environment state."""
```
```suggestion
        """Create an observation from the raw observation and environment state."""
```

mava/wrappers/jaxmarl.py
```diff
@@ -246,6 +222,30 @@ def step(
 
         return JaxMarlState(env_state, key, state.step + 1), ts
 
+    def _create_observation(
+        self,
+        obs: Dict[str, Array],
+        env_state: State,
```
```suggestion
        brax_state: BraxState,
```

mava/wrappers/jaxmarl.py
```diff
@@ -246,6 +222,30 @@ def step(
...
+            "agents_view": batchify(obs, self.agents),
+            "action_mask": self.action_mask(env_state),
+        }
+        if reset:
+            obs_data["step_count"] = jnp.zeros(self.num_agents, dtype=int)
+        elif jaxmarl_state is not None:
+            obs_data["step_count"] = jnp.repeat(jaxmarl_state.step, self.num_agents)
+
+        if self.has_global_state:
+            obs_data["global_state"] = self.get_global_state(env_state, obs)
```
```suggestion
            obs_data["global_state"] = self.get_global_state(brax_state, obs)
```

mava/wrappers/jaxmarl.py
```diff
@@ -246,6 +222,30 @@ def step(
...
+    ) -> Union[Observation, ObservationGlobalState]:
+        """
+        Create an observation from the raw observation and environment state."""
+        obs_data = {
+            "agents_view": batchify(obs, self.agents),
+            "action_mask": self.action_mask(env_state),
+        }
+        if reset:
+            obs_data["step_count"] = jnp.zeros(self.num_agents, dtype=int)
+        elif jaxmarl_state is not None:
```
sash-a: Might cause recompilation :thinking:

sash-a (changes requested):
Couple more minor things

mava/configs/env/lbf.yaml
```diff
@@ -4,7 +4,8 @@ defaults:
   - scenario: 2s-8x8-2p-2f-coop # [2s-8x8-2p-2f-coop, 8x8-2p-2f-coop, 2s-10x10-3p-3f, 10x10-3p-3f, 15x15-4p-3f, 15x15-4p-5f]
 # Further environment config details in "2s-8x8-2p-2f-coop" file.
 
-env_name: LevelBasedForaging-v0
+env_name: LevelBasedForaging
```
RuanJohn: Can we please just add a comment in the env yaml files that the env_name is just for logging purposes.

mava/wrappers/jaxmarl.py
```diff
@@ -168,86 +188,79 @@ def __init__(self, env: MultiAgentEnv, has_global_state: bool = False, timelimit
...
+        obs: Dict[str, Array],
+        brax_state: BraxState,
+        jaxmarl_state: Optional[JaxMarlState] = None,
+        reset: bool = False,
+    ) -> Union[Observation, ObservationGlobalState]:
+        """Create an observation from the raw observation and environment state."""
+        obs_data = {
+            "agents_view": batchify(obs, self.agents),
+            "action_mask": self.action_mask(brax_state),
+            "step_count": jnp.zeros(self.num_agents, dtype=int),
```
WiemKhlifi: The extra line here will be removed in the next push 🤞 
```python
"step_count": jnp.zeros(self.num_agents, dtype=int),
```

mava/configs/env/lbf.yaml
WiemKhlifi: Sure! and for Jaxmal it's used for logging and selecting the wrapper

mava/configs/env/smax.yaml
```diff
@@ -16,3 +17,5 @@ kwargs:
   see_enemy_actions: True # Whether to enable enemy vision. If True, the enemy will be able to see the actions of the agent.
   walls_cause_death: True # Whether to kill the agent if it collides with a wall.
   attack_mode: "closest" # "closest" or "random". Whether to attack the closest enemy or a random enemy.
+
+add_agent_ids_to_state: False # This should still be False for now.
```
sash-a:
```suggestion
add_agent_ids_to_state: False # Currently has no effect
```

sash-a (dismissed):
1 tiny thing but I am happy with this :smile:

mava/utils/make_env.py
```diff
@@ -110,8 +113,15 @@ def make_jaxmarl_env(
...
-    eval_env = JaxMarlWrapper(jaxmarl.make(env_name, **kwargs), add_global_state)
+    wrapper_name = config.env.env_name
+    add_agent_ids_to_state = config.env.add_agent_ids_to_state
+
+    env = _jaxmarl_wrappers[wrapper_name](
+        jaxmarl.make(env_name, **kwargs), add_global_state, add_agent_ids_to_state
+    )
+    eval_env = _jaxmarl_wrappers[wrapper_name](
+        jaxmarl.make(env_name, **kwargs), add_global_state, add_agent_ids_to_state
+    )
```
RuanJohn: NIT: but a bit cleaner in my opinion to just use the config directly. 

```suggestion

    env = _jaxmarl_wrappers[config.env.env_name](
        jaxmarl.make(env_name, **kwargs), add_global_state, config.env.add_agent_ids_to_state
    )
    eval_env = _jaxmarl_wrappers[config.env.env_name](
        jaxmarl.make(env_name, **kwargs), add_global_state, config.env.add_agent_ids_to_state
    )
```

mava/wrappers/jaxmarl.py
```diff
@@ -271,22 +283,110 @@ def action_spec(self) -> specs.Spec:
...
+
+    @property
+    @abstractmethod
+    def state_size(self) -> chex.Array:
+        "Get the sate size of the global observation"
+        ...
+
+
+class SmaxWrapper(JaxMarlWrapper):
+    """WWrapper for SMAX environment"""
```
RuanJohn: NIT
```suggestion
    """Wrapper for SMAX environment"""
```

mava/wrappers/jaxmarl.py
```diff
@@ -271,22 +283,110 @@ def action_spec(self) -> specs.Spec:
...
+        avail_actions = self._env.get_avail_actions(state)
+        return jnp.array(batchify(avail_actions, self.agents), dtype=bool)
+
+    def get_global_state(self, brax_state: BraxState, obs: Dict[str, Array]) -> Array:
+        """Get global state from observation and copy it for each agent."""
+        return jnp.tile(jnp.array(obs["world_state"]), (self.num_agents, 1))
+
+
+class MabraxWrapper(JaxMarlWrapper):
+    """Wrraper for the Mabrax environment."""
```
RuanJohn, line 350:
```suggestion
    """Wrapper for the Mabrax environment."""
```

RuanJohn (commented):
Some minor suggestions for me :)

RuanJohn (approved):
Looks good! Thank you for all the effort @WiemKhlifi :fire:

sash-a (approved):
:jax: :marl: :fire:

PR #1020, opened 2024-02-14
feat: system folders

sash-a:
## What?

Seeing as we're soon to have many more systems and each system is likely to have a ff and rec and ma and i version, I think moving the systems into folders makes sense. We can then have a `types.py` inside those folders so that our main `types.py` doesn't get too bloated (since each system will have a different params type, different learner state etc.)


mava/utils/checkpointing.py
```diff
@@ -146,6 +147,8 @@ def restore_params(
         input_params: Params,
         timestep: Optional[int] = None,
         restore_hstates: bool = False,
+        TParams: Type[Params] = Params,  # noqa: N803
```
RuanJohn: Please add a docstring to explain the types here.

liamclarkza (dismissed):
LGTM 👍

mava/types.py
```diff
@@ -147,7 +78,7 @@ class RNNEvalState(NamedTuple):
     episode_return: chex.Array
 
 
-MavaState = TypeVar("MavaState", LearnerState, RNNLearnerState, EvalState, RNNEvalState)
+MavaState = TypeVar("MavaState")  # Types like: LearnerState | EvalState.
```
WiemKhlifi: This seems hard to understand since the learner state is in the ppo/types.py file, can we make it clearer?
It's not the best idea, but we can say something like "For more info on the LearnerState type, please refer to ppo/types.py."?

WiemKhlifi (dismissed):
LGTM 🥳 Just one minor comment for MavaState .

WiemKhlifi (approved):
🛥️

liamclarkza (approved):
👍

PR #1022, opened 2024-02-14
Add return metrics for experiments and corresponding Hydra config option

liamclarkza:
## What?
We now return a performance evaluation metric from our experiment.

## Why?
Allows for us to hyperparameter tune using Hydra & OpTuna.

## How?
Modifies the run_experiment method and hydra_entry_point in the various systems to return a metric to be used for hyperparameter tuning through Hydra (using something like OpTuna). The Hydra config option, config.env.eval_metric,is used to select the metric (episode_return or win_rate) for an environment.

### DIFF
--- a/mava/configs/env/lbf.yaml
+++ b/mava/configs/env/lbf.yaml
@@ -8,5 +8,9 @@ env_name: LevelBasedForaging-v0
 # Choose whether to aggregate the list of individual rewards and use the team reward (default setting) OR use_individual_rewards=True.
 use_individual_rewards: False  # If True, use the list of individual rewards.
 
+# Defines the metric that will be used to evaluate the performance of the agent.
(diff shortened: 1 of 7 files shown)


EdanToledo:
LGTM

sash-a (dismissed):
Looks great @liamclarkza, awesome first PR :fire:

RuanJohn (changes requested):
Really epic first PR @liamclarkza :fire: 

Just a small detail from my side. Right now, this will return the evaluator output of the absolute metric evaluator. Can we please change it so that we return the results from the final online evaluation run?

RuanJohn (approved):
Thank you for this @liamclarkza! :fire:

EdanToledo (approved):

PR #1023, opened 2024-02-15
chore: remove neptune project argument from system level logger configs

RuanJohn:
## What?
Only set the Neptune project name in the base logger config.

### DIFF
--- a/mava/configs/logger/ff_ippo.yaml
+++ b/mava/configs/logger/ff_ippo.yaml
@@ -5,5 +5,4 @@ system_name: ff_ippo
 
 # --- Other logger kwargs ---
 kwargs:
-  neptune_project: Instadeep/Mava
   neptune_tag: [ff-ippo]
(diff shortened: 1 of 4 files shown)


sash-a (dismissed):
:rabbit:

sash-a (approved):
:rabbit2:

OmaymaMahjoub (changes requested):
Can we keep neptune tag per system? because when you run a sweep you want each of the system have a different tag automatically without the need to set neptune tag in the cmd

OmaymaMahjoub:
> Can we keep neptune tag per system? because when you run a sweep you want each of the system have a different tag automatically without the need to set neptune tag in the cmd

We had an internal discussion: Since we have config.logger.system_name we can use it to filter the systems, no need to have different tags for different systems

OmaymaMahjoub (approved):
Thanks @RuanJohn :fire:

PR #1028, opened 2024-02-16
feat: add matrax environment

JemmaLDaniel:
## What?
Added support for Matrax via a wrapper and a config file.
## Why?
This makes Mava compatible with Matrax.
## How?
The Matrax wrapper transforms observations to the standard Mava API, and the config allows us to choose between and run Matrax's Climbing, Penality, Conflict and NoConflict games with Mava's implemented MARL algorithms.

### DIFF
--- a/mava/configs/env/matrax.yaml
+++ b/mava/configs/env/matrax.yaml
@@ -20,6 +20,8 @@ scenario:
   task_name: Climbing-stateless-v0
   name: Matrax
 
+# Defines the metric that will be used to evaluate the performance of the agent.
+# This metric is returned at the end of an experiment and can be used for hyperparameter tuning.
 eval_metric: episode_return
 
 kwargs:
--- a/mava/configs/env/matrax.yaml
+++ b/mava/configs/env/matrax.yaml
@@ -1,0 +1,26 @@
+# ---Environment Configs---
+defaults:
+  - _self_
+
+env_name: Matrax
+
+scenario:
+  task_name: Climbing-stateless-v0
+
+eval_metric: episode_return
+
+# Possible scenarios:
+#   1. Penalty-{k}-{state}-v0
+#       1.1. k = {0, 25, 50, 75, 100}
+#       1.2. state = {stateful, stateless}
+#   2. Climbing-{state}-v0
+#       2.1. state = {stateful, stateless}
+#   3. NoConflict-{id}-{state}-v0
+#       3.1. id = {1, 2, 3, ..., 20, 21}
+#       3.2. state = {stateful, stateless}
+#   4. Conflict-{id}-{state}-v0
+#       4.1. id = {1, 2, 3, ..., 56, 57}
+#       4.2. state = {stateful, stateless}
+
+kwargs:
+  time_limit: 500
--- a/mava/configs/env/matrax.yaml
+++ b/mava/configs/env/matrax.yaml
@@ -2,12 +2,7 @@
 defaults:
   - _self_
 
-env_name: Matrax
-
-scenario:
-  task_name: Climbing-stateless-v0
-
-eval_metric: episode_return
+env_name: Matrax  # Used for logging purposes.
 
 # Possible scenarios:
 #   1. Penalty-{k}-{state}-v0
@@ -21,6 +16,11 @@ eval_metric: episode_return
 #   4. Conflict-{id}-{state}-v0
 #       4.1. id = {1, 2, 3, ..., 56, 57}
 #       4.2. state = {stateful, stateless}
+scenario:
+  task_name: Climbing-stateless-v0
+  name: Matrax
+
+eval_metric: episode_return
 
 kwargs:
   time_limit: 500
--- a/mava/utils/make_env.py
+++ b/mava/utils/make_env.py
@@ -16,6 +16,7 @@ from typing import Tuple
 
 import jaxmarl
 import jumanji
+import matrax
 from jaxmarl.environments.smax import map_name_to_scenario
 from jumanji.env import Environment
 from jumanji.environments.routing.lbf.generator import (
@@ -32,6 +33,7 @@ from mava.wrappers import (
     GlobalStateWrapper,
     JaxMarlWrapper,
     LbfWrapper,
+    MatraxWrapper,
     RecordEpisodeMetrics,
     RwareWrapper,
 )
@@ -40,6 +42,11 @@ from mava.wrappers import (
 _jumanji_registry = {
     "RobotWarehouse-v0": {"generator": RwareRandomGenerator, "wrapper": RwareWrapper},
     "LevelBasedForaging-v0": {"generator": LbfRandomGenerator, "wrapper": LbfWrapper},
+}
+
+# Define a different registry for Matrax since it has no generator.
+_matrax_registry = {
+    "Matrax": {"wrapper": MatraxWrapper},
 }
 
 
@@ -124,6 +131,38 @@ def make_jaxmarl_env(
     return env, eval_env
 
 
+def make_matrax_env(
+    env_name: str, config: DictConfig, add_global_state: bool = False
+) -> Tuple[Environment, Environment]:
+    """
+    Creates Matrax environments for training and evaluation.
+
+    Args:
+        env_name (str): The name of the environment to create.
+        config (Dict): The configuration of the environment.
+        add_global_state (bool): Whether to add the global state to the observation.
+
+    Returns:
+        A tuple of the Matrax environments.
+    """
+    # Select the Matrax wrapper.
+    wrapper = _matrax_registry[env_name]["wrapper"]
+
+    # Create envs.
+    task_name = config["env"]["scenario"]["task_name"]
+    env = matrax.make(task_name, **config.env.kwargs)
+    eval_env = matrax.make(task_name, **config.env.kwargs)
+    env, eval_env = wrapper(env), wrapper(eval_env)
+
+    env = add_optional_wrappers(env, config, add_global_state)
+    eval_env = add_optional_wrappers(eval_env, config, add_global_state)
+
+    env = AutoResetWrapper(env)
+    env = RecordEpisodeMetrics(env)
+
+    return env, eval_env
+
+
 def make(config: DictConfig, add_global_state: bool = False) -> Tuple[Environment, Environment]:
     """
     Create environments for training and evaluation..
@@ -141,5 +180,7 @@ def make(config: DictConfig, add_global_state: bool = False) -> Tuple[Environmen
         return make_jumanji_env(env_name, config, add_global_state)
     elif env_name in jaxmarl.registered_envs:
         return make_jaxmarl_env(env_name, config, add_global_state)
+    elif env_name in _matrax_registry:
+        return make_matrax_env(env_name, config, add_global_state)
     else:
         raise ValueError(f"{env_name} is not a supported environment.")
--- a/mava/utils/make_env.py
+++ b/mava/utils/make_env.py
@@ -145,12 +145,12 @@ def make_matrax_env(
     Creates Matrax environments for training and evaluation.
 
     Args:
-        env_name (str): The name of the environment to create.
-        config (Dict): The configuration of the environment.
-        add_global_state (bool): Whether to add the global state to the observation.
+        env_name: The name of the environment to create.
+        config: The configuration of the environment.
+        add_global_state: Whether to add the global state to the observation.
 
     Returns:
-        A tuple of the Matrax environments.
+        A tuple containing a train and evaluation Matrax environment.
     """
     # Select the Matrax wrapper.
     wrapper = _matrax_registry[env_name]["wrapper"]
--- a/mava/utils/make_env.py
+++ b/mava/utils/make_env.py
@@ -46,9 +46,7 @@ _jumanji_registry = {
 }
 
 # Define a different registry for Matrax since it has no generator.
-_matrax_registry = {
-    "Matrax": {"wrapper": MatraxWrapper},
-}
+_matrax_registry = {"Matrax": MatraxWrapper}
 
 _jaxmarl_wrappers = {"Smax": SmaxWrapper, "MaBrax": MabraxWrapper}
 
@@ -153,7 +151,7 @@ def make_matrax_env(
         A tuple containing a train and evaluation Matrax environment.
     """
     # Select the Matrax wrapper.
-    wrapper = _matrax_registry[env_name]["wrapper"]
+    wrapper = _matrax_registry[env_name]
 
     # Create envs.
     task_name = config["env"]["scenario"]["task_name"]
--- a/mava/wrappers/matrax.py
+++ b/mava/wrappers/matrax.py
@@ -1,0 +1,93 @@
+# Copyright 2022 InstaDeep Ltd. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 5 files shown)


mava/configs/env/matrax.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - _self_
+
+env_name: Matrax
+
+scenario:
+  task_name: Climbing-stateless-v0
+
+eval_metric: episode_return
+
+# Possible scenarios:
```
RuanJohn, line 7: NIT: Please move this up to be above `scenario` to be consistent with the other env configs.

mava/utils/make_env.py
```diff
@@ -124,6 +131,38 @@ def make_jaxmarl_env(
     return env, eval_env
 
 
+def make_matrax_env(
```
RuanJohn, line 139: Very nice! :sunglasses:

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+
+    def modify_timestep_reset(self, timestep: TimeStep) -> TimeStep[Observation]:
+        """Modify the timestep for `step` and `reset`."""
+        observation = Observation(
+            agents_view=timestep.observation.agent_obs,
+            action_mask=jnp.ones((self._num_agents, self._num_actions), dtype=bool),
+            step_count=jnp.repeat(timestep.observation.step_count, self._num_agents),
+        )
+        reward = jnp.repeat(timestep.reward, self._num_agents)
```
RuanJohn: Can you please install the latest version of Matrax and double check that this still runs? I am pretty sure we fixed this so that the reset won't require reward duplication anymore.

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,93 @@
...
+    def modify_timestep_step(self, timestep: TimeStep) -> TimeStep[Observation]:
+        """Modify the timestep for `step` and `reset`."""
+        observation = Observation(
+            agents_view=timestep.observation.agent_obs,
+            action_mask=jnp.ones((self._num_agents, self._num_actions), dtype=bool),
+            step_count=jnp.repeat(timestep.observation.step_count, self._num_agents),
+        )
+        return timestep.replace(
+            observation=observation,
+            reward=timestep.reward * 1.0,
```
RuanJohn: Is the `* 1.0` to convert the reward to a float?

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,93 @@
...
+            jnp.bool_,
+            "action_mask",
+        )
+        return specs.Spec(
+            Observation,
+            "ObservationSpec",
+            agents_view=env_observation_spec,
+            action_mask=action_mask,
+            step_count=step_count,
+        )
```
RuanJohn: Since we only use this once, we can just define it in the returned `Spec`.

```suggestion
        """Specification of the observation of the environment."""
        step_count = specs.BoundedArray(
            (self._num_agents,),
            jnp.int32,
            [0] * self._num_agents,
            [self.time_limit] * self._num_agents,
            "step_count",
        )
        action_mask = specs.Array(
            (self._num_agents, self._num_actions),
            jnp.bool_,
            "action_mask",
        )
        return specs.Spec(
            Observation,
            "ObservationSpec",
            agents_view=self._env.observation_spec().agent_obs,
            action_mask=action_mask,
            step_count=step_count,
        )
```

requirements/requirements.txt
```diff
@@ -7,6 +7,7 @@ jax
 jaxlib
 jaxmarl
 jumanji @ git+https://github.com/sash-a/jumanji
+matrax
```
RuanJohn:
```suggestion
matrax @ git+https://github.com/instadeepai/matrax
```

mava/wrappers/matrax.py
RuanJohn: This should also allow you to only have single `modify_timstep` method that should work for both `step` and `reset`.

mava/utils/make_env.py
```diff
@@ -124,6 +131,38 @@ def make_jaxmarl_env(
...
+    """
+    Creates Matrax environments for training and evaluation.
+
+    Args:
+        env_name (str): The name of the environment to create.
+        config (Dict): The configuration of the environment.
+        add_global_state (bool): Whether to add the global state to the observation.
+
+    Returns:
+        A tuple of the Matrax environments.
```
RuanJohn: NIT: 
```suggestion
        A tuple containing a train and evaluation Matrax environment.
```

mava/utils/make_env.py
```diff
@@ -124,6 +131,38 @@ def make_jaxmarl_env(
...
 
 
+def make_matrax_env(
+    env_name: str, config: DictConfig, add_global_state: bool = False
+) -> Tuple[Environment, Environment]:
+    """
+    Creates Matrax environments for training and evaluation.
+
+    Args:
+        env_name (str): The name of the environment to create.
```
RuanJohn: NIT: 
```suggestion
        env_name: The name of the environment to create.
```

mava/utils/make_env.py
```diff
@@ -124,6 +131,38 @@ def make_jaxmarl_env(
...
+        config (Dict): The configuration of the environment.
+        add_global_state (bool): Whether to add the global state to the observation.
```
RuanJohn: NIT: 
```suggestion
        config: The configuration of the environment.
        add_global_state: Whether to add the global state to the observation.
```

RuanJohn (changes requested):
Thank you for this @JemmaLDaniel! A very nice first PR into Mava :fire: 
Just a few minor comments from my side.

mava/utils/make_env.py
```diff
@@ -42,6 +44,11 @@
     "LevelBasedForaging-v0": {"generator": LbfRandomGenerator, "wrapper": LbfWrapper},
 }
 
+# Define a different registry for Matrax since it has no generator.
+_matrax_registry = {
+    "Matrax": {"wrapper": MatraxWrapper},
+}
```
sash-a: Because we only have 1 wrapper we don't need a registry for matrax :smile: 
```suggestion
```

mava/utils/make_env.py
```diff
@@ -124,6 +131,38 @@ def make_jaxmarl_env(
...
+        A tuple of the Matrax environments.
+    """
+    # Select the Matrax wrapper.
+    wrapper = _matrax_registry[env_name]["wrapper"]
+
+    # Create envs.
+    task_name = config["env"]["scenario"]["task_name"]
+    env = matrax.make(task_name, **config.env.kwargs)
+    eval_env = matrax.make(task_name, **config.env.kwargs)
+    env, eval_env = wrapper(env), wrapper(eval_env)
```
sash-a: We can just use the wrapper directly here

```suggestion
    # Create envs.
    task_name = config["env"]["scenario"]["task_name"]
    env = matrax.make(task_name, **config.env.kwargs)
    eval_env = matrax.make(task_name, **config.env.kwargs)
    env, eval_env = MatraxWrapper(env), MatraxWrapper(eval_env)
```

mava/utils/make_env.py
```diff
@@ -141,5 +180,7 @@ def make(config: DictConfig, add_global_state: bool = False) -> Tuple[Environmen
         return make_jumanji_env(env_name, config, add_global_state)
     elif env_name in jaxmarl.registered_envs:
         return make_jaxmarl_env(env_name, config, add_global_state)
+    elif env_name in _matrax_registry:
```
sash-a, line 188:
```suggestion
    elif env_name == "Matrax":
```

mava/utils/make_env.py
sash-a: I think the `make_matrax_env` and the `make_jumanji_env` methods are basically the same, so what we could do is join the matrax and jumanji registries and then just use the `make_jumanji_env` methods. 

@RuanJohn @JemmaLDaniel what do you think?

mava/wrappers/matrax.py
sash-a: if so: `timestep.reward.astype(float)` is more clear

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,93 @@
...
+        step_count = specs.BoundedArray(
+            (self._num_agents,),
+            jnp.int32,
+            [0] * self._num_agents,
+            [self.time_limit] * self._num_agents,
+            "step_count",
+        )
+        action_mask = specs.Array(
+            (self._num_agents, self._num_actions),
+            jnp.bool_,
```
sash-a:
```suggestion
            bool,
```

sash-a (dismissed):
Awesome first PR! Looking almost ready to go :fire: :rocket:

mava/utils/make_env.py
JemmaLDaniel: That is definitely more concise, but I think keeping them separate is slightly easier to read and understand because it is in keeping with the choice for Matrax as a separate suite of environments to Jumanji. Another solution is to merge them into one registry but change the name to be more all-encompassing than `make_jumanji_env` and `_jumanji_registry`?

mava/wrappers/matrax.py
JemmaLDaniel: Fixed--thanks!

mava/wrappers/matrax.py
JemmaLDaniel: It was, but it's no longer necessary with the updated Matrax install 😄

mava/utils/make_env.py
JemmaLDaniel: It seems like this change and the previous one to use `MatraxWrapper` directly make `_matrax_registry` completely unnecessary. Could we leave it out altogether?

mava/utils/make_env.py
RuanJohn: I agree that the methods are essentially the same, but I think it is better to keep them separate and to have a matrax registry. The reason for this is that the poly games may become part of matrax later on keeping env suites separate from each other sets a nice pattern.

sash-a: Ok sure happy to keep with this pattern :smile:

mava/utils/make_env.py
sash-a: Ye I think leave it in for now, just like @RuanJohn said it keeps the pattern

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,78 @@
...
+        self._num_agents = self._env.num_agents
+        self.time_limit = self._env.time_limit
+        self._num_actions = self._env.num_actions
+        self.num_obs_features = self._env.observation_spec().agent_obs.shape[-1]
+
+    def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
+        """Modify the timestep for `step` and `reset`."""
+        observation = Observation(
+            agents_view=timestep.observation.agent_obs,
+            action_mask=jnp.ones((self._num_agents, self._num_actions), dtype=bool),
```
WiemKhlifi: Since this is fixed, can we move it up so we don't have to calculate it over and over at each timestep:
```suggestion
            action_mask=self.action_mask

```

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,78 @@
...
+from mava.types import Observation, State
+
+
+class MatraxWrapper(Wrapper):
+    def __init__(self, env: Environment):
+        super().__init__(env)
+        self._num_agents = self._env.num_agents
+        self.time_limit = self._env.time_limit
+        self._num_actions = self._env.num_actions
+        self.num_obs_features = self._env.observation_spec().agent_obs.shape[-1]
```
WiemKhlifi: Same,  the `num_obs_features` isn't used here can you remove it please and add this as discussed in the suggestion above:
```suggestion
        self.action_mask = jnp.ones((self._num_agents, self._num_actions), dtype=bool)
```

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,78 @@
...
+        state, timestep = self._env.step(state, action)
+        return state, self.modify_timestep(timestep)
+
+    def observation_spec(self) -> specs.Spec[Observation]:
+        """Specification of the observation of the environment."""
+        step_count = specs.BoundedArray(
+            (self._num_agents,),
+            jnp.int32,
+            [0] * self._num_agents,
+            [self.time_limit] * self._num_agents,
```
WiemKhlifi:
``` suggestion
            [self._env.time_limit] * self._num_agents,
```

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,78 @@
...
+from jumanji.wrappers import Wrapper
+
+from mava.types import Observation, State
+
+
+class MatraxWrapper(Wrapper):
+    def __init__(self, env: Environment):
+        super().__init__(env)
+        self._num_agents = self._env.num_agents
+        self.time_limit = self._env.time_limit
```
WiemKhlifi: Can we access it directly in the spec, since we won't have to use it again
``` suggestion

```

mava/wrappers/matrax.py
```diff
@@ -0,0 +1,78 @@
...
+import jax.numpy as jnp
+from jumanji import specs
+from jumanji.env import Environment
+from jumanji.types import TimeStep
+from jumanji.wrappers import Wrapper
+
+from mava.types import Observation, State
+
+
+class MatraxWrapper(Wrapper):
```
WiemKhlifi, line 27:
``` suggestion
class MatraxWrapper(Wrapper):
    """Multi-agent wrapper for the Matrax environment."""

```

mava/utils/make_env.py
```diff
@@ -131,6 +138,38 @@ def make_jaxmarl_env(
...
+    Args:
+        env_name: The name of the environment to create.
+        config: The configuration of the environment.
+        add_global_state: Whether to add the global state to the observation.
+
+    Returns:
+        A tuple containing a train and evaluation Matrax environment.
+    """
+    # Select the Matrax wrapper.
+    wrapper = _matrax_registry[env_name]["wrapper"]
```
WiemKhlifi:
``` suggestion
    wrapper = _matrax_registry[env_name]
```

mava/utils/make_env.py
```diff
@@ -43,6 +45,11 @@
     "LevelBasedForaging-v0": {"generator": LbfRandomGenerator, "wrapper": LbfWrapper},
 }
 
+# Define a different registry for Matrax since it has no generator.
+_matrax_registry = {
+    "Matrax": {"wrapper": MatraxWrapper},
+}
```
WiemKhlifi: Small suggestion here since we only using a wrapper (no generators like in jumanji):
``` suggestion
_matrax_registry = {
    "Matrax": {MatraxWrapper},
}
```

WiemKhlifi (changes requested):

mava/configs/env/matrax.yaml
```diff
@@ -0,0 +1,26 @@
...
+#   3. NoConflict-{id}-{state}-v0
+#       3.1. id = {1, 2, 3, ..., 20, 21}
+#       3.2. state = {stateful, stateless}
+#   4. Conflict-{id}-{state}-v0
+#       4.1. id = {1, 2, 3, ..., 56, 57}
+#       4.2. state = {stateful, stateless}
+scenario:
+  task_name: Climbing-stateless-v0
+  name: Matrax
+
```
WiemKhlifi, line 22:
``` suggestion
# Defines the metric that will be used to evaluate the performance of the agent.

```

mava/utils/make_env.py
JemmaLDaniel: I agree, makes more sense with the separately defined registries to tailor it more to Matrax 😄

WiemKhlifi (approved):
Very clean PR! Thank you @JemmaLDaniel 🔥🚀

RuanJohn (approved):
Epic @JemmaLDaniel :fire:

PR #1029, opened 2024-02-19
refactor: Remove agent's ID from the global_state

WiemKhlifi:
## What?
These changes modify the AgentIDWrapper and JaxMARL wrappers to exclusively support the addition of the agent ID in the agent view, rather than also including it in the global state.

## Why?
The adjustment is made because adding an agent's ID to its view can be useful where agents require unique identification. However, adding this information to the global state does not seem necessary, especially given current support for centralized training and the use of a shared critic.

### DIFF
--- a/mava/configs/default_ff_ippo.yaml
+++ b/mava/configs/default_ff_ippo.yaml
@@ -3,5 +3,5 @@ defaults:
   - arch: anakin
   - system: ff_ippo
   - network: mlp
-  - env: smax
+  - env: rware
   - _self_
--- a/mava/configs/default_ff_ippo.yaml
+++ b/mava/configs/default_ff_ippo.yaml
@@ -3,5 +3,5 @@ defaults:
   - arch: anakin
   - system: ff_ippo
   - network: mlp
-  - env: rware
+  - env: smax
   - _self_
--- a/mava/utils/make_env.py
+++ b/mava/utils/make_env.py
@@ -16,6 +16,7 @@ from typing import Tuple
 
 import jaxmarl
 import jumanji
+import matrax
 from jaxmarl.environments.smax import map_name_to_scenario
 from jumanji.env import Environment
 from jumanji.environments.routing.lbf.generator import (
@@ -32,6 +33,7 @@ from mava.wrappers import (
     GlobalStateWrapper,
     LbfWrapper,
     MabraxWrapper,
+    MatraxWrapper,
     RecordEpisodeMetrics,
     RwareWrapper,
     SmaxWrapper,
@@ -42,6 +44,9 @@ _jumanji_registry = {
     "RobotWarehouse-v0": {"generator": RwareRandomGenerator, "wrapper": RwareWrapper},
     "LevelBasedForaging-v0": {"generator": LbfRandomGenerator, "wrapper": LbfWrapper},
 }
+
+# Define a different registry for Matrax since it has no generator.
+_matrax_registry = {"Matrax": MatraxWrapper}
 
 _jaxmarl_wrappers = {"Smax": SmaxWrapper, "MaBrax": MabraxWrapper}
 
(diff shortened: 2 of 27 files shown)


mava/configs/default_ff_ippo.yaml
```diff
@@ -3,5 +3,5 @@ defaults:
   - arch: anakin
   - system: ff_ippo
   - network: mlp
-  - env: rware
+  - env: smax
```
WiemKhlifi: This will be reverted with the next push

mava/utils/make_env.py
```diff
@@ -52,16 +52,18 @@
...
+    env: Environment,
+    config: DictConfig,
+    add_global_state: bool = False,
 ) -> Environment:
     # Add the global state to observation.
     if add_global_state:
         env = GlobalStateWrapper(env)
 
     # Add agent id to observation.
+
```
sash-a: nit:
```suggestion
```

mava/wrappers/jaxmarl.py
sash-a, line 1: So much cleaner :heart_eyes:

mava/wrappers/observation.py
```diff
@@ -34,28 +36,19 @@ def __init__(self, env: Environment, has_global_state: bool = False):
     def _add_agent_ids(
         self, timestep: TimeStep, num_agents: int
     ) -> Union[Observation, ObservationGlobalState]:
+        """Adds agent IDs to the observation."""
         agent_ids = jnp.eye(num_agents)
-        new_agents_view = jnp.concatenate([agent_ids, timestep.observation.agents_view], axis=-1)
+        obs_data = {
```
sash-a: Instead of all this couldn't we just `return obsevation.replace(agents_view=new_agents_view)`?  (might be `._replace` I can never remember)

mava/wrappers/observation.py
```diff
@@ -87,14 +80,10 @@ def observation_spec(
 
         if self.has_global_state:
             wrapped_state_spec = obs_spec.global_state
-            wrapped_state_shape = wrapped_state_spec.shape
-            wrapped_state_dtype = wrapped_state_spec.dtype
 
-            wrapped_state_shape = (
-                *wrapped_state_shape[:-1],
-                wrapped_state_shape[-1] + self._env.num_agents,
+            global_state = specs.Array(
```
sash-a: We can remove this whole `if` as now we know the global state isn't changing so it's actually not doing anything :smile:

mava/wrappers/observation.py
```diff
@@ -34,28 +36,19 @@ def __init__(self, env: Environment, has_global_state: bool = False):
     def _add_agent_ids(
         self, timestep: TimeStep, num_agents: int
     ) -> Union[Observation, ObservationGlobalState]:
+        """Adds agent IDs to the observation."""
```
sash-a, line 38: Double check this, but I *think* we don't need the `has_global_state` arg anymore?

sash-a (changes requested):
Looking great some minor comments and please update the PR description and name :smile:

mava/wrappers/observation.py
WiemKhlifi: Totally agree I forgot this hack I'll update it now 🙌

mava/wrappers/observation.py
WiemKhlifi: Yes with the .replace() we can remove the `has_global_state` flag.

mava/wrappers/observation.py
WiemKhlifi: Right 😅

sash-a (approved):
Really like this change, thanks @WiemKhlifi!

RuanJohn (approved):
Thank you @WiemKhlifi :fire:

PR #1030, opened 2024-02-19
Fix: fixes win_rate not being logged due to config change

liamclarkza:
## What?
Small fix in the logger for a bug in the config files that caused the logger not to record win_rate


WiemKhlifi (approved):
Nice catch @liamclarkza :raised_hands:

RuanJohn (approved):
Thanks @liamclarkza :fire:

PR #1034, opened 2024-02-20
Fix: converts dict values to floats in the logger

liamclarkza:
## What?
Small bug fix that prevents crashes that occur when integer values are formatted using float format strings in the logger.


sash-a (approved):
Thanks :raised_hands:

WiemKhlifi (approved):
Just faced this error now 😅 Thank you @liamclarkza for the fix 🙏

PR #1035, opened 2024-02-20
feat: Jumanji Connector wrapper

SimonDuToit:
Add wrapper for Jumanji Connector environment and some scenarios. The wrapper includes a custom global state.


mava/configs/env/scenario/con-10x10x5a.yaml
```diff
@@ -0,0 +1,6 @@
+# The config of the 15x15-4p-5f scenario with the VectorObserver set as default
```
sash-a: I assume this is from lbf :wink: 

```suggestion
```

mava/configs/env/scenario/con-15x15x10a.yaml
```diff
@@ -0,0 +1,6 @@
+# The config of the 15x15-4p-5f scenario with the VectorObserver set as default
```
sash-a:
```suggestion
```

mava/configs/env/scenario/con-5x5x3a.yaml
```diff
@@ -0,0 +1,6 @@
+# The config of the 15x15-4p-5f scenario with the VectorObserver set as default
```
sash-a:
```suggestion
```

mava/configs/env/scenario/con-7x7x5a.yaml
```diff
@@ -0,0 +1,6 @@
+# The config of the 15x15-4p-5f scenario with the VectorObserver set as default
```
sash-a:
```suggestion
```

mava/wrappers/jumanji.py
```diff
@@ -117,3 +118,74 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+
+    def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
+        """Modify the timestep for the Connector environment."""
+
+        def create_agents_view(grid: chex.Array) -> chex.Array:
+            positions = jnp.where(grid % 3 == 2, True, False)   
+            targets = jnp.where((grid % 3 == 0) & (grid != 0), True, False)
+            paths = jnp.where(grid % 3 == 1, True, False)
+            my_position = jnp.where(grid == 2, True, False)
+            my_target = jnp.where(grid == 3, True, False)
```
sash-a: Can we please import these constants from jumanji so it doesn't look like we're using magic numbers here please

mava/wrappers/jumanji.py
```diff
@@ -117,3 +118,74 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+            agents_view = jnp.stack((positions, targets, paths, my_position, my_target), -1)
+            return agents_view
+        
+        def create_global_state(grid: chex.Array) -> chex.Array:
+            positions = jnp.where(grid % 3 == 2, True, False)   
+            targets = jnp.where((grid % 3 == 0) & (grid[0] != 0), True, False)
+            paths = jnp.where(grid % 3 == 1, True, False)
```
sash-a: Same here

mava/wrappers/jumanji.py
```diff
@@ -117,3 +118,74 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+        return timestep.replace(observation=observation)
+    
+    
+    def observation_spec(self) -> specs.Spec[Observation]:
+        """Specification of the observation of the environment."""
+        step_count = specs.BoundedArray(
+            (self._num_agents,),
+            jnp.int32,
+            [0] * self._num_agents,
+            [self.time_limit] * self._num_agents,
```
sash-a: Use jnp functions wherever possible

```suggestion
            jnp.zeros(self._num_agents, dtype=int),
            jnp.repeat(self.time_limit, self._num_agents),
```

mava/wrappers/jumanji.py
```diff
@@ -117,3 +118,74 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+            "step_count",
+        )
+
+        agents_view = specs.BoundedArray(
+            shape=(self._env.num_agents, self._env.grid_size, self._env.grid_size, 5),
+            dtype=jnp.bool,
```
sash-a: We usually just use bool/int/float for types, jax will infer it
```suggestion
            dtype=bool,
```

mava/wrappers/jumanji.py
```diff
@@ -117,3 +118,74 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+            name="agents_view",
+            minimum=False,
+            maximum=True,
+        )
+
+        global_state = specs.BoundedArray(
+            shape=(self._env.num_agents, self._env.grid_size, self._env.grid_size, 3),
+            dtype=jnp.bool,
```
sash-a:
```suggestion
            dtype=bool,
```

mava/wrappers/jumanji.py
```diff
@@ -117,3 +118,74 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+            step_count=jnp.repeat(timestep.observation.step_count, self._num_agents),
+        ) 
+        return timestep.replace(observation=observation)
+    
+    
+    def observation_spec(self) -> specs.Spec[Observation]:
+        """Specification of the observation of the environment."""
+        step_count = specs.BoundedArray(
+            (self._num_agents,),
+            jnp.int32,
```
sash-a:
```suggestion
            int,
```

mava/wrappers/jumanji.py
```diff
@@ -117,3 +118,74 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+            paths = jnp.where(grid % 3 == 1, True, False)
+            global_state = jnp.stack((positions, targets, paths), -1)
+            return global_state
+        
+        observation = ObservationGlobalState(
+            global_state = create_agents_view(timestep.observation.grid),
+            agents_view= create_global_state(timestep.observation.grid),
+            action_mask=timestep.observation.action_mask,
+            step_count=jnp.repeat(timestep.observation.step_count, self._num_agents),
+        )
```
sash-a: 3 things here:

1. You're always returning `ObservationGlobalState` we should only return (and compute) the global state if `add_global_state` is true, you can see this in other wrappers.
2. You're passing `agents_view` to `global_state` and visa versa
3. You're not vmapping `create_agents_view` so you're only passing out a single observation, there should be n_agents of these, so you need to vmap that over agent_ids to correctly rotate the grid

sash-a (changes requested):
Nice, most of the ground work is done, just some things you've got to change in the `modify_timestep` method

mava/wrappers/jumanji.py
SimonDuToit: That code is actually copied from the generic wrapper. I guess we should make another PR for that too?

mava/wrappers/jumanji.py
SimonDuToit: On the last point, create_agents_view works when the observation is batched over the agents, so theres no need to vmap.

SimonDuToit: The rotation is allready done by the Jumanji env. Agent ids are not used.

sash-a: Ah you're right! My bad

mava/wrappers/jumanji.py
sash-a: That's quite related to this PR, so you can squeeze it in here if it's easier :smile:

mava/wrappers/jumanji.py
```diff
@@ -117,3 +118,73 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+            my_target = jnp.where(grid == 3, True, False)
+            agents_view = jnp.stack((positions, targets, paths, my_position, my_target), -1)
+            return agents_view
+
+        def create_global_state(grid: chex.Array) -> chex.Array:
+            positions = jnp.where(grid % 3 == 2, True, False)
+            targets = jnp.where((grid % 3 == 0) & (grid[0] != 0), True, False)
+            paths = jnp.where(grid % 3 == 1, True, False)
+            global_state = jnp.stack((positions, targets, paths), -1)
+            return global_state
```
sash-a: Seeing as the gird is already all agents views can we just use `grid[0]` here? And then tile the output so that we have a single global state?

sash-a: Nevermind this, it's the same

SimonDuToit:
I think Ive implemented all your suggestions @sash-a

mava/wrappers/jumanji.py
```diff
@@ -123,68 +129,85 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
 class ConnectorWrapper(MultiAgentWrapper):
     """Multi-agent wrapper for the MA Connector environment."""
 
-    def __init__(self, env: MaConnector):
+    def __init__(self, env: MaConnector, has_global_state: bool = True):
```
sash-a: This should default to false as we do in other wrappers

```suggestion
    def __init__(self, env: MaConnector, has_global_state: bool = False):
```

mava/wrappers/jumanji.py
```diff
@@ -123,68 +129,85 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
 
         def create_agents_view(grid: chex.Array) -> chex.Array:
-            positions = jnp.where(grid % 3 == 2, True, False)
-            targets = jnp.where((grid % 3 == 0) & (grid != 0), True, False)
-            paths = jnp.where(grid % 3 == 1, True, False)
-            my_position = jnp.where(grid == 2, True, False)
-            my_target = jnp.where(grid == 3, True, False)
+            positions = jnp.where(grid % TARGET == POSITION, True, False)
+            targets = jnp.where((grid % TARGET == 0) & (grid != EMPTY), True, False)
+            paths = jnp.where(grid % TARGET == PATH, True, False)
```
sash-a, line 148: Sorry for the confusion, but `grid % TARGET` is a bit confusing to me and I see there isn't a better constant in connector for this, but technically `TARGET` isn't gaurenteed to always be the largest item e.g if we added another item to connector in the future, so I would put `n_grid_items = 3` at the top of `modify_timestep` and always ` % n_grid_items` that should make it super clear what you're doing :+1:

sash-a (changes requested):
Two super minor things otherwise I'm happy :+1:

mava/wrappers/jumanji.py
SimonDuToit: Its a bit tricky since in that case TARGET wouldn't appear at all. Ive added a comment that clarifies things.

sash-a (dismissed):
LGTM :ship:

mava/configs/env/connector.yaml
```diff
@@ -0,0 +1,9 @@
+# ---Environment Configs---
+defaults:
+  - _self_
+  - scenario: con-10x10x5a
+
+env_name: MaConnector-v2
```
RuanJohn: Since this is only used for logging can this just be `MaConnector` please? And can you add a comment mentioning this in this file like we have in other env configs as well please?

mava/configs/env/connector.yaml
```diff
@@ -0,0 +1,9 @@
...
+
```
RuanJohn, line 12: Can you also please add an `eval_metric` here?

mava/configs/env/connector.yaml
```diff
@@ -0,0 +1,9 @@
+# ---Environment Configs---
+defaults:
+  - _self_
+  - scenario: con-10x10x5a
```
RuanJohn: Can we please have a comment here with all possible scenarios configs like in other envs?

mava/configs/env/scenario/con-10x10x5a.yaml
```diff
@@ -0,0 +1,5 @@
+task_name: con-10x10x5a
+
```
RuanJohn, line 4: To all the scenario configs can we please add a `name` attribute which is the env name used for initialisation and then use that to create the envs please? Similar to other envs. 
Similar for all other scenario configs as well please.

mava/wrappers/jumanji.py
```diff
@@ -117,3 +124,94 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+    ) -> TimeStep[Union[Observation, ObservationGlobalState]]:
+        """Modify the timestep for the Connector environment."""
+
+        # TARGET = 3 = The number of different types of items on the grid.
+        def create_agents_view(grid: chex.Array) -> chex.Array:
+            positions = jnp.where(grid % TARGET == POSITION, True, False)
+            targets = jnp.where((grid % TARGET == 0) & (grid != EMPTY), True, False)
+            paths = jnp.where(grid % TARGET == PATH, True, False)
+            my_position = jnp.where(grid == POSITION, True, False)
+            my_target = jnp.where(grid == TARGET, True, False)
```
RuanJohn: The naming here is a bit confusing? Is this the position and target of a current agent?

mava/wrappers/jumanji.py
```diff
@@ -117,3 +124,94 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+            return agents_view
+
+        def create_global_state(grid: chex.Array) -> chex.Array:
+            positions = jnp.where(grid % TARGET == POSITION, True, False)
+            targets = jnp.where((grid % TARGET == 0) & (grid != EMPTY), True, False)
+            paths = jnp.where(grid % TARGET == PATH, True, False)
+            global_state = jnp.stack((positions, targets, paths), -1)
+            return global_state
+
+        obs_data = {
```
RuanJohn, line 163: Really like this!

RuanJohn (changes requested):
Really nice PR @SimonDuToit :fire: 
Just a few style requests here and there.

mava/wrappers/jumanji.py
SimonDuToit: Yes, this is the per-agent position and target.

RuanJohn: Can we please name this in a way where this is easier to infer? Maybe something like `per_agent_position`, `per_agent_target`?

SimonDuToit: Done, along with all the other changes.

mava/configs/env/connector.yaml
```diff
@@ -0,0 +1,14 @@
...
+defaults:
+  - _self_
+  - scenario: con-10x10x5a # [con-5x5x3a, con-7x7x5a, con-10x10x5a, con-15x15x10a]
+# Further environment config details in "con-10x10x5a" file.
+
+env_name: MaConnector # Used for logging purposes.
+
+# Defines the metric that will be used to evaluate the performance of the agent.
+# This metric is returned at the end of an experiment and can be used for hyperparameter tuning.
+eval_metric: episode_return
```
RuanJohn, line 11: Is there a reason we aren't returning the ratio connected yet?

mava/wrappers/jumanji.py
```diff
@@ -117,3 +124,96 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+
+class ConnectorWrapper(MultiAgentWrapper):
+    """Multi-agent wrapper for the MA Connector environment.
+
+    Do not use the AgentID wrapper with this env, it has implicit agent IDs.
+    """
+
+    def __init__(self, env: MaConnector, has_global_state: bool = False):
+        super().__init__(env)
+        self.has_global_state = has_global_state
```
RuanJohn, line 137: Will MaConnector have a global state?

mava/wrappers/jumanji.py
```diff
@@ -117,3 +124,96 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
+
+    def modify_timestep(
+        self, timestep: TimeStep
+    ) -> TimeStep[Union[Observation, ObservationGlobalState]]:
+        """Modify the timestep for the Connector environment."""
+
+        # TARGET = 3 = The number of different types of items on the grid.
+        def create_agents_view(grid: chex.Array) -> chex.Array:
+            positions = jnp.where(grid % TARGET == POSITION, True, False)
```
RuanJohn, line 146: Bit of a naive question here, but are we allowing all agents to always see where all goals for all other agents are? Just to double check that is is intended if it is the case @sash-a?

RuanJohn (commented):
Thank you @SimonDuToit, a few more minor questions from my side.

mava/configs/env/connector.yaml
sash-a: Ye it's a bit tricky to do this, so @SimonDuToit and I agreed the changes can come in a future PR just so this can be merged easily :smile:

mava/wrappers/jumanji.py
sash-a: Yup, this is how connector works :+1:

mava/wrappers/jumanji.py
sash-a: Ye, it's just the obs of agent 0

SimonDuToit: Yes. The global state is the agent observation without the per_agent parts. So just a full view of the grid.

mava/configs/env/connector.yaml
SimonDuToit: Yes. With how Mava currently works this argument doesnt actually do much... Im working on a more general solution.

RuanJohn: Sounds good. Thank you :)

mava/utils/make_env.py
```diff
@@ -43,6 +47,7 @@
 _jumanji_registry = {
     "RobotWarehouse-v0": {"generator": RwareRandomGenerator, "wrapper": RwareWrapper},
     "LevelBasedForaging-v0": {"generator": LbfRandomGenerator, "wrapper": LbfWrapper},
+    "MaConnector-v2": {"generator": ConnectorRandomGenerator, "wrapper": ConnectorWrapper},
```
sash-a, line 50: Should this not be the random walk generator?

sash-a (changes requested):
Small question about the generator

mava/wrappers/jumanji.py
RuanJohn: Spoke offline. This makes sense to me for now.

mava/utils/make_env.py
SimonDuToit: That is the random walk generator, its just renamed in the import

sash-a (approved):
:fire:

RuanJohn (approved):
Thank you @SimonDuToit :fire: :robot:

PR #1037, opened 2024-02-22
Feat: Support Continuous environments with unvmapped Nets

### LINKED ISSUE #990: [FEATURE] Continuous Actions PPO
### Feature
Enhance PPO systems by introducing support for continuous actions

### Proposal
Two implementation approaches are considered:
1. Create a `select_actions` file containing both discrete and continuous action selections, and integrate this into the single-file PPO systems.
2. Split the existing ff_ippo file into two separate files: one for discrete actions and another for continuous actions.

* [x] Implement continuous actions network and function
* [x] Make all the different PPO architecture support the continous actions (rec/ff) (IL/CTDE)
* [x] Benchmarking on MaBrax or a different jax environment that support continoues actions

### Links/references
* Current branch `feat/continuous_actions`.

WiemKhlifi:
## What?
Full support for continuous actions in all Mava PPO systems. 

## Why?
Add support for environments with continuous actions.

## How?
* ContinuousActionHead added
* Support for TensorFlow probability instead of distrax.
* Access `num_actions `and `num_agents` directly from environments (whether continuous or discrete).

## Extra
This is the non-vamped version of the PPO networks.


mava/evaluator.py
```diff
@@ -178,7 +178,9 @@ def _env_step(eval_state: RNNEvalState) -> RNNEvalState:
             hstate, pi = apply_fn(params, hstate, ac_in)
 
             if config.arch.evaluation_greedy:
-                action = pi.mode()
+                action = (
+                    pi._mode()
+                )  # Not valid for ContinuousActionHead (issue with: pi.distribution.mode())
```
sash-a: nit:
```suggestion
                # Not valid for ContinuousActionHead (issue with: pi.distribution.mode())
                action = pi._mode()
```

mava/networks.py
```diff
@@ -108,7 +110,54 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
             jnp.finfo(jnp.float32).min,
         )
 
-        return distrax.Categorical(logits=masked_logits)
+        return tfd.TransformedDistribution(
+            distribution=tfd.Categorical(logits=masked_logits), bijector=tfb.Identity()
+        )
```
sash-a:
```suggestion
        # we transform this distribution with the `Identity` transform to allow 
        # us to call `pi.distribution.entropy()` and keep the API the same as the 
        # continuous distribution.
        return tfd.TransformedDistribution(
            distribution=tfd.Categorical(logits=masked_logits), bijector=tfb.Identity()
        )
```

mava/networks.py
```diff
@@ -108,7 +110,54 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
+
+
+class ContinuousActionHead(nn.Module):
+    """Continuous Action Head"""
+
+    action_dim: int
+    minimum: float = -1
+    maximum: float = 1
```
sash-a: I wonder if it would be cleaner to just pass the env in here and then get the minimum and maximum from there.

mava/networks.py
```diff
@@ -108,7 +110,54 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
+        self.actor_log_std = self.param("log_std", nn.initializers.zeros, (self.action_dim,))
+        scale = 0.5 * (self.maximum - self.minimum)
+        self.bijector = tfb.Chain(
+            [tfb.Shift(self.minimum), tfb.Scale(scale=scale), tfb.Shift(1), tfb.Tanh()]
+        )
+
+    def _sigma_activation(
+        self, sigma: chex.Array, sigma_min: float = -4, sigma_max: float = 0
+    ) -> chex.Array:
+        # Not sure if this is the right way also for sigma_min and max?
```
sash-a:
```suggestion
        """Don't allow sigma to grow too large for numerical stability."""
```

mava/networks.py
```diff
@@ -108,7 +110,54 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
+    def setup(self) -> None:
+        self.actor_mean = nn.Dense(self.action_dim, kernel_init=orthogonal(0.01))
+        self.actor_log_std = self.param("log_std", nn.initializers.zeros, (self.action_dim,))
+        scale = 0.5 * (self.maximum - self.minimum)
+        self.bijector = tfb.Chain(
+            [tfb.Shift(self.minimum), tfb.Scale(scale=scale), tfb.Shift(1), tfb.Tanh()]
+        )
+
+    def _sigma_activation(
+        self, sigma: chex.Array, sigma_min: float = -4, sigma_max: float = 0
```
sash-a: In SAC I use `[-5, 2]`, but I know rlax uses these values, I don't really know which is better without running experiments, but I suspect it makes little difference.

But also can we rather make these values you can pass into the network and default them at the top :smile:

mava/networks.py
```diff
@@ -108,7 +110,54 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
+    ) -> distrax.MultivariateNormalDiag:
+        """Action selection for continuous action space environments.
+
+        Args:
+            obs_embedding: Observation embedding from network torso.
+            observation: Observation object containing `agents_view`, `action_mask` and
+                `step_count`.
+
+        Returns:
+            A distrax.MultivariateNormalDiag distribution.
```
sash-a:
```suggestion
            A transformed distrax.MultivariateNormalDiag distribution.
```

mava/networks.py
```diff
@@ -108,7 +110,54 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
+        """
+
+        mean = self.actor_mean(obs_embedding)
+        return tfd.TransformedDistribution(
+            distribution=tfd.MultivariateNormalDiag(
+                loc=mean, scale_diag=self._sigma_activation(self.actor_log_std)
+            ),
+            bijector=self.bijector,
+        )
```
sash-a: nit:
```suggestion
        mean = self.actor_mean(obs_embedding)
        std = self._sigma_activation(self.actor_log_std)
        
        return tfd.TransformedDistribution(
            distribution=tfd.MultivariateNormalDiag(loc=mean, scale_diag=std),
            bijector=self.bijector,
        )
```

mava/systems/ppo/ff_ippo.py
```diff
@@ -81,6 +81,8 @@ def _env_step(learner_state: LearnerState, _: Any) -> Tuple[LearnerState, PPOTra
             key, policy_key = jax.random.split(key)
             actor_policy = actor_apply_fn(params.actor_params, last_timestep.observation)
             value = critic_apply_fn(params.critic_params, last_timestep.observation)
+            # sample,log_prob= actor_policy._sample_and_log_prob(
+            # seed=seed, sample_shape=(num_actions,))
```
sash-a: Unfortunate that tfp doesn't have `sample_and_log_prob`

```suggestion
```

mava/wrappers/jaxmarl.py
```diff
@@ -339,10 +345,13 @@ def state_size(self) -> chex.Array:
...
-    def n_actions(self) -> chex.Array:
+    def num_agents(self) -> chex.Array:
+        return self._env.num_agents
+
+    @cached_property
+    def num_actions(self) -> chex.Array:
         "Get the number of actions for each agent."
-        single_agent_action_space = self._env.action_space(self.agents[0])
-        return single_agent_action_space.n
+        return self.action_spec().shape[0]
```
sash-a: Given that SMAX is discrete I don't think this is correct :thinking:

mava/wrappers/jumanji.py
```diff
@@ -49,10 +50,10 @@ def step(self, state: State, action: chex.Array) -> Tuple[State, TimeStep]:
     def observation_spec(self) -> specs.Spec[Observation]:
         """Specification of the observation of the environment."""
         step_count = specs.BoundedArray(
-            (self._num_agents,),
+            (self.num_agents,),
             jnp.int32,
-            [0] * self._num_agents,
-            [self.time_limit] * self._num_agents,
+            [0] * self.num_agents,
+            [self.time_limit] * self.num_agents,
```
sash-a: While we're here:

```suggestion
            jnp.zeros(self.num_agents, dtype=int),
            jnp.repeat(self.time_limit, self.num_agents),
```

mava/systems/ppo/ff_mappo.py
```diff
@@ -333,8 +333,8 @@ def learner_setup(
     n_devices = len(jax.devices())
 
     # Get number of actions and agents.
-    num_actions = int(env.action_spec().num_values[0])
-    num_agents = env.action_spec().shape[0]
+    num_actions = env.num_actions
```
sash-a: Can we call this `act_dim` rather? Both in the env and here, as I think this works better for continuous/discrete systems. `num_actions` only really explains discrete

sash-a (changes requested):
Thanks Wiem this looks great! Can we confirm that performance is still good?

mava/networks.py
WiemKhlifi: Yeah seems cleaner and we no longer need to pass actions_dim as well 🤔

mava/networks.py
WiemKhlifi: Yeah maybe add them to hyperparams sweep and see what we can get 🤔 

Here I'm afraid we'll have different attributes for both discrete and continuous action heads, isn't the best way to add them all into a config (cfg.network) and access everything related to networks there (even the action_dim and min, max)??

mava/systems/ppo/ff_ippo.py
WiemKhlifi: Yeah we'll get some typing issues 😿

mava/wrappers/jaxmarl.py
WiemKhlifi: Yes thanks, I wanted to move this method in the jaxmarl wrapper but then changed my mind and forgot to revert this 😅

mava/networks.py
sash-a: Ye so for sigma min/max it will just work with how hydra creates the networks, but I we definitely shouldn't pass in action_dim, act_min/max as these are runtime variables and we don't want to set them in the config. I think if we pass in env to all networks (discrete and continuous) then we should be able to get all the info we need

mava/networks.py
```diff
@@ -81,10 +84,10 @@ def __call__(self, observation: chex.Array) -> chex.Array:
 class DiscreteActionHead(nn.Module):
     """Discrete Action Head"""
 
-    action_dim: int
+    env: Environment
 
     @nn.compact
-    def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distrax.Categorical:
+    def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.Categorical:
```
sash-a: I think it now returns this

```suggestion
    def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.Transformed:
```

mava/networks.py
```diff
@@ -93,22 +96,60 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
+class ContinuousActionHead(nn.Module):
+    """ContinuousActionHead using a transformed Normal distribution."""
+
+    env: Environment
+    min_scale: float = 1e-3
+
+    def setup(self) -> None:
+        self.mean = nn.Dense(self.env.action_dim, kernel_init=orthogonal(0.01))
+        self.log_std = self.param("log_std", nn.initializers.zeros, (self.env.action_dim,))
+        min_action, max_action = self.env.min_max_action
```
sash-a: What is `min_max_action`?

I think rather `self.env.action_spec().minimum`

mava/networks.py
```diff
@@ -93,22 +96,60 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
         masked_logits = jnp.where(
             observation.action_mask,
             actor_logits,
             jnp.finfo(jnp.float32).min,
         )
 
-        return distrax.Categorical(logits=masked_logits)
+        #  We transform this distribution with the `Identity()` transformation to
+        # keep the API identical to the ContinuousActionHead.
+        return IdentityTransformation(distribution=tfd.Categorical(logits=masked_logits))
```
sash-a, line 116: The new continuous distribution has an entropy function 

```suggestion
        tfd.Categorical(logits=masked_logits)
```

mava/wrappers/jaxmarl.py
```diff
@@ -280,6 +281,11 @@ def observation_spec(self) -> specs.Spec:
     def action_spec(self) -> specs.Spec:
         return jaxmarl_space_to_jumanji_spec(merge_space(self._env.action_spaces))
 
+    @cached_property
+    def min_max_action(self) -> Tuple[chex.Array, chex.Array]:
+        """Returns the minimum and maximum values from the action space"""
+        return self.action_spec().minimum, self.action_spec().maximum
+
```
sash-a: I think it would be better to leave this out and just call this in the network init when we need it, otherwise it adds another requirement to our wrappers and I think it's better to keep our wrappers as lightweight as possible :smile: 

```suggestion
```

mava/wrappers/jumanji.py
```diff
@@ -56,14 +58,26 @@ def step(self, state: State, action: chex.Array) -> Tuple[State, TimeStep]:
...
+            jnp.repeat(self.time_limit, self.num_agents),
             "step_count",
         )
         return self._env.observation_spec().replace(step_count=step_count)
 
+    @cached_property
+    @abstractmethod
+    def min_max_action(self) -> Tuple[chex.Array, chex.Array]:
+        """Returns the minimum and maximum values from the action space"""
+        ...
```
sash-a: Same reason I think we should remove this

```suggestion
```

sash-a (changes requested):
Looking really good! Just a couple things I think we should remove, otherwise I'm happy :fire:

mava/distributions.py
```diff
@@ -0,0 +1,110 @@
...
+          threshold: Clipping value for the action when computing the log_prob.
+          validate_args: Whether to validate input with respect to distribution parameters.
+        """
+        if bijector is None:
+            bijector = tfb.Tanh()
+        super().__init__(distribution=distribution, bijector=bijector, validate_args=validate_args)
+        # Computes the log of the average probability distribution outside the
+        # clipping range, i.e. on the interval [-inf, -atanh(threshold)] for
+        # log_prob_left and [atanh(threshold), inf] for log_prob_right.
+        self._threshold = threshold
```
sash-a, line 52: Looking at the clipping I *think* that this should be our min/max of actions, but I'm not entirely sure.

RuanJohn: I think it just keeps you from being allowed to compute arctanh of 1, which will be infinity.

mava/networks.py
```diff
@@ -93,22 +96,60 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
+
+
+class ContinuousActionHead(nn.Module):
+    """ContinuousActionHead using a transformed Normal distribution."""
+
+    env: Environment
```
RuanJohn: In general, I think we shouldn't be passing envs in to network heads. We should rather just get the information we need beforehand and then pass that in.

mava/systems/ppo/ff_ippo.py
```diff
@@ -145,14 +146,15 @@ def _update_minibatch(train_state: Tuple, batch_info: Tuple) -> Tuple:
...
-                params, opt_states = train_state
+                params, opt_states, entropy_key = train_state
                 traj_batch, advantages, targets = batch_info
 
                 def _actor_loss_fn(
                     actor_params: FrozenDict,
                     actor_opt_state: OptState,
                     traj_batch: PPOTransition,
                     gae: chex.Array,
+                    entropy_key: chex.PRNGKey,
```
RuanJohn: Can we rather just call this `key`.

mava/networks.py
sash-a: Maybe rather action_spec then, we should be able to get everything we need from this for both continuous and discrete

mava/networks.py
WiemKhlifi: Yeah but doesn't require a seed attribute compared to what we do in Tanh distribution( used in dist.sample) 😞

mava/networks.py
WiemKhlifi: Yeah then we make default values for both networks and mention in the discrete one that this is needed only for continuous case

mava/networks.py
```diff
@@ -131,10 +130,11 @@ def setup(self) -> None:
         self.log_std = self.param("log_std", nn.initializers.zeros, (self.action_dim,))
 
         minimum, maximum = self.action_spec.minimum, self.action_spec.maximum
-        scale = 0.5 * (maximum - minimum)
-        self.bijector = tfb.Chain(
-            [tfb.Shift(minimum), tfb.Scale(scale=scale), tfb.Shift(1), tfb.Tanh()]
+        space_act_checker = lambda min, max: checkify.check(
+            min + max == 0,
+            "This network only deals with the case where actions lie in the interval [-1, 1]",
```
sash-a: Couple things, I wouldn't use checkify here, just a normal assert because we only want this check to be run once at compile time and asserts don't run once a function has been jitted. Also I think put the assert in `setup` rather. And also I wouldn't check `min + max == 0` because we could a `[-5,5]` action space and that would pass the assert, so rather just `assert (min == -1) & (max == 1)`

sash-a (changes requested):
Just a small request with the assert, otherwise I'm happy and we can finally merge this!

mava/networks.py
WiemKhlifi: I was going to push the min-max check correction (`max - min == 2`) thanks 😅 As for the assert it didn't work in the `setup` again due to `jax.errors.TracerBoolConversionError `and I thought maybe we add this check outside in the system files but we'll have to check which head we're using and do it in all files 😢

mava/distributions.py
```diff
@@ -0,0 +1,109 @@
...
+from typing import Any, Optional
+
+import chex
+import jax.numpy as jnp
+import tensorflow_probability.substrates.jax.bijectors as tfb
+import tensorflow_probability.substrates.jax.distributions as tfd
+
+
+class TanhTransformedDistribution(tfd.TransformedDistribution):
+    """A distribution transformed using the `tanh` function."""
```
RuanJohn: Maybe add a note on where this is adapted from.

RuanJohn: Ie for the original, people can see the link in acme.

WiemKhlifi: How about now?

sash-a (dismissed):
Thanks @WiemKhlifi for this huge effort!! :1st_place_medal:

mava/networks.py
```diff
@@ -108,7 +114,48 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
+    def setup(self) -> None:
+        self.mean = nn.Dense(self.action_dim, kernel_init=orthogonal(0.01))
+        self.log_std = self.param("log_std", nn.initializers.zeros, (self.action_dim,))
+
+        minimum, maximum = self.action_spec.minimum, self.action_spec.maximum
+        space_act_checker = lambda min, max: checkify.check(
+            (max == 1) & (min == -1),
+            "This network only handles the case where actions lie in the interval [-1, 1].",
+        )
+        checkify.checkify(space_act_checker)(minimum, maximum)
```
RuanJohn: We should not be using anything from jax.experimental please.

mava/systems/ppo/ff_ippo.py
```diff
@@ -145,14 +146,15 @@ def _update_minibatch(train_state: Tuple, batch_info: Tuple) -> Tuple:
                 """Update the network for a single minibatch."""
 
                 # UNPACK TRAIN STATE AND BATCH INFO
-                params, opt_states = train_state
+                params, opt_states, entropy_key = train_state
```
RuanJohn: We need to split and return a new key at each minibatch please. It looks like we are just splitting once at the epoch level and using the same seed for all minibatches.

RuanJohn: Same applies to all systems :pray:

mava/systems/ppo/rec_mappo.py
```diff
@@ -332,21 +334,12 @@ def _critic_loss_fn(
...
-                advantages,
-                targets,
-                key,
-            ) = update_state
-            key, shuffle_key = jax.random.split(key)
+            params, opt_states, init_hstates, traj_batch, advantages, targets, key = update_state
+            key, shuffle_key, entropy_key = jax.random.split(key, 3)
 
             # SHUFFLE MINIBATCHES
-            permutation = jax.random.permutation(shuffle_key, config.arch.num_envs)
```
RuanJohn, line 349: Good spot for removing this! :100:

mava/networks.py
```diff
@@ -16,13 +16,16 @@
...
 
 import chex
-import distrax
 import jax
 import jax.numpy as jnp
 import numpy as np
+import tensorflow_probability.substrates.jax.distributions as tfd
 from flax import linen as nn
 from flax.linen.initializers import orthogonal
+from jax.experimental import checkify
```
RuanJohn: Similiar to the other comment. But we should please not be using anything from `jax.experimental`

RuanJohn (changes requested):
Thanks @WiemKhlifi for all this work! It looks mostly good, just some minor changes needed.

mava/systems/ppo/ff_ippo.py
WiemKhlifi: I missed it thank you 🙏

mava/networks.py
```diff
@@ -120,23 +117,12 @@ def __call__(
 
 
 class ContinuousActionHead(nn.Module):
-    """ContinuousActionHead using a transformed Normal distribution."""
+    """ContinuousActionHead using a transformed Normal distribution.
+    Note: This network only handles the case where actions lie in the interval [-1, 1]."""
```
sash-a:
```suggestion
    """ContinuousActionHead using a transformed Normal distribution.
    
    Note: This network only handles the case where actions lie in the interval [-1, 1].
    """
```

sash-a (changes requested):
Couple small things, sent offline :smile:

mava/distributions.py
```diff
@@ -0,0 +1,110 @@
...
+
+import chex
+import jax.numpy as jnp
+import tensorflow_probability.substrates.jax.bijectors as tfb
+import tensorflow_probability.substrates.jax.distributions as tfd
+
+
+class TanhTransformedDistribution(tfd.TransformedDistribution):
+    """A distribution transformed using the `tanh` function.
+    This transformation was adapted from acme repo: http://tinyurl.com/2x5xea57"""
```
RuanJohn:
```suggestion
    """
    A distribution transformed using the `tanh` function.
    
    This transformation was adapted acme's implementation.
    For details, please see: http://tinyurl.com/2x5xea57
    """
```

mava/networks.py
```diff
@@ -108,7 +111,41 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> distr
...
 
-        return distrax.Categorical(logits=masked_logits)
+        #  We transform this distribution with the `Identity()` transformation to
+        # keep the API identical to the ContinuousActionHead.
+        return IdentityTransformation(distribution=tfd.Categorical(logits=masked_logits))
+
+
+class ContinuousActionHead(nn.Module):
+    """ContinuousActionHead using a transformed Normal distribution.
+    Note: This network only handles the case where actions lie in the interval [-1, 1]."""
```
sash-a:
```suggestion
    """ContinuousActionHead using a transformed Normal distribution.
    
    Note: This network only handles the case where actions lie in the interval [-1, 1].
    """
```

sash-a (changes requested):
Just a docstring formatting

mava/networks.py
WiemKhlifi: Sorry I think was changed with the revert 😅

sash-a (approved):
:1st_place_medal:

RuanJohn (approved):
:fire:

PR #1038, opened 2024-02-22
Feat/remove agent vmapping

RuanJohn:
## What?
Remove `vmap`ing manually over agents and instead rely on Flax's autobatching since parameters are shared. 
## Why?
Since we return distributions which can be sampled from our policies, we ran into issues with both [Distrax](https://github.com/google-deepmind/distrax?tab=readme-ov-file#a-note-about-vmap-and-pmap) and [Tensorflow probability](https://github.com/tensorflow/probability/issues/1271) where the creation of distributions for continuous action space policies cannot be `vmap`ed. 
To keep things consistent across system we opted to remove `vmap`ing over agents.


mava/evaluator.py
```diff
@@ -250,6 +248,9 @@ def evaluator_fn(
             dtype=bool,
         )
 
+        # Adding an extra batch dim for the vmapped eval functions.
+        init_hstate = jnp.expand_dims(init_hstate, axis=1)
```
sash-a: nit:
```suggestion
        init_hstate = init_hstate[:, jnp.newaxis, ...]
```

sash-a (commented):
One minor thing, looks good though!

sash-a (approved):
:shipit:

WiemKhlifi (approved):
LGTM 🎆

Issue #1033, opened 2024-02-20
[FEATURE]: Reduce config files


WiemKhlifi:
### Tthe purpose of the feature:
Integrate the systems' logger default into a single yaml file and move the **common** training controlling flags from systems files into the arch yaml files (e.g. `seed` and `total_timesteps`). This helps avoid setting things in each system in case we want to run a sweep for example and avoid adding a logger_default for each system.

### Describe the solution you'd like
Loggers merged in one file and  setting the common training flags outside each system 😄 

### How do we know when the implementation of this feature is complete?
Checklist:
- [x] Merge the logger files into one yaml file and add the `system_name` in the logger.py code by using the `tags.append(system.system_name)`.
- [ ] Add the common items related to training all systems in the arch files such as seed, total_timesteps, and num_updates.

### Additional context
Check this PR #987 for some inspiration (An example was started in this PR but for now, we keep the systems' default separate) 🙌 

🌟 Additional request: Merge system defaults for the same system type (such PPOs) and find the best way to select the system config without changing the default.

PR #1039, opened 2024-02-23
Refactor the configs to mirror the structure of the python files

liamclarkza:
## What?
Small refactor that restructures the system config files to mirror the structure used for the python modules.


EdanToledo:
I know that I'm just a community reviewer ;) but I'd like to just ask if one should reorganise the default config files as well? The problem persists that even if system configs are organised, as the algorithm variants increase the number of default configs will be very large as well.

WiemKhlifi:
Hey @EdanToledo, you're right! We aim to fix this soon before the config folder explodes, please see issue #1033 😄

RuanJohn (approved):
Thanks @liamclarkza

WiemKhlifi (approved):
Thanks @liamclarkza 🙏

PR #1041, opened 2024-02-26
Fix: dtype of agentid wrapper.

### LINKED ISSUE #1032: [BUG]: Flasbax bug on LBF (type issue)
### Describe the bug
In the AgentID wrapper, the new_agents_view type is not enforced to be consistent, so for LBF with flashbax a dtype error emerges.

### To Reproduce
1. Run LBF with flashbax.

### Expected behavior
Expected the observation to be added easily to the buffer.

### Context (Environment)
 - Updated jumanji

### Additional context
This is somewhat an exercise in opening an issue!

### Possible Solution
Will make a PR soon! Basically a cast to the agents_view type.

lbeyers:
## What?
Fixes #1032 :)
## Why?
Dtype was not enforced to be consistent, caused errors.
## How?
Enforced dtype in concatenation operation.

### DIFF
--- a/mava/wrappers/observation.py
+++ b/mava/wrappers/observation.py
@@ -38,7 +38,11 @@ class AgentIDWrapper(Wrapper):
         """Adds agent IDs to the observation."""
         obs = timestep.observation
         agent_ids = jnp.eye(num_agents)
-        agents_view = jnp.concatenate([agent_ids, obs.agents_view], axis=-1)
+        agents_view = jnp.concatenate(
+            [agent_ids, obs.agents_view],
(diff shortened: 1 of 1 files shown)


callumtilbury (approved):
LGTM 🚀 

@WiemKhlifi we haven't tested on RWARE or Matrax, but I am assuming it will be fine? This fix is sufficiently general. Let me know if you are happy!

WiemKhlifi (approved):
Thanks @lbeyers I forgot to add this detail🙏 
@callumtilbury works fine for the rest of envs 🙌

PR #1042, opened 2024-02-27
fix: sometimes logging 0 episode return

sash-a:
## What?
We were sometimes logging episode returns when our rollout length was too short to hit the end of an episode. Made sure we don't log these episodes.


liamclarkza (dismissed):
Looks good

RuanJohn (changes requested):
Looks good to me! Could you just please add this to the advanced usage systems as well?

liamclarkza (approved):
Looks good 👍

RuanJohn (approved):
LGTM

PR #1043, opened 2024-02-27
Feat : MultiCVRP wrapper

Louay-Ben-nessir:
## What?
adds a wrapper for multicvrp.
## Why?
To make mulctivcrp mava compatible.
## How?
This wrapper takes in the output from the multicvrp environment and converts it into a Mava suitable input.


mava/utils/make_env.py
```diff
@@ -61,7 +66,10 @@ def add_optional_wrappers(
 ) -> Environment:
     # Add the global state to observation.
     if add_global_state:
-        env = GlobalStateWrapper(env)
+        if hasattr(env, "has_global_state"):
+            env.has_global_state = True
+        else:
+            env = GlobalStateWrapper(env)
```
sash-a: I assume this is will change once we merge the global state fix PR?

mava/wrappers/jumanji.py
```diff
@@ -217,3 +220,128 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
         return spec
+
+
+class MultiCVRPWrapper(Wrapper):
+    """Wrapper for MultiCVRP environment."""
+
+    def __init__(self, env: MultiCVRP, has_global_state: bool = False):
+        self.num_agents = env._num_vehicles
+        self._env = env
+        self.has_global_state = has_global_state
```
sash-a: Missing the call to `super`

```suggestion
    def __init__(self, env: MultiCVRP, has_global_state: bool = False):
        super().__init__(env)
        self._env: MultiCVRP
        
        self.num_agents = env._num_vehicles
        self.has_global_state = has_global_state
```

mava/wrappers/jumanji.py
```diff
@@ -217,3 +220,128 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
             )
 
         return spec
+
+
+class MultiCVRPWrapper(Wrapper):
```
sash-a: Inherit from the `MultiAgentWrapper`

```suggestion
class MultiCVRPWrapper(MultiAgentWrapper):
```

mava/wrappers/jumanji.py
```diff
@@ -217,3 +220,128 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+
+    def reset(self, key: chex.PRNGKey) -> Tuple[State, TimeStep]:
+        state, timestep = self._env.reset(key)
+        timestep = self.modify_timestep(timestep, state.step_count)
+        return state, timestep
+
+    def step(self, state: State, action: chex.Array) -> Tuple[State, TimeStep]:
+        state, timestep = self._env.step(state, action)
+        timestep = self.modify_timestep(timestep, state.step_count)
+        return state, timestep
```
sash-a: No need for this as the `MultiAgentWrapper` will take care of it for you

```suggestion
```

mava/wrappers/jumanji.py
```diff
@@ -217,3 +220,128 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+
+    def modify_timestep(self, timestep: TimeStep, step_count: chex.Array) -> TimeStep[Observation]:
```
sash-a: Can we use the same method as in the base class, just remove the step_count

mava/wrappers/jumanji.py
```diff
@@ -217,3 +220,128 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+            observation = Observation(**obs_data)
+
+        reward = jnp.repeat(timestep.reward, (self.num_agents))
+        discount = jnp.repeat(timestep.discount, (self.num_agents))
+        timestep = timestep.replace(observation=observation, reward=reward, discount=discount)
+        return timestep
+
+    def _format_observation(
+        self, observation: MultiCVRPState
+    ) -> Tuple[chex.Array, Union[None, chex.Array]]:
```
sash-a: Also the observation isn't a `MultiCVRPState` it's an observation from MultiCvrp, can you import that observation, something like: `from jumanji.routing.MultiCVRP import Observation as MultiCvrpObservation` and use that here instead

```suggestion
    def _flatten_observation(
        self, observation: MultiCVRPState
    ) -> Tuple[chex.Array, Union[None, chex.Array]]:
```

mava/wrappers/jumanji.py
```diff
@@ -217,3 +220,128 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+        """
+        Formats the observation dictionary from the environment into a format suitable for mava.
+
+        Args:
+            observation (Dict[chex.Array]): The raw observation dict provided by jumanji
```
sash-a: Observation is a named tuple
```suggestion
        Concatenates all observation fields into a single array.

        Args:
            observation (MultiCvrpObservation): The raw observation NamedTuple provided by jumanji.
```

mava/wrappers/jumanji.py
```diff
@@ -217,3 +220,128 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+
+        Returns:
+            observations (chex.Array): Concatenated individual observations for each agent,
+                                      shaped (num_agents, vehicle_info + customer_info).
+            global_observation (Union[None, chex.Array]): Concatenated global observation
+                                                          shaped (num_agents, global_info)
+                                                          if has_global_state = True,
+                                                          None otherwise.
```
sash-a: nit:
```suggestion
            observations (chex.Array): Concatenated individual observations for each agent,
                shaped (num_agents, vehicle_info + customer_info).
            global_observation (Union[None, chex.Array]): Concatenated global observation
                shaped (num_agents, global_info) if has_global_state = True, None otherwise.
```

mava/wrappers/jumanji.py
```diff
@@ -217,3 +220,128 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+        customers_info = jnp.column_stack(customers_info).ravel()
+        vehicles_info = jnp.column_stack(vehicles_info)
+
+        if self.has_global_state:
+            global_observation = jnp.concat((customers_info, vehicles_info.ravel()))
+            global_observation = jnp.tile(global_observation, (self.num_agents, 1))
+
+        customers_info = jnp.tile(customers_info, (self.num_agents, 1))
+        observations = jnp.column_stack((vehicles_info, customers_info))
+        return observations, global_observation
```
sash-a: Can we get some shape comments throughout this method e.g:

```
# C: number of customers
# V: number of vehicles
# N: number of nodes
# W: number of windows
# Co: coeffs

# Tuple[(C,N), (C,W), (C,Co)]
customers_info, _ = tree_util.tree_flatten(
            (observation.nodes, observation.windows, observation.coeffs)
        )
```

sash-a (changes requested):
Good start! Big thing, make sure you use the `MultiAgentWrapper` Base class to avoid duplicating too much code

mava/utils/make_env.py
WiemKhlifi: Yeah this is just temporary, we should merge the fix first!

mava/wrappers/jumanji.py
```diff
@@ -217,3 +222,129 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+        if self.has_global_state:
+            obs_data["global_state"] = global_observation
+            observation = ObservationGlobalState(**obs_data)
+        else:
+            observation = Observation(**obs_data)
+
+        reward = jnp.repeat(timestep.reward, (self.num_agents))
+        discount = jnp.repeat(timestep.discount, (self.num_agents))
+        timestep = timestep.replace(observation=observation, reward=reward, discount=discount)
+        return timestep
```
sash-a: nit
```suggestion
        return timestep.replace(observation=observation, reward=reward, discount=discount)
```

sash-a (commented):
Looks pretty great to me! Just gotta wait for the global state PR to go in so we can get rid of the `hasattr`s

sash-a (dismissed):
LGTM to me also tested locally and seems to be running well!

mava/configs/env/multicvrp.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+env_name: MultiCVRP
+
+# Whether the environment observations encode implicit agent IDs. If True, the AgentID wrapper is not used.
+# This should not be changed.
+implicit_agent_id: False
+
+eval_metric: episode_return
+
+kwargs: {}
```
RuanJohn, line 16: We can just set it to None if it is not getting used. 
```suggestion
kwargs: ~
```

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+        Returns:
+            observations (chex.Array): Concatenated individual observations for each agent,
+                shaped (num_agents, vehicle_info + customer_info).
+            global_observation (Union[None, chex.Array]): Concatenated global observation
+                shaped (num_agents, global_info) if has_global_state = True, None otherwise.
+        """
+        global_observation = None
+        # N: number of nodes, same as _num_customers + 1
+        # V: number of vehicles, same as num_agents
+        # nodes are composed of (x, y, demands)
```
RuanJohn:
```suggestion
        # Nodes are composed of (x, y, demands)
```

mava/configs/env/multicvrp.yaml
Louay-Ben-nessir: it's not getting used but for consistency with other Jumanji env it needs to be there.
If it's set to ~ it will lead to an error on `train_env = jumanji.make(env_name, generator=generator, **config.env.kwargs)` due to attempting to unpack a Nonetype.
should i make the unpacking of kwargs optional? something like
```
if kwargs in config.env:
     train_env = jumanji.make(env_name, generator=generator, **config.env.kwargs)
else:
     train_env = jumanji.make(env_name, generator=generator)
```

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+        # Windows are composed of (start_time, end_time)
+        # Coeffs are composed of (early, late)
+        # Vehicles have ((x, y), local_time, capacity)
+
+        # Tuple[(N, 3), (N, 2), (N, 2)]
+        customers_info, _ = tree_util.tree_flatten(
+            (observation.nodes, observation.windows, observation.coeffs)
+        )
+        # Tuple[(V, 2), (V, 1), (V, 1)]
+        vehicles_info, _ = tree_util.tree_flatten(observation.vehicles)
```
RuanJohn, line 441: Just a question here, we are only flattening one attribute so the shapes here is a bit tough to follow. Can you make this a bit clearer please.

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+
+        # (N * 7, )
```
RuanJohn: Shouldn't this be `(N, 7)`?

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+        # (V, 4)
+        vehicles_info = jnp.column_stack(vehicles_info)
+
+        if self.has_global_state:
+            # (V * 4 + N * 7, )
+            global_observation = jnp.concatenate((vehicles_info.ravel(), customers_info))
+            # (V, N * 7 + V * 4)
+            global_observation = jnp.tile(global_observation, (self.num_agents, 1))
+
+        # (V, N * 7)
```
RuanJohn, line 454: Just a general note, it is quite hard to follow where the 4 and 7 in the shapes come from. Is it possible to please elaborate a bit more in the comments?

RuanJohn (changes requested):
Thank you @Louay-Ben-nessir! :fire: Just a few question and suggestions from my side.

mava/wrappers/jumanji.py
Louay-Ben-nessir: we are also using `.ravel()` which flattens the observations leading to the N*7 shape

mava/configs/env/multicvrp.yaml
RuanJohn: Ah good point, let's just leave it as is in that case.

sash-a (dismissed):

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+
+        return specs.Spec(Observation, "ObservationSpec", **obs_data)
+
+    def reward_spec(self) -> specs.Array:
+        return specs.Array(shape=(self.num_agents,), dtype=float, name="reward")
+
+    def discount_spec(self) -> specs.BoundedArray:
+        return specs.BoundedArray(
+            shape=(self.num_agents,), dtype=float, minimum=0.0, maximum=1.0, name="discount"
+        )
```
WiemKhlifi: Can you please remove the reward and discount specs, all we need for now are the `action_spec `and `obs_spec` 🙏

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+class MultiCVRPWrapper(MultiAgentWrapper):
+    """Wrapper for MultiCVRP environment."""
+
+    def __init__(self, env: MultiCVRP, add_global_state: bool = False):
+        env.num_agents = env._num_vehicles
+        env.time_limit = env._num_customers + 1  # added for consistency
+        env.action_dim = env._num_customers + 1  # n_costumers + 1 starter node
+        super().__init__(env, False)
+        self.has_global_state = add_global_state
+        self._env = env
```
WiemKhlifi: NIT:
``` suggestion
        self.num_agents = env._num_vehicles
        self.num_customers = env._num_customers
        self.time_limit = self._num_customers + 1  # added for consistency
        self.action_dim = self._num_customers + 1  # n_customers + 1 starter node
        super().__init__(env, False)
        self.has_global_state = add_global_state
        self._env: MultiCVRP
```

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+
+    def modify_timestep(self, timestep: TimeStep, state: State) -> TimeStep[Observation]:
+        observation, global_observation = self._flatten_observation(timestep.observation)
+        obs_data = {
+            "agents_view": observation,
+            "action_mask": timestep.observation.action_mask,
+            "step_count": jnp.repeat(state.step_count, (self.num_agents)),
```
WiemKhlifi: NIT:
```  suggestion
            "step_count": jnp.repeat(state.step_count, self.num_agents),
```

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+        }
+        if self.has_global_state:
+            obs_data["global_state"] = global_observation
+            observation = ObservationGlobalState(**obs_data)
+        else:
+            observation = Observation(**obs_data)
+
+        reward = jnp.repeat(timestep.reward, (self.num_agents))
+        discount = jnp.repeat(timestep.discount, (self.num_agents))
```
WiemKhlifi: NIT:
``` suggestion
        reward = jnp.repeat(timestep.reward, self.num_agents)
        discount = jnp.repeat(timestep.discount, self.num_agents)
```

mava/wrappers/jumanji.py
```diff
@@ -265,3 +272,124 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+
+        # (V, N * 7)
+        customers_info = jnp.tile(customers_info, (self.num_agents, 1))
+        # (V, 4 + N * 7)
+        observations = jnp.column_stack((vehicles_info, customers_info))
+        return observations, global_observation
+
+    def observation_spec(self) -> specs.Spec[Observation]:
+        step_count = specs.BoundedArray(
+            (self.num_agents,), jnp.int32, 0, self._env._num_customers + 1, "step_count"
```
WiemKhlifi: `self._env._num_customers` -> `self.num_customers`

WiemKhlifi (changes requested):

mava/wrappers/jumanji.py
```diff
@@ -372,3 +379,117 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+    """Wrapper for MultiCVRP environment."""
+
+    def __init__(self, env: MultiCVRP, add_global_state: bool = False):
+        env.num_agents = env._num_vehicles
+        env.time_limit = env._num_customers + 1  # added for consistency
+        env.action_dim = env._num_customers + 1  # n_costumers + 1 starter node
+        self.has_global_state = add_global_state
+        self.num_customers = env._num_customers
+        self._env = env
+        super().__init__(env, False)
```
WiemKhlifi, line 394: I removed this in my last comment because there's no need to initialize the environment since the parent class does it for you.

```suggestion
        super().__init__(env, False)
```

WiemKhlifi: I removed this in my last comment because there's no need to initialize the environment since the parent class does it for you.

```suggestion
        super().__init__(env, False)
```

mava/wrappers/jumanji.py
```diff
@@ -372,3 +379,117 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
+            )
+            obs_data["global_state"] = global_state
+            return specs.Spec(ObservationGlobalState, "ObservationSpec", **obs_data)
+
+        return specs.Spec(Observation, "ObservationSpec", **obs_data)
+
+    def action_spec(self) -> specs.Spec:
+        return specs.MultiDiscreteArray(
+            num_values=jnp.full(self.num_agents, self.num_customers + 1)
+        )
```
WiemKhlifi: I didn't pay attention to this last time but can you add a comment here to clarify why you added this `action_spec` also I would suggest overwriting the `action_dim` method instead of defining it manually and it helps to link things better for the why using this `action_spec` instead of the original 🙏

WiemKhlifi (changes requested):

mava/wrappers/jumanji.py
Louay-Ben-nessir: Turns out the base `action_spec` is already sufficient!

mava/wrappers/jumanji.py
```diff
@@ -372,3 +379,112 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
...
 
         return specs.Spec(Observation, "ObservationSpec", **obs_data)
+
+
+class MultiCVRPWrapper(MultiAgentWrapper):
+    """Wrapper for MultiCVRP environment."""
+
+    def __init__(self, env: MultiCVRP, add_global_state: bool = False):
+        env.num_agents = env._num_vehicles
+        env.time_limit = env._num_customers + 1  # added for consistency
```
OmaymaMahjoub: Since I'm not very familiar with MultiCVRP 😅, I wanted to ask if the time_limit is only controlled in this way or if it can be set manually like in other environments?

Louay-Ben-nessir: Jumanji's Multicvrp doesn't offer a way to manually set the `time_limit`. I changed it to None to better indicate this.

OmaymaMahjoub:
This PR has been closed as it requires extensive maintenance and testing before merging. Therefore, we have created an [issue](https://github.com/instadeepai/Mava/issues/1138) to track and address this!

PR #1044, opened 2024-02-27
fix: update Matrax's default episode length in keeping with literature

JemmaLDaniel:
## What?
This changes `time_limit` from 500 to 25 for a Matrax episode.

## Why?
[Describe why the change was made.]
This change aligns our default Matrax episode lengths with the episode length used in _Benchmarking Multi-Agent Deep Reinforcement Learning Algorithms in Cooperative Tasks_, so our measure set results can be compared against this well-known paper.

### DIFF
--- a/mava/configs/env/matrax.yaml
+++ b/mava/configs/env/matrax.yaml
@@ -25,5 +25,4 @@ scenario:
 eval_metric: episode_return
 
 kwargs:
-  # This value is in keeping with 'Benchmarking Multi-Agent Deep Reinforcement Learning Algorithms in Cooperative Tasks'
   time_limit: 25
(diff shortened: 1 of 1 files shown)


mava/configs/env/matrax.yaml
```diff
@@ -25,4 +25,5 @@ scenario:
 eval_metric: episode_return
 
 kwargs:
-  time_limit: 500
+  # This value is in keeping with 'Benchmarking Multi-Agent Deep Reinforcement Learning Algorithms in Cooperative Tasks'
```
WiemKhlifi: Can you please remove this line, we follow the same idea for other env (e.g lbf)🙏

JemmaLDaniel: Sure thing!

WiemKhlifi (approved):
LGTM 🙌

RuanJohn (approved):
Thank you @JemmaLDaniel :fire:

PR #1046, opened 2024-02-27
Feat: Custom global state for jumanji envs

SimonDuToit:
## What?
The jumanji MA wrapper has been given a get_global_state method to allow jumanji wrappers to specify their own global state. The GlobalStateWrapper has been removed and the generic global state based on concatenating agent observations has been moved to the generic jumanji MA wrapper. The connector wrapper, which implements its own global state, has also been modified accordingly.
## Why?
To allow jumanji wrappers to implement a custom global state.


mava/utils/make_env.py
```diff
@@ -59,10 +58,6 @@
 def add_optional_wrappers(
     env: Environment, config: DictConfig, add_global_state: bool = False
```
WiemKhlifi: The  `add_global_state` is no longer used.

mava/utils/make_env.py
```diff
@@ -91,10 +86,12 @@ def make_jumanji_env(
     # Create envs.
     env = jumanji.make(env_name, generator=generator, **config.env.kwargs)
     eval_env = jumanji.make(env_name, generator=generator, **config.env.kwargs)
-    env, eval_env = wrapper(env), wrapper(eval_env)
+    env, eval_env = wrapper(env, has_global_state=add_global_state), wrapper(
+        eval_env, has_global_state=add_global_state
+    )
```
WiemKhlifi: This way is cleaner:
``` Suggestion
    env = wrapper(env, add_global_state)
   eval_env = wrapper( eval_env, add_global_state)
```

mava/utils/make_env.py
```diff
@@ -91,10 +86,12 @@ def make_jumanji_env(
...
 
-    env = add_optional_wrappers(env, config, add_global_state)
-    eval_env = add_optional_wrappers(eval_env, config, add_global_state)
+    env = add_optional_wrappers(env, config)
+    eval_env = add_optional_wrappers(eval_env, config)
 
     env = AutoResetWrapper(env)
     env = RecordEpisodeMetrics(env)
```
WiemKhlifi: One tiny request since we're here can you please add these wrappers to the `add_optional_wrappers` function for only "`env`" since it's used in all make fcts 🙏

mava/wrappers/jumanji.py
```diff
@@ -34,35 +34,82 @@
...
+    def __init__(self, env: Environment, has_global_state: bool = False):
         super().__init__(env)
         self._num_agents = self._env.num_agents
         self.time_limit = self._env.time_limit
+        self.has_global_state = has_global_state
 
     def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
         """Modify the timestep for `step` and `reset`."""
-        pass
+        return timestep
```
WiemKhlifi: I vote for defining it as abstract or maybe raising an error ( message: not implemented method).  What do you think?

``` suggestion
    @abstarctmethod
    def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
        """Modify the timestep for `step` and `reset`."""
        ...

```

mava/wrappers/jumanji.py
```diff
@@ -34,35 +34,82 @@
...
+                "global_state",
+            )
+
+            return specs.Spec(
+                ObservationGlobalState,
+                "ObservationSpec",
+                agents_view=obs_spec.agents_view,
+                action_mask=obs_spec.action_mask,
+                global_state=global_state,
+                step_count=obs_spec.step_count,
```
WiemKhlifi: The '`step_count` should be changed for `ObservationGlobalState` as well.

mava/wrappers/jumanji.py
```diff
@@ -34,35 +34,82 @@
...
+            global_state = self.get_global_state(timestep.observation)
+            observation = ObservationGlobalState(
+                global_state=global_state,
+                agents_view=timestep.observation.agents_view,
+                action_mask=timestep.observation.action_mask,
+                step_count=timestep.observation.step_count,
+            )
+            return state, timestep.replace(observation=observation)
+        else:
+            return state, timestep
```
WiemKhlifi:
``` suggestion
         return state, timestep
```

mava/wrappers/jumanji.py
```diff
@@ -34,35 +34,82 @@
...
```
WiemKhlifi:
``` suggestion
         return state, timestep
```

mava/wrappers/jumanji.py
```diff
@@ -153,24 +198,13 @@ def create_agents_view(grid: chex.Array) -> chex.Array:
...
         }
 
-        if self.has_global_state:
-            obs_data["global_state"] = create_global_state(timestep.observation.grid)
-            return timestep.replace(observation=ObservationGlobalState(**obs_data))
-        else:
-            return timestep.replace(observation=Observation(**obs_data))
+        return timestep.replace(observation=Observation(**obs_data))
 
     def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalState]]:
```
WiemKhlifi, line 234: Suggestion for the obs_spec:
``` suggestion
sef observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalState]]:
        """ Specification of the observation of the environment."""
            (self.num_agents,),
            int,
            jnp.zeros(self.num_agents, dtype=int),
            jnp.repeat(self.time_limit, self.num_agents),
            "step_count",
        )

        agents_view = specs.BoundedArray(
            shape=(self._env.num_agents, self._env.grid_size, self._env.grid_size, 5),
            dtype=bool,
            name="agents_view",
            minimum=False,
            maximum=True,
        )
        obs_data = {
            "agents_view": agents_view,
            "action_mask": self._env.observation_spec().action_mask,
            "step_count": step_count,
        }
        if self.has_global_state:
            global_state = specs.BoundedArray(
                shape=(self._env.num_agents, self._env.grid_size, self._env.grid_size, 3),
                dtype=bool,
                name="global_state",
                minimum=False,
                maximum=True,
            )
            obs_data["global_state"] = global_state
            return specs.Spec(ObservationGlobalState, "ObservationSpec", **obs_data)

```

WiemKhlifi (changes requested):
Thank you @SimonDuToit for these changes 🚀
Just minor changes to address 🙏

mava/utils/make_env.py
SimonDuToit: Im not sure about this. They aren't used in the `eval_env`, which also uses `add_optional_wrappers`.

mava/wrappers/jumanji.py
SimonDuToit: Yes, I think this makes sense.

mava/utils/make_env.py
WiemKhlifi: Yeah the best way is to return both train_env and eval_train from the `add_optional_wrappers` function to reduce redundancy and to add the optional wrappers without extra if statements.

WiemKhlifi: At the moment this function is doing only one job so let's increase its importance 😅

mava/wrappers/jumanji.py
SimonDuToit: Yes, this is much neater, thank you.

mava/utils/make_env.py
SimonDuToit: Okay, that makes sense

SimonDuToit: Unless, we just remove `add_optional_wrappers`. I mean, these two wrappers aren't really "optional".

mava/utils/make_env.py
SimonDuToit: I agree, thanks

mava/utils/make_env.py
```diff
@@ -91,10 +86,12 @@ def make_jumanji_env(
     # Create envs.
     env = jumanji.make(env_name, generator=generator, **config.env.kwargs)
     eval_env = jumanji.make(env_name, generator=generator, **config.env.kwargs)
-    env, eval_env = wrapper(env), wrapper(eval_env)
+    env, eval_env = wrapper(env, has_global_state=add_global_state), wrapper(
```
sash-a: Can we standardize the names also, I think I prefer `add_global_state` over `has_global_state`

mava/utils/make_env.py
WiemKhlifi: Perhaps change the name of the function to `add_extra_wrappers ` or another more appropriate name? Otherwise we'll still have at least 7 repeated lines in every make function if we remove this fct 🤔

mava/wrappers/jumanji.py
WiemKhlifi: My pleasure 🙏

mava/utils/make_env.py
sash-a: I think we should just remove this function since it's only doing one thing

mava/wrappers/jumanji.py
```diff
@@ -34,35 +34,82 @@
...
         self.time_limit = self._env.time_limit
+        self.has_global_state = has_global_state
 
     def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
         """Modify the timestep for `step` and `reset`."""
-        pass
+        return timestep
+
+    def get_global_state(self, obs: Observation) -> chex.Array:
+        """Add global state to observations"""
```
sash-a:
```suggestion
        """The default way to create a global state for an environment if it has no 
        available global state - concatenate all observations
        """
```

mava/wrappers/jumanji.py
```diff
@@ -217,3 +251,7 @@ def observation_spec(self) -> specs.Spec[Union[Observation, ObservationGlobalSta
             )
 
         return spec
+
+    def get_global_state(self, obs: Observation) -> chex.Array:
+        global_state = obs.agents_view[..., :3]
+        return global_state
```
sash-a: Can you add a docstring here please, it's not clear why you do `[..., :3]`

sash-a (changes requested):
Looking good! Just needs some docstring updates

mava/utils/make_env.py
SimonDuToit: Ah I see what you mean. Okay, Ill move everything to an `add_extra_wrappers` function

SimonDuToit:
I've implemented all the suggested changes.

mava/utils/make_env.py
```diff
@@ -93,17 +92,14 @@ def make_jumanji_env(
...
-    env, eval_env = wrapper(env), wrapper(eval_env)
-
-    env = add_optional_wrappers(env, config, add_global_state)
-    eval_env = add_optional_wrappers(eval_env, config, add_global_state)
-
-    env = AutoResetWrapper(env)
-    env = RecordEpisodeMetrics(env)
+    train_env, eval_env = wrapper(train_env, add_global_state=add_global_state), wrapper(
+        eval_env, add_global_state=add_global_state
+    )
```
WiemKhlifi: NIT:

```suggestion
    train_env = wrapper(train_env, add_global_state=add_global_state) 
    eval_env = wrapper(eval_env, add_global_state=add_global_state)

```

mava/wrappers/jumanji.py
```diff
@@ -153,67 +208,51 @@ def create_agents_view(grid: chex.Array) -> chex.Array:
...
+            jnp.repeat(self.time_limit, self.num_agents),
             "step_count",
         )
-
         agents_view = specs.BoundedArray(
             shape=(self._env.num_agents, self._env.grid_size, self._env.grid_size, 5),
             dtype=bool,
             name="agents_view",
             minimum=False,
             maximum=True,
```
WiemKhlifi: Can you please double check the min-max values in the `ConnectorWrapper `spec for both `agents_view` and `global_state`? Since they're inferred from the grid spec they should have values outer the interval [0,1] ([grid_spec in jumanji](https://github.com/instadeepai/jumanji/blob/632958f2e1652bd786dc25ddd43681de2137c18d/jumanji/environments/routing/connector/env.py#L330)).

mava/wrappers/jumanji.py
```diff
@@ -34,26 +35,58 @@
...
-        return state, self.modify_timestep(timestep)
+        timestep = self.modify_timestep(timestep)
+        if self.add_global_state:
+            global_state = self.get_global_state(timestep.observation)
+            observation = ObservationGlobalState(
+                global_state=global_state,
+                agents_view=timestep.observation.agents_view,
+                action_mask=timestep.observation.action_mask,
+                step_count=timestep.observation.step_count,
+            )
```
WiemKhlifi, line 70: Can we please move this from `reset` and `step`  to the `get_global_state` function and rename it `get_global_obs` 🙏

mava/wrappers/jumanji.py
```diff
@@ -34,26 +35,58 @@
 
 
 class MultiAgentWrapper(Wrapper):
-    def __init__(self, env: Environment):
+    def __init__(self, env: Environment, add_global_state: bool = False):
```
WiemKhlifi: NIT: 
``` suggestion
    def __init__(self, env: Environment, add_global_state: bool):
```

WiemKhlifi (changes requested):

mava/wrappers/jumanji.py
SimonDuToit: The wrapper transforms the observation representation to make it much easier to learn from, which also changes the interval. This change will likely eventually go into Jumanji itself but for now its implemented in the wrapper.

mava/wrappers/jumanji.py
SimonDuToit: The problem with moving it into `get_global_obs` is that it will have to be repeated every time a wrapper overwrites the method. So I think the current way has less code duplication.

mava/wrappers/jumanji.py
WiemKhlifi: I see... Thanks for the explanation 🙏

mava/wrappers/jumanji.py
WiemKhlifi: Humm right this will add redundancy with the upcoming wrappers. Otherwise I'm happy with the current changes 🙌

mava/wrappers/jumanji.py
```diff
@@ -153,67 +208,51 @@ def create_agents_view(grid: chex.Array) -> chex.Array:
...
-            )
+        return specs.Spec(Observation, "ObservationSpec", **obs_data)
+
+    def get_global_state(self, obs: Observation) -> chex.Array:
+        """Constructs the global state from the global information
+        in the agent observations (positions, targets and paths.)"""
 
-        return spec
+        global_state = obs.agents_view[..., :3]
+        return global_state
```
WiemKhlifi: NIT:
``` suggestion
        return obs.agents_view[..., :3]
```

WiemKhlifi (dismissed):
LGTM 🥳

mava/wrappers/jumanji.py
```diff
@@ -153,67 +208,51 @@ def create_agents_view(grid: chex.Array) -> chex.Array:
...
-            spec = specs.Spec(
-                Observation,
-                "ObservationSpec",
-                agents_view=agents_view,
-                action_mask=self._env.observation_spec().action_mask,
-                step_count=step_count,
-            )
+        return specs.Spec(Observation, "ObservationSpec", **obs_data)
+
+    def get_global_state(self, obs: Observation) -> chex.Array:
```
sash-a: Very minor: can the spec methods always be the last methods in the class

sash-a (dismissed):
Also happy, one tiny comment

mava/wrappers/matrax.py
```diff
@@ -21,30 +21,33 @@
...
-            reward=timestep.reward,
-            discount=timestep.discount,
-        )
+        obs_data = {
+            "agents_view": timestep.observation.agent_obs,
+            "action_mask": self.action_mask,
+            "step_count": jnp.repeat(timestep.observation.step_count, self._num_agents),
+        }
+        if self.add_global_state:
+            obs_data["global_state"] = timestep.observation.agent_obs
```
RuanJohn: Even though the obs are just zeros, I wonder if we shouldn't still concatenate here just to make the centralised critic condition on something of at least a different shape to the agents.

SimonDuToit: Okay. Its changed now.

mava/wrappers/jumanji.py
```diff
@@ -35,26 +37,58 @@
...
 
 class MultiAgentWrapper(Wrapper):
-    def __init__(self, env: Environment):
+    def __init__(self, env: Environment, add_global_state: bool):
         super().__init__(env)
         self.num_agents = self._env.num_agents
         self.time_limit = self._env.time_limit
+        self.add_global_state = add_global_state
 
+    @abstractmethod
```
sash-a, line 46: If you mark this as abstract then the class should also be abstract

mava/wrappers/jumanji.py
```diff
@@ -35,26 +37,58 @@
...
     def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
         """Modify the timestep for `step` and `reset`."""
         pass
 
+    def get_global_state(self, obs: Observation) -> chex.Array:
+        """The default way to create a global state for an environment if it has no
+        available global state - concatenate all observations
```
sash-a: nit
```suggestion
        available global state - concatenate all observations.
```

sash-a (dismissed):
Looks good, two super minor things

WiemKhlifi (dismissed):
LGTM thanks @SimonDuToit 🚀 🥳

mava/configs/env/mabrax.yaml
```diff
@@ -15,5 +15,6 @@ eval_metric: episode_return
 
 kwargs:
   episode_length: 1000
-  homogenisation_method: max # Default is None. Options: [max, concat] if dimensions of observations and actions are not homogeneous across agents,
-  # "max" adds one-hot for the agent's ID. In this case, system.add_ids_to_agent_view should be set to False.
+  homogenisation_method: max # Default is None. Options: [max, concat] if dimensions of observations and actions are not homogeneous across agents.
+  # Note: When using "max", a one-shot vector representing the agent ID is added. Therefore ensure that the `system.add_ids_to_agent_view=False`
```
sash-a: Sorry while we're here can we use the correct config name
```suggestion
  # Note: When using "max", a one-shot vector representing the agent ID is added. Therefore ensure that the `system.add_agent_ids=False`
```

mava/utils/make_env.py
```diff
@@ -97,6 +97,10 @@ def make_jumanji_env(
     train_env = wrapper(train_env, add_global_state=add_global_state)
     eval_env = wrapper(eval_env, add_global_state=add_global_state)
 
+    if env_name in {"MaConnector-v2"}:
+        # The environment has implicit agent IDs
+        config.system.add_agent_id = False
```
sash-a: As discussed offline add an `implicit_agent_id` config option

sash-a (changes requested):
Slightly more general solution :smile:

mava/configs/env/mabrax.yaml
```diff
@@ -13,6 +13,10 @@ scenario:
 # This metric is returned at the end of an experiment and can be used for hyperparameter tuning.
 eval_metric: episode_return
 
+# Whether the environment observations encode implicit agent IDs. If True, the AgentID wrapper is not used.
+# This should not be changed.
+implicit_agent_id: False
```
WiemKhlifi: This should be True since we're using the `homogenisation_method: max` that adds the IDs.  Can we add the note down on setting the agent_ids to false next to the `implicit_agent_id` flag 🙏

mava/utils/make_env.py
```diff
@@ -97,9 +97,8 @@ def make_jumanji_env(
     train_env = wrapper(train_env, add_global_state=add_global_state)
     eval_env = wrapper(eval_env, add_global_state=add_global_state)
 
-    if env_name in {"MaConnector-v2"}:
-        # The environment has implicit agent IDs
-        config.system.add_agent_id = False
+    # Disable the AgentID wrapper if the environment has implicit agent IDs.
+    config.system.add_agent_id = config.system.add_agent_id & (~config.env.implicit_agent_id)
```
WiemKhlifi:
```suggestion
    config.system.add_agent_id = ~config.env.implicit_agent_id
```

WiemKhlifi (changes requested):
Minor changes

mava/configs/env/mabrax.yaml
SimonDuToit: Makes sense, I changed it.

mava/utils/make_env.py
SimonDuToit: I chatted with Sasha and we think we shouldn't do this. It makes config.system.add_agent_id redundant. There may be cases where we can use the AgentID wrapper but arent sure whether it improves performance.

mava/configs/env/mabrax.yaml
```diff
@@ -13,8 +12,11 @@ scenario:
 # This metric is returned at the end of an experiment and can be used for hyperparameter tuning.
 eval_metric: episode_return
 
+# Whether the environment observations encode implicit agent IDs. If True, the AgentID wrapper is not used.
+# Note: When using homogenisation_method = "max", a one-shot vector representing the agent ID is added.
```
sash-a: one-hot not shot

```suggestion
# Note: When using homogenisation_method = "max", a one-hot vector representing the agent ID is added.
```

mava/utils/make_env.py
```diff
@@ -93,17 +92,16 @@ def make_jumanji_env(
...
-
-    env = add_optional_wrappers(env, config, add_global_state)
-    eval_env = add_optional_wrappers(eval_env, config, add_global_state)
+    train_env = wrapper(train_env, add_global_state=add_global_state)
+    eval_env = wrapper(eval_env, add_global_state=add_global_state)
 
-    env = AutoResetWrapper(env)
-    env = RecordEpisodeMetrics(env)
+    # Disable the AgentID wrapper if the environment has implicit agent IDs.
+    config.system.add_agent_id = config.system.add_agent_id & (~config.env.implicit_agent_id)
```
sash-a: Can we do this for all the envs? Maybe move this to `add_extra_wrappers`?

mava/wrappers/jumanji.py
```diff
@@ -159,24 +215,19 @@ def create_agents_view(grid: chex.Array) -> chex.Array:
...
-        if self.has_global_state:
-            obs_data["global_state"] = create_global_state(timestep.observation.grid)
-            return timestep.replace(observation=ObservationGlobalState(**obs_data))
-        else:
-            return timestep.replace(observation=Observation(**obs_data))
+        return timestep.replace(observation=Observation(**obs_data))
+
+    def get_global_state(self, obs: Observation) -> chex.Array:
+        """Constructs the global state from the global information
+        in the agent observations (positions, targets and paths.)"""
```
sash-a: nit: newline
```suggestion
        in the agent observations (positions, targets and paths.)
        """
```

sash-a (changes requested):
Couple minor things, but this looks pretty much good to go from me

sash-a (approved):
:tada: :fire: :fire_extinguisher:

WiemKhlifi (approved):
Thanks @SimonDuToit 🛥️

PR #1051, opened 2024-03-04
New mava network: epsilon greedy action head for discrete action spaces

lbeyers:
## What?
An additional network was added to the mava networks file. It fits into the final part of an RNN for Q-learning, and outputs a distribution that can be sampled in the same way that action distributions for the other systems are sampled, as well as the q-values obtained.
## Why?
To be consistent with other mava systems in the Q-learning implementations. The change has the side effect of simplifying the rec-iql file.
## How?
A new network was added which performs a lot of the epsilon greedy maths.
## Extra
Concerns: 
- Though the speed seems comparable, it is possible that this way might be slower.
- I see there is this IdentityTransformation applied in the other discrete action head. Why is this done, and should I also do this?
- Should some of the calculations be moved to the "distributions.py" folder?


mava/networks.py
```diff
@@ -150,6 +150,81 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+
+        # UNIFORM PART (eps %)
+        # generate uniform probabilities across all allowable actions at most granular level
+        # maybe I should just make a uniform categorical and sample its probs. Anyways
+        uniform_action_probs = jnp.ones_like(q_values)
+        masked_uniform_action_probs = jnp.where(
+            observation.action_mask,
+            uniform_action_probs,
+            jnp.zeros_like(q_values),
+        )
```
sash-a: Isn't observation.action_mask the same as `masked_uniform_action_probs`? Seems like you can skip the `where`

mava/networks.py
```diff
@@ -150,6 +150,81 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+        # get num avail actions to generate probabilities for choosing
+        inner_sum = jnp.sum(masked_uniform_action_probs, axis=-1)[..., jnp.newaxis]
```
sash-a: Slightly clearer name

```suggestion
        n_available_actions = jnp.sum(masked_uniform_action_probs, axis=-1)[..., jnp.newaxis]
```

mava/networks.py
```diff
@@ -150,6 +150,81 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+        # divide with sum along axis to get uniform per action choice
+        masked_uniform_action_probs = masked_uniform_action_probs.astype("int32") / inner_sum
```
sash-a: As a general rule don't use strings for array types. @jcformanek is teaching bad habits :stuck_out_tongue: 

```suggestion
        masked_uniform_action_probs = masked_uniform_action_probs.astype(int) / inner_sum
```

mava/networks.py
```diff
@@ -150,6 +150,81 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+
+        # GREEDY PART (1-eps %)
+        # set masked actions to value not chosen by argmax
+        q_vals_masked = jnp.where(
```
sash-a: very minor nit, this reads a bit better
```suggestion
        masked_q_vals = jnp.where(
```

mava/networks.py
```diff
@@ -150,6 +150,81 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+            observation.action_mask,
+            q_values,
+            jnp.finfo(jnp.float32).min, 
+        )
+
+        # greedy argmax over action-value dim
+        greedy_actions = jnp.argmax(q_vals_masked, axis=-1)
+        # get one-hot so that shapes are equal again and ready to be made into a prob
+        oh_greedy_actions = jax.nn.one_hot(greedy_actions, self.action_dim)
```
sash-a: The `oh` prefix is a bit confusing, just keep the variable name the same as the line above
```suggestion
        greedy_actions = jax.nn.one_hot(greedy_actions, self.action_dim)
```

mava/networks.py
```diff
@@ -150,6 +150,81 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+        NOTE: We pass both the observation embedding and the observation object to the action head
+        since the observation object contains the action mask and other potentially useful
+        information.
+        """
+
+        # q values
+        q_values = nn.Dense(self.action_dim, kernel_init=orthogonal(0.01))(obs_embedding)
+
+        # action mask needs to fit onto the array of action q-vals
+        chex.assert_equal_shape([q_values, observation.action_mask])
```
sash-a: We don't usually have these in networks, but I quite like this!

mava/networks.py
```diff
@@ -150,6 +150,81 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+
+        mixed_eps_greedy_probs = (
+            epsilon * masked_uniform_action_probs + (1 - epsilon) * oh_greedy_actions
+        )
+
+        # if we get a normalisation error here that means
+        # machine error likely creates a discrepancy in the sum
+        # then options are:
+        # -to renormalise and risk erasing an option, (not preferable)
+        # -or revert to the system that doesn't mix eps-greedy into one distribution
```
sash-a: As discussed offline I think we can take this out :smile:

mava/networks.py
```diff
@@ -150,6 +150,81 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+        eps_greedy_dist = tfd.Categorical(probs=mixed_eps_greedy_probs)
+
+        # q values must be returned for q learning, else we can't double-q-learning-select
+        return q_values, eps_greedy_dist
```
sash-a: Would be nice if we could just return a single thing here as we do for other network heads, lets see if we can work out a way to do this

sash-a (changes requested):
Some minor comments, but looking really nice :smile:

mava/networks.py
lbeyers: Good catch, I agree I can skip the where. Maybe just a cast.

mava/networks.py
lbeyers: Okay cool, will leave it there! It saved me some debugging.

mava/networks.py
```diff
@@ -151,6 +151,71 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+        """Action selection for distrete action space environments.
+
+        Args:
+            obs_embedding: Observation embedding from network torso.
+            observation: Observation object containing `agents_view`, `action_mask` and
+                `step_count`.
+
+        Returns:
+            Q values for double Q-learning selection
+            A tfd.Categorical distribution over the action space for sampling actions from.
```
RuanJohn: Nit:

```suggestion
            q_values: used for double Q-learning selection.
            eps_greedy_dist: tfd.Categorical distribution for sampling actions.
```

mava/networks.py
```diff
@@ -151,6 +151,71 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+
+        NOTE: We pass both the observation embedding and the observation object to the action head
+        since the observation object contains the action mask and other potentially useful
+        information.
+        """
+
+        # q values
```
RuanJohn:
```suggestion
```

RuanJohn (changes requested):
Thanks @lbeyers, just a few minor comments from me for now. Rest discussed offline.

mava/networks.py
EdanToledo: You can just make your own tfp distribution that acts the same way as the distrax epsilon greedy distribution, that way you can access the q values using the preferences/q_values attribute and you can sample from it since its a distribution.

lbeyers: Thanks for your comment! Sasha and I had a conversation about this. If you want to follow what's happening, there will be an upcoming PR, but essentially we need more functionality than the distrax distribution, and the suitable place to put the q values was decided to not be in the eps-greedy distribution.

EdanToledo: sounds good! I will be following (i know its not really my business but i like checking out the mava PRs to see the great changes). I'd love to know the reason why its better to not be in dist. looking forward to the PR 🚀🚀🚀

lbeyers: Essentially, it's my opinion that the q values are not really "part" of this distribution, since the only thing reflected in the distribution from the q values is the argmax. The idea is in future to make different calls to the network, so that if we don't need the eps-greedy distribution (i.e. when we are only accessing the q-values), we don't have to go through the process of making an eps-greedy distribution. We never need both from the same network call.

lbeyers:
I think this is ready to merge - definitely check the docstrings I added to the distribution, it's not the traditional format but I think it may be helpful!

mava/networks.py
```diff
@@ -151,6 +155,42 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
         )
 
 
+class DiscreteActionEpsGreedyMaskedHead(nn.Module):
```
sash-a: Discrete action is implied in my opinion

```suggestion
class MaskedEpsilonGreedyHead(nn.Module):
```

mava/networks.py
```diff
@@ -151,6 +155,42 @@ def __call__(self, obs_embedding: chex.Array, observation: Observation) -> tfd.I
...
+                `step_count`.
+
+        Returns:
+            q_values: used for double Q-learning selection.
+            eps_greedy_dist: an eps-greedy initialised tfd.Categorical for sampling actions.
+
+        NOTE: We pass both the observation embedding and the observation object to the action head
+        since the observation object contains the action mask and other potentially useful
+        information.
+        """
```
sash-a: Convention is for extra docs to go 1 line below the main explainer, but before args/return
```suggestion
        """Action selection for distrete action space environments.

        NOTE: We pass both the observation embedding and the observation object to the action head
        since the observation object contains the action mask and other potentially useful
        information.
        
        Args:
            obs_embedding: Observation embedding from network torso.
            observation: Observation object containing `agents_view`, `action_mask` and
                `step_count`.

        Returns:
            q_values: used for double Q-learning selection.
            eps_greedy_dist: an eps-greedy initialised tfd.Categorical for sampling actions.
        """
```

sash-a (commented):
Very minor changes + what we chatted about offline about the `get_value` method

lbeyers:
Discussed offline: will go in with the IQL PR!

PR #1052, opened 2024-03-04
feat: mkdocs documentation

liamclarkza:
Adds mkdocs support for readthedocs. View the current feature branch docs here: https://id-mava.readthedocs.io/en/feat-mkdocs-documentation/

Additionally, the readme has been simplified as things like benchmarks have been moved to the docs pages.

Note: the User Guide and Evaluation pages are currently "in the works" and are not currently displayed on readthedocs (they have been commented out in the mkdocs.yml file)


sash-a:
Something that could be good to add to the sharp bits section is to have a part on how `total_timesteps` and `num_updates` interact with each other and why we have both config options

sash-a:
Unfortunately this is too stale :(

PR #1053, opened 2024-03-04
fix: fix CNN torso to maintain batch and agent dimension

RuanJohn:
## What?
The `CNNTorso` was still expecting networks to be `vmap`-ed over the agent dimension so the final reshape before going to a policy head was remove the agent dim. 
This is just a small fix in the reshape. 

## Extra
Added a batch dimension to the feedforward evaluator (similar to what we do in the recurrent systems)

## To test
`python mava/systems/ppo/ff_ippo.py env=connector network=cnn` similarly for `ff_mappo`. 

## Note
We do not support recurrent systems with CNN pre-torsos yet.

### DIFF
--- a/mava/evaluator.py
+++ b/mava/evaluator.py
@@ -63,12 +63,18 @@ def get_ff_evaluator_fn(
 
             # Select action.
             key, policy_key = jax.random.split(key)
-            pi = apply_fn(params, last_timestep.observation)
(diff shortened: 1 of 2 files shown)


OmaymaMahjoub (approved):
LGTM!

sash-a (approved):
:spider:

PR #1055, opened 2024-03-04
Generalise the evaluator factory to accept a custom apply function instead of enforcing network.apply

lbeyers:
## What?
The evaluator now accepts an apply function (must still fit RecActorApply or ActorApply types depending on the flags) instead of taking network.apply one level into the process.
## Why?
For Q-learning, it will likely be necessary. It also makes the evaluator slightly more general while keeping the effort of building the evaluator to pretty much the same. Only downside is people have more freedom to enter what they like and so it may be easier to make a mistake there. With slightly greater power comes slightly greater responsibility.
## How?
The make_eval_fns function now takes in a function instead of a network, which it passes directly to the evaluator builder.
## Extra
All systems on develop were slightly altered to build their evaluators like this so that if the change is accepted, we still have functional PPO. The change is just network -> network.apply when we build the evaluator (so once in each file).


RuanJohn (dismissed):
Thank you for this @lbeyers :raised_hands:

mava/evaluator.py
```diff
@@ -280,7 +280,7 @@ def evaluator_fn(
 
 def make_eval_fns(
     eval_env: Environment,
-    network: Any,
+    network_apply_fn: Any,
```
sash-a: Typing this would be nice :smile: 
You may need to import Callable

```suggestion
    network_apply_fn: Callable[[FrozenDict, Observation], Distribution],
```

mava/evaluator.py
sash-a, line 1: While we're here could we also add better docs to the `make_eval_fns` method.

sash-a (changes requested):
Looks great, very minor thing so that people know what the `network_apply_fn` looks like

mava/evaluator.py
lbeyers: Typing would be nice, only problem is that at that level it can legally be a RecActorApply or ActorApply function, so it can never be more specific than a generalisation of those. What do you think of assigning it just "Callable" ?

sash-a: Ah lets do `Union[ActorApply, RecActorApply]`

lbeyers:
Thanks for all the comments - please let me know if you are happy with this. I do not yet know how to mark the requested changes as handled.

mava/evaluator.py
```diff
@@ -286,36 +287,60 @@ def evaluator_fn(
...
     eval_env: Environment,
-    network: Any,
+    network_apply_fn: Union[ActorApply, RecActorApply],
     config: DictConfig,
     use_recurrent_net: bool = False,
+    hidden_carry_size: Optional[int] = None,
     scanned_rnn: Optional[nn.Module] = None,
 ) -> Tuple[EvalFn, EvalFn]:
-    """Initialise evaluator_fn."""
+    """ "Initialize evaluator functions for reinforcement learning.
```
sash-a: nit
```suggestion
    """Initialize evaluator functions for reinforcement learning.
```

mava/evaluator.py
```diff
@@ -286,36 +287,60 @@ def evaluator_fn(
...
+        config (DictConfig): The configuration settings for the evaluation.
+        use_recurrent_net (bool, optional): Whether to use a rnn. Defaults to False.
+        hidden_carry_size (Optional[int], optional): Size of the hidden carry to
+            initialise for the RNN. Required if `use_recurrent_net` is True. Defaults to None.
+        scanned_rnn (Optional[nn.Module], optional): The rnn module.
+            Required if `use_recurrent_net` is True. Defaults to None.
+
+    Returns:
+        Tuple[EvalFn, EvalFn]: A tuple of two evaluation functions:
+                            one for use during training and one for overall metrics.
```
sash-a: nit
```suggestion
        Tuple[EvalFn, EvalFn]: A tuple of two evaluation functions:
            one for use during training and one for overall metrics.
```

mava/evaluator.py
```diff
@@ -286,36 +287,60 @@ def evaluator_fn(
...
+
+    Raises:
+        AssertionError: If `use_recurrent_net` is True but `scanned_rnn` is not provided.
```
sash-a, line 307: Never seen this before in a docstring, but it's good extra info :smile:

sash-a (commented):
Looking really good!

mava/evaluator.py
```diff
@@ -286,36 +286,54 @@ def evaluator_fn(
...
+        eval_env (Environment): The environment used for evaluation.
+        network_apply_fn (Union[ActorApply,RecActorApply]): Creates a policy to sample.
+        config (DictConfig): The configuration settings for the evaluation.
+        use_recurrent_net (bool, optional): Whether to use a rnn. Defaults to False.
+        scanned_rnn (Optional[nn.Module], optional): The rnn module.
+            Required if `use_recurrent_net` is True. Defaults to None.
+
+    Returns:
+        Tuple[EvalFn, EvalFn]: A tuple of two evaluation functions:
+        one for use during training and one for overall metrics.
```
sash-a: nit
```suggestion
        one for use during training and one for absolute metrics.
```

sash-a (dismissed):
Super minor nitpick, otherwise I'm happy! :smile:

mava/evaluator.py
```diff
@@ -286,36 +286,54 @@ def evaluator_fn(
...
+    """
     # Check if win rate is required for evaluation.
     log_win_rate = config.env.eval_metric == "win_rate"
     # Vmap it over number of agents and create evaluator_fn.
     if use_recurrent_net:
         assert scanned_rnn is not None
         evaluator = get_rnn_evaluator_fn(
             eval_env,
-            network.apply,
+            network_apply_fn,  # type: ignore
```
RuanJohn, line 316: What is the mypy error we get when we don't type ignore here?

lbeyers: It complains because get_rnn_evaluator_fn expects RecActorApply type there, but the input is Union[ActorApply, RecActorApply] - so it could be either type, which we need. For reference, before this PR we used the type "Any" for passing the network through, so it didn't make mypy angry.

RuanJohn: Makes total sense. I think we can just type ignore in that case.

RuanJohn (changes requested):
This is super close to merging @lbeyers! Can we please just update the sac and masac as well now to pass in the network apply function to the evaluator? 
Thank you :raised_hands:

lbeyers:
Thanks Ruan, updated!

RuanJohn (approved):
Thank you @lbeyers! :rocket:

sash-a (approved):
Lekker :firecracker:

PR #1059, opened 2024-03-08
feat: uncouple hiddenstate form previous layer in recurrent policies

RuanJohn:
## What?
The size of the agent hidden states are current coupled to the dimension of the preceding layer. This PR uncouples and allows for setting the hidden state sizes in the network config. 

## Extra 
Maintains the current defaults. Ie. the hidden state size is 128 since the size of the last layer in the pre torso is 128.


OmaymaMahjoub (approved):
Thanks for the PR @RuanJohn LGTM!

sash-a (approved):
Dankie!

PR #1065, opened 2024-03-18
Make sure that config params passed to fbx.make_item_buffer are ints

liamclarkza:
## What?
Casts the configuration parameters used to set up the replay buffer to be integers as currently if the configuration parameters are set as floats, such as system.buffer_size = 1e6, Mava will crash.


WiemKhlifi (approved):
Nice catch @liamclarkza ✌️

sash-a (approved):
Thanks @liamclarkza

PR #1066, opened 2024-03-18
fix: final return value for SAC systems

sash-a:
## What?
Fix the final return value of the SAC systems to bring them inline with PPO for hydra sweeps


RuanJohn (approved):
:raised_hands:

WiemKhlifi (approved):
✌️

PR #1067, opened 2024-03-19
feat: cnn support for recurrent systems

SimonDuToit:
Previously it was not possible to use a CNN in Mava's recurrent systems. This PR makes CNNs compatible with recurrent systems and adds a relevant config for a recurrent CNN network.


mava/networks.py
```diff
@@ -81,7 +81,7 @@ def __call__(self, observation: chex.Array) -> chex.Array:
             x = self.activation_fn(x)
 
         # Reshape should keep the batch and agent dimensions unchanged.
-        return x.reshape((x.shape[0], x.shape[1], -1))
+        return jax.lax.collapse(x, -3)
```
sash-a, line 85: I'd say remove the comment above for these two
```suggestion
        # Collapse (merge) the last three dimensions (width, height, channels)
        # Leave the batch, agent and time (if recurrent) dims unchanged.
        return jax.lax.collapse(x, -3)
```

sash-a (changes requested):
Small change otherwise looks good. Can we just wait till we've run a couple experiments on connector to see that the recurrent systems perform ok :pray:

sash-a (approved):
Thanks @SimonDuToit :fire:

OmaymaMahjoub (approved):
Thanks @SimonDuToit :fire:

PR #1068, opened 2024-03-20
feat: allow scenarios to set env kwargs

sash-a:
## What?
Allow scenarios to set env kwargs. This is useful for cleaner right now where we want to set different time limits for different scenarios. This will likely be useful in future other attributes of other envs.

Note this is only for jumanji envs as other envs don't have scenario files.


SimonDuToit (approved):
LGTM

OmaymaMahjoub (approved):
LGTM!

PR #1069, opened 2024-03-20
feat: separate config option for logging winrate

sash-a:
## What?
Logging win rate used to depend on `config.env.eval_metric == "win_rate"` however in certain envs we don't want the win rate to be the eval_metric as it is too sparse to optimize for, but we do still want to see the win rate.

This is a stop gap solution for now, we should spend time finding out how we can cleanly log all env metrics regardless of name.

### DIFF
--- a/mava/configs/env/cleaner.yaml
+++ b/mava/configs/env/cleaner.yaml
@@ -13,6 +13,9 @@ eval_metric: episode_return
 # Whether the environment observations encode implicit agent IDs. If True, the AgentID wrapper is not used.
 # This should not be changed.
 implicit_agent_id: True
+# Whether or not to log the winrate of this environment. This should not be changed as not all
+# environments have a winrate metric.
+log_win_rate: True
 
 kwargs:
   time_limit: 100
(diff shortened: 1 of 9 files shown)


SimonDuToit (approved):
LGTM

mava/configs/env/cleaner.yaml
```diff
@@ -13,6 +13,9 @@ eval_metric: episode_return
 # Whether the environment observations encode implicit agent IDs. If True, the AgentID wrapper is not used.
 # This should not be changed.
 implicit_agent_id: True
+# Whether or not to log the winrate of this environment. This should not be changed as not all
+# environments have a winrate metric.
+log_win_rate: True
```
WiemKhlifi, line 18: NIT: can we move them up next to the eval_metric for better visibility 🙏

WiemKhlifi (approved):
LGTM 🙌

mava/configs/env/cleaner.yaml
sash-a: I put it here because both `log_win_rate` and `implicit_agent_id` shouldn't ever be changed I think, so we don't want to it to be too high up because we shouldn't be changing it, what do you think?

PR #1070, opened 2024-03-25
feat: integration tests

### LINKED ISSUE #1049: [FEATURE] Integration tests
We are constantly having to check that new features don't trivially break the existing systems/environments. So it would be nice to run all the systems on a subset of the environments for a very small amount of steps to double check that everything pipes through.

sash-a:
## What?
Adds integration tests

todo:

- [x] fix gigastep wrapper
- [x] fix matrax wrapper
- [x] make github run tests

closes #1049


test/integration_test.py
```diff
@@ -0,0 +1,89 @@
...
+    dconf: dict = OmegaConf.to_container(cfg, resolve=True)
+    dconf["system"] |= fast_config["system"]
+    dconf["arch"] |= fast_config["arch"]
+    cfg = OmegaConf.create(dconf)
+
+    return cfg
+
+
+@pytest.mark.parametrize("system_path", system_paths)
+def test_mava_system(fast_config: dict, system_path: str):
```
WiemKhlifi: If ` ppo.ff_ippo` and ` sac.ff_isac `systems will be tested over all the envs (discrete & continuous) we can remove them from here else we'll have 2 extra unnecessary runs I guess 🤔

test/integration_test.py
```diff
@@ -0,0 +1,89 @@
...
+    with initialize(version_base=None, config_path="../mava/configs/"):
+        cfg = compose(config_name=f"default_{system_name}")
+        cfg = _get_fast_config(cfg, fast_config)
+
+    _run_system(system_path, cfg)
+
+
+@pytest.mark.parametrize("env_name", discrete_envs)
+def test_discrete_env(fast_config: dict, env_name: str):
+    system_path = "ppo.ff_ippo"
```
WiemKhlifi: `rec_iql` system will be discarded? Also I'm wondering if we could add a test for the `rec_ippo` system.

test/conftest.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+
+
+@pytest.fixture
+def fast_config():
+    return {
+        "system": {
+            # common
+            "num_updates": 3,
```
WiemKhlifi: Maybe 2 updates?

mava/configs/default_ff_isac.yaml
```diff
@@ -3,5 +3,5 @@ defaults:
   - logger: ff_isac
   - arch: anakin
   - system: sac/ff_isac
-  - network: mlp
+  - network: continuous_mlp
```
WiemKhlifi: I like the idea 🔥 For a smooth switching between discrete and continuous case can we add this to the other systems' default option:
```
- network: mlp # [mlp, continuous_mlp ]
```

WiemKhlifi (dismissed):
Thanks @sash-a! It's really satisfying to see my chaotic shell script evolve into a more sophisticated and cleaner test version 😅💃

test/integration_test.py
sash-a: Ye, good point!

test/integration_test.py
sash-a: I'd rather do ff systems here so that it's a bit faster. `rec_iql` and `rec_ippo` should be tested in `test_mava_system`.

I think if we do all systems across all envs it'll take too long, so I think as long as we cover all our envs and all our systems we should be ok

WiemKhlifi: Yess I agree with this, thanks for the clarification 🙌

mava/utils/logger.py
```diff
@@ -292,7 +292,7 @@ def log_dict(self, data: Dict, step: int, eval_step: int, event: LogEvent) -> No
         # Replace underscores with spaces and capitalise keys.
         keys = [k.replace("_", " ").capitalize() for k in data.keys()]
         # Round values to 3 decimal places if they are floats.
-        values = [v if isinstance(v, int) else f"{float(v):.3f}" for v in data.values()]
+        values = [v if isinstance(v, (float, int)) else f"{v.item():.3f}" for v in data.values()]
```
RuanJohn: Just needs a small fix here. We will never round with how it is right now if I am not mistaken.

test/integration_test.py
```diff
@@ -0,0 +1,114 @@
...
+    "smax",
+]
+cnn_envs = ["cleaner", "connector"]
+continuous_envs = ["mabrax"]
+
+
+def _run_system(system_name: str, cfg: DictConfig) -> float:
+    """Runs a system."""
+    OmegaConf.set_struct(cfg, False)
+    cfg.logger.use_neptune = False  # we never want to log these tests to neptune
```
RuanJohn:
```suggestion
    # we never want to log these tests anywhere
    cfg.logger.use_neptune = False
    cfg.logger.use_tb = False
    cfg.logger.use_json = False
```

We can do the same for tensorboard and json logging as well here please.

test/integration_test.py
```diff
@@ -0,0 +1,114 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import importlib
+
+import pytest
+from hydra import compose, initialize
+from omegaconf import DictConfig, OmegaConf
+
```
RuanJohn, line 21: Maybe just add a comment here to explain the logic behind the way we are running one system over all envs and then all systems over one env to save time.

RuanJohn: We can also just add some logic that will randomly select an env to run all systems on and then also random select a system to run all envs on.

RuanJohn (changes requested):
Thanks for this @sash-a :fire: 
Just some minor suggestions

RuanJohn (approved):
This is great! Thank you @sash-a :fire:

test/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+            # iql
+            "min_buffer_size": 4,
+            "sample_batch_size": 1,
+            "sample_sequence_length": 1,
+        },
+        "arch": {
+            "num_envs": 1,
+            "num_eval_episodes": 1,
+            "num_evaluation": 1,
+            "absolute_metric": False,
```
OmaymaMahjoub, line 46: Optional Request: Could we test with the absolute metric set to true? We often encounter bugs because we forget to test the absolute metric, especially since it is at the very end of the training and frequently overlooked. Including this test would help reduce that risk.

OmaymaMahjoub (approved):
Thanks @sash-a, I made a final optional request,  If you believe it's unnecessary, feel free to ignore it :sweat_smile:

test/conftest.py
sash-a: Agreed, but I've tested and unfortunately it takes too long with this on

OmaymaMahjoub: Updates: it is taking longer then expected when setting absolute metric to true, so we will keep the flag to false

PR #1071, opened 2024-03-26
Feature: Control number of vmapped envs in evaluator using `arch.num_envs` 

OmaymaMahjoub:
## What?
Modify the evaluator to limit the number of vmapped envs to `arch.num_envs` when the total number of evaluation episodes `arch.num_eval_episodes` exceeds this limit (instead of parallelise all the `arch.num_eval_episodes`). In such cases, evaluations are conducted in sequential batches, with each batch containing `arch.num_envs` parallel envs.

## Why?
Limiting parallel evaluations to num_envs prevents out-of-memory issues by avoiding vmap over all episodes at once.

### DIFF
--- a/mava/evaluator.py
+++ b/mava/evaluator.py
@@ -112,21 +112,21 @@ def get_ff_evaluator_fn(
 
         n_devices = len(jax.devices())
         episodes_per_device = config.arch.num_eval_episodes * eval_multiplier // n_devices
-        parallel_eval_batch_size = min(config.arch.num_envs, episodes_per_device)
+        num_vmapped_episodes = min(config.arch.num_envs, episodes_per_device)
 
         # Initialize eval state for the parallel environments.
-        key, *env_keys = jax.random.split(key, parallel_eval_batch_size + 1)
+        key, *env_keys = jax.random.split(key, num_vmapped_episodes + 1)
         env_states, timesteps = jax.vmap(env.reset)(jnp.stack(env_keys))
 
         # Split keys for parallel execution in vmap.
-        key, *step_keys = jax.random.split(key, parallel_eval_batch_size + 1)
-        step_keys = jnp.stack(step_keys).reshape(parallel_eval_batch_size, -1)
+        key, *step_keys = jax.random.split(key, num_vmapped_episodes + 1)
+        step_keys = jnp.stack(step_keys).reshape(num_vmapped_episodes, -1)
 
         eval_state = EvalState(
             key=step_keys,
             env_state=env_states,
             timestep=timesteps,
-            step_count=jnp.zeros((parallel_eval_batch_size, 1)),
+            step_count=jnp.zeros((num_vmapped_episodes, 1)),
             episode_return=jnp.zeros_like(timesteps.reward),
         )
         # Evaluate over a batch of environments in parallel using vmap.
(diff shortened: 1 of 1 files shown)


mava/evaluator.py
```diff
@@ -104,41 +104,68 @@ def not_done(carry: Tuple) -> bool:
...
 
+    def evaluator_fn(trained_params: FrozenDict, key: chex.PRNGKey) -> ExperimentOutput[EvalState]:
+        """Evaluator function"""
+        n_devices = len(jax.devices())
+
+        # Calculate episodes to evaluate per device.
+        episodes_per_device = config.arch.num_eval_episodes * eval_multiplier // n_devices
+        # Determine parallel evaluation batch size.
+        # Limited by architecture's max environments or episodes per device.
+        parallel_eval_batch_size = min(config.arch.num_envs, episodes_per_device)
```
sash-a: Parallel can be a bit confusing, I thought at first this was the `pmap` dim. Please also change below
```suggestion
        # Determine how many eval episodes to vmap over.
        # Limited by architecture's max environments or episodes per device.
        num_vmapped_episodes = min(config.arch.num_envs, episodes_per_device)
```

mava/evaluator.py
```diff
@@ -104,41 +104,68 @@ def not_done(carry: Tuple) -> bool:
...
-        """Evaluator function."""
+    def parallel_evaluator(
+        carry: Tuple[FrozenDict, chex.PRNGKey], _: Any
+    ) -> Tuple[Tuple[FrozenDict, chex.PRNGKey], ExperimentOutput[EvalState]]:
+        """Evaluates the agent on multiple environments in parallel."""
+        trained_params, key = carry
 
-        # Initialise environment states and timesteps.
         n_devices = len(jax.devices())
+        episodes_per_device = config.arch.num_eval_episodes * eval_multiplier // n_devices
```
WiemKhlifi, line 114: Sadly jax doesn't allow us to carry these fixed values (`parallel_eval_batch_size`) 😢

WiemKhlifi (dismissed):
LGTM, thanks @OmaymaMahjoub for making things more flexible 🙏

mava/evaluator.py
```diff
@@ -104,41 +104,68 @@ def not_done(carry: Tuple) -> bool:
...
+        n_devices = len(jax.devices())
+
+        # Calculate episodes to evaluate per device.
+        episodes_per_device = config.arch.num_eval_episodes * eval_multiplier // n_devices
+        # Determine parallel evaluation batch size.
+        # Limited by architecture's max environments or episodes per device.
+        parallel_eval_batch_size = min(config.arch.num_envs, episodes_per_device)
+        # Compute the number of sequential evaluation batches required per device
+        # to cover all episodes.
+        sequential_eval_batches = episodes_per_device // parallel_eval_batch_size
```
sash-a: If you make keys of shape `(sequential_eval_batches, num_vmapped_episodes)` then you don't need the extra method, you can just call:

```python
jax.lax.scan(jax.vmap(eval_one_episode), None, eval_states)
```

I like this for two reasons, you don't have to calculate `parallel_eval_batch_size` twice and it is also clear that you're scanning and then vmapping over episodes. But this is a big change so if you don't have time it's ok. I think the evaluator is due for a big overhaul anyways

sash-a (changes requested):
Two small things, one would take a while though so if there's no time we can wait till we refactor the evaluator

mava/evaluator.py
WiemKhlifi: I back this idea also 🔥

mava/evaluator.py
OmaymaMahjoub: I am quite sure there is a better way to implement this, but we can keep it as it is at the moment and create an issue calling for cleaning and readibility of the evaluator

mava/evaluator.py
OmaymaMahjoub: I see so even the eval_init will be created inside the one episode, I will give it a try

OmaymaMahjoub:
This PR has been archived because PR #1084 has resolved the issue and includes additional enhancements.

PR #1073, opened 2024-04-12
fix: pin mujoco version for jaxmarl to work

RuanJohn:
## What?
Mujoco had a new release and it breaks JaxMARL at the moment. This is just a quick version pin PR.


OmaymaMahjoub (approved):
Thanks @RuanJohn!

sash-a (approved):

PR #1076, opened 2024-05-09
fix: fix off by one error in rnn dones used in the transition

RuanJohn:
Will add a more detailed description later.


OmaymaMahjoub (approved):
Thanks @RuanJohn :fire:

sash-a (approved):
:mermaid:

PR #1077, opened 2024-05-13
Chore: Connector Update

SimonDuToit:
Update to Connector wrapper and scenarios to bring it in line with the version used in benchmarks. Previously, other agents and targets were marked with binary variables. This has been replaced with integers, so that agents can be distinguished from each other and matched to their corresponding target. Also, rewards and discounts are shared between agents, as this gave the best performance.


mava/wrappers/jumanji.py
```diff
@@ -208,30 +208,44 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
 
         # TARGET = 3 = The number of different types of items on the grid.
         def create_agents_view(grid: chex.Array) -> chex.Array:
-            positions = jnp.where(grid % TARGET == POSITION, True, False)
-            targets = jnp.where((grid % TARGET == 0) & (grid != EMPTY), True, False)
-            paths = jnp.where(grid % TARGET == PATH, True, False)
-            position_per_agent = jnp.where(grid == POSITION, True, False)
-            target_per_agent = jnp.where(grid == TARGET, True, False)
+            positions = jnp.where(grid % TARGET == POSITION, jnp.ceil(grid / TARGET), 0)
+            targets = jnp.where((grid % TARGET == 0) & (grid != EMPTY), jnp.ceil(grid / TARGET), 0)
```
sash-a: One thing that worries me here is that these numbers get large as you scale the number of agents, can we do `jnp.ceil(grid / TARGET) / num_agents` so that the value is normalized? And then also change the obs spec as necessary.

mava/wrappers/jumanji.py
```diff
@@ -208,30 +208,44 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
             )
             return agents_view
 
+        def aggregate_rewards(
+            timestep: TimeStep,
+        ) -> TimeStep[Observation]:
+            """Aggregate individual rewards and discounts across agents."""
+            team_reward = jnp.sum(timestep.reward)
+            reward = jnp.repeat(team_reward, self.num_agents)
+            discount = jnp.repeat(jnp.max(timestep.discount), self.num_agents)
```
sash-a: I still think we should leave the discounts as per agent, especially since we don't even use them in PPO.

sash-a (changes requested):
Looks really good! Just 2 minor things :pray:

mava/wrappers/jumanji.py
SimonDuToit: Allright. In the past doing this gave a big performance boost, likely due to some old Mava bug, but I tested it now and it doesnt change performance.

sash-a (commented):
:ship:

sash-a (dismissed):

mava/wrappers/jumanji.py
```diff
@@ -208,30 +208,48 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
 
         # TARGET = 3 = The number of different types of items on the grid.
         def create_agents_view(grid: chex.Array) -> chex.Array:
-            positions = jnp.where(grid % TARGET == POSITION, True, False)
-            targets = jnp.where((grid % TARGET == 0) & (grid != EMPTY), True, False)
-            paths = jnp.where(grid % TARGET == PATH, True, False)
-            position_per_agent = jnp.where(grid == POSITION, True, False)
-            target_per_agent = jnp.where(grid == TARGET, True, False)
+            positions = (
```
OmaymaMahjoub, line 212: Can you add a commnet here to explain the motivation for having ceil(grid/Target) and why div at the end with num_agents

mava/configs/env/connector.yaml
```diff
@@ -1,7 +1,7 @@
 # ---Environment Configs---
 defaults:
   - _self_
-  - scenario: con-10x10x5a # [con-5x5x3a, con-7x7x5a, con-10x10x5a, con-15x15x10a]
+  - scenario: con-5x5x3a # [con-5x5x3a, con-7x7x5a, con-10x10x10a, con-15x15x23a]
```
OmaymaMahjoub, line 4: I am worried that someone will set the time limit in this file and think it is what will be used however later time_limit is controlled by the param in the scenario.py

OmaymaMahjoub (changes requested):
Thanks @SimonDuToit :fire:, just requires minor changes :)

RuanJohn:
Thank you for this @SimonDuToit! Something that would be really nice is if we could add the vector obs versions of cleaner and connector in Mava as well.

mava/configs/env/connector.yaml
SimonDuToit: Yes, this time limit is redundant.

mava/wrappers/jumanji.py
SimonDuToit: I added a comment. Can you check if you think its sufficient?

SimonDuToit: As an example of what I mean by the comment, the positions of the first two out of N agents would go: [2, 5] -> [1, 2] -> [1/N, 2/N]

sash-a:
> Thank you for this @SimonDuToit! Something that would be really nice is if we could add the vector obs versions of cleaner and connector in Mava as well.

Can we do this in a separate PR though?

mava/wrappers/jumanji.py
OmaymaMahjoub: Yeah that's enough :) The key part that I didn't get was div by agents number, but by mentioning normalization all makes sense!

OmaymaMahjoub (approved):
Thanks @SimonDuToit :medal_sports:

sash-a (approved):
:1st_place_medal:

PR #1084, opened 2024-07-04
feat: refactor evaluator

### LINKED ISSUE #996: Use `TrainState` in evaluator
As discussed in #994 we should now be able to use the `MetricsWrapper` and the `TrainState` in the evaluator instead of manually recording the episode return and creating a special `EvalState`.

### LINKED ISSUE #1001: [FEATURE] Generalise Evaluator Fn
It would be very nice if the evaluator function was generic enough for future algorithms. It can be frustrating that a highly specific form is required. i.e. currently the network apply is force vmapped and it uses the default call function. Complex algorithms have complex networks and multiple different call functions and might not be suitable for this paradigm.

An easy way to do this would be that a user passes in an "actor_fn" which has a specific input and output api. A user would create this and perhaps make a partial fn to fit the correct api. This could be suitable for feedforward and recurrent networks as long as the API is adhered to. I feel that although this is adding structure that could make a user confused, its a simple enough request that it wouldn't require the user to dig deep at all and would benefit creating new algorithms.

sash-a:
## What?
Refactor the evaluator to reduce code duplication and reduce memory usage.
## Why?
The evaluator was one of the oldest functions in mava and it needed a bit of a rewrite as can be seen in the issues #996 and #1001. It was getting a bit complicated to extend and if you set the number of eval episodes too high it would try to create too many parallel environments.
## How?
* Have a single evaluator function that can work for recurrent and feed forward systems (and hopefully any future systems) by taking in a acting function that conforms to the API defined in `evaluator.py`. 
* Only create `num_envs` parallel environments and loop for the required iterations so that we do at least `num_eval_episodes` number of evaluation rollouts.

Closes: #996 and #1001
Also this achieves the goal #1071 so we can close it if this is merged


mava/evaluator.py
```diff
@@ -12,329 +12,140 @@
...
-        Tuple[EvalFn, EvalFn]: A tuple of two evaluation functions:
-        one for use during training and one for absolute metrics.
-
-    Raises:
-        AssertionError: If `use_recurrent_net` is True but `scanned_rnn` is not provided.
+        env: an environment that conforms to the mava environment spec.
+        act_fn: a function that takes in params, timestep, key and optionally a state
+                and returns actions and optionally a state (see `EvalActFn`).
+        config: the systems config.
+        eval_episodes: how many episodes to evaluate the agents. At least this many episodes
```
OmaymaMahjoub: Maybe sending the config only instead of config and eval episodes as the eval episodes lives in the config

mava/evaluator.py
```diff
@@ -12,329 +12,140 @@
...
+
+            # find the first instance of done to get the metrics at that timestep we don't
+            # care about subsequent steps because we only the results from the first episode
+            done_idx = jnp.argmax(timesteps.last(), axis=0)
+            metrics = jax.tree_map(lambda m: m[done_idx, jnp.arange(config.arch.num_envs)], metrics)
+            del metrics["is_terminal_step"]  # uneeded for logging
+
+            return key, metrics
+
+        # todo: do we want to divide this by n devices?
```
OmaymaMahjoub: About this todo, I think yes we need to devide them over devices and then over num_envs

OmaymaMahjoub: I think we need to make a warning in case math.ceil(eval_episodes/num_envs) is not equal eval_episodes/num_envs

mava/evaluator.py
```diff
@@ -12,329 +12,140 @@
...
+        episode_loops = math.ceil(eval_episodes / config.arch.num_envs)
+        # This loop is important because we don't want too many parallel envs.
+        # So in evaluation we have config.arch.num_envs parallel envs and loop
+        # enough times so that we do at least `eval_episodes` number of episodes
+        _, metrics = jax.lax.scan(_episode, key, xs=None, length=episode_loops)
```
OmaymaMahjoub, line 155: We need to consider the case when eval_episodes<num_envs

mava/evaluator.py
```diff
@@ -12,329 +12,140 @@
...
+            return (env_state, ts, key, actor_state), ts
+
+        def _episode(key: PRNGKey, _: Any) -> Tuple[PRNGKey, Metrics]:
+            """Simulates `config.arch.num_envs` episodes."""
+            key, reset_key = jax.random.split(key)
+            reset_keys = jax.random.split(reset_key, config.arch.num_envs)
+            env_state, ts = jax.vmap(env.reset)(reset_keys)
+
+            step_state = env_state, ts, key, init_act_state
+            _, timesteps = jax.lax.scan(_env_step, step_state, jnp.arange(env.time_limit))
```
OmaymaMahjoub, line 138: how do you make sure the env didn't do an autorest or it gets wrong info after done equal true?

mava/evaluator.py
OmaymaMahjoub: Is the motivation for keep it like this because you want to keep same format for absolute metric?

OmaymaMahjoub (changes requested):
Thanks @sash-a  for the PR, I left some requests :pray:

mava/evaluator.py
sash-a: Yes that was the motivation, because otherwise we have an `if is_absolute_metric` and I thought this was a bit cleaner. But I think I'll refactor this a bit so that we pass that value into the `eval_fn` instead of the `make_eval_fn`. What do you think?

mava/evaluator.py
sash-a: Great question, I let the env run for `time_limit` even if it is passed done (I don't see a way around with jax), but I make sure I only take the metrics from the final step in the first episode. This line gets the index: `done_idx = jnp.argmax(timesteps.last(), axis=0)` and this uses that index on the metrics: `metrics = jax.tree_map(lambda m: m[done_idx, jnp.arange(config.arch.num_envs)], metrics)`.

mava/evaluator.py
sash-a: Ye I agree, although we weren't doing this before the refactor.

Agreed on the warning, I like the idea

mava/evaluator.py
sash-a: Agreed, I see two options: 
1. warning and do `num_envs` episodes
2. just do `eval_episdoes` parallel envs once

mava/evaluator.py
OmaymaMahjoub: or maybe make a flag to check if it is absolute metric so you use num_absolute_metric_eval_episodes instead of eval_episodes

mava/evaluator.py
OmaymaMahjoub: We were doing this before: `eval_batch = (config.arch.num_eval_episodes // n_devices) * eval_multiplier`

mava/evaluator.py
OmaymaMahjoub: I prefer the second because you know sometimes we do 64 parallel envs but we want to follow evaluation protocol by making only 32 evaluation episodes :pray:

mava/evaluator.py
OmaymaMahjoub: oh interesting, yeah the argmax will take the first time we have done=1, yeah nothing to worry about especially that all our envs have timelimit

mava/evaluator.py
sash-a: ah ok I missed that, then I must definitely divide by the devices

mava/evaluator.py
```diff
@@ -12,340 +12,179 @@
...
+
+        end_time = time.time()
+        total_timesteps = jnp.sum(metrics["episode_length"])
+        metrics["steps_per_second"] = total_timesteps / (end_time - start_time)
+        return metrics
+
+    return timed_eval_fn
+
+
+def make_ff_eval_act_fn(actor_network: nn.Module, config: DictConfig) -> EvalActFn:
```
RuanJohn: A general note here, I actually think we should make it so that this takes in an apply function instead of the network object. THis way we make things more flexible for the future.

sash-a: Done :+1:

OmaymaMahjoub (approved):
Thanks @sash-a :fire: :fire:

RuanJohn (approved):
Really nice clean up! Thank you @sash-a :fire:

PR #1085, opened 2024-07-08
feat: API for all Anakin Mava Envs

sash-a:
## What?
Add an API for Mava envs so that we can use type checking. We were previously duck typing and the extra attributes in the API were not enforced.
## Why?
Safer code, type checking and type hints.
## How?
Added a [protocol](https://peps.python.org/pep-0544/) which simply extends the jumanji env base class and add 3 attributes: `time_limit`, `action_dim` and `num_agents`. This is something all envs already had, but the variable names were not always consistent. We are now guaranteed that they will all have these variables.


mava/utils/make_env.py
```diff
@@ -65,25 +65,24 @@
...
+    train_env: MarlEnv, eval_env: MarlEnv, config: DictConfig
+) -> Tuple[MarlEnv, MarlEnv]:
     # Disable the AgentID wrapper if the environment has implicit agent IDs.
     config.system.add_agent_id = config.system.add_agent_id & (~config.env.implicit_agent_id)
 
-    # Add agent id to observation.
     if config.system.add_agent_id:
-        train_env = AgentIDWrapper(train_env)
-        eval_env = AgentIDWrapper(eval_env)
+        train_env = AgentIDWrapper(train_env)  # type: ignore
```
OmaymaMahjoub: too many `type: ignore` here is there a reason behind that

OmaymaMahjoub (changes requested):
Thanks for the PR! Just minor concerning about the mypy ignore comments

mava/wrappers/observation.py
```diff
@@ -28,8 +28,15 @@ class AgentIDWrapper(Wrapper):
     It can be useful in multi-agent environments where agents require unique identification.
     """
 
+    # This init isn't really needed as jumanji.Wrapper will forward the attributes,
+    # but mypy doesn't realize this.
```
OmaymaMahjoub, line 32: That is weird of mypy :(

OmaymaMahjoub (dismissed):
Thanks @sash-a :fire:

mava/utils/make_env.py
```diff
@@ -59,31 +60,30 @@
 # Define a different registry for Matrax since it has no generator.
 _matrax_registry = {"Matrax": MatraxWrapper}
 
-_jaxmarl_wrappers = {"Smax": SmaxWrapper, "MaBrax": MabraxWrapper}
+_jaxmarl_wrappers: Dict[str, Type[JaxMarlWrapper]] = {"Smax": SmaxWrapper, "MaBrax": MabraxWrapper}
```
WiemKhlifi: Is the typing here was required for jaxmarl? Can't we do similar thing to jumanji's wrappers?

mava/wrappers/observation.py
```diff
@@ -17,20 +17,26 @@
...
+from mava.types import MarlEnv, Observation, ObservationGlobalState, State
 
 
 class AgentIDWrapper(Wrapper):
     """A wrapper to add a one-hot vector as agent IDs to the original observation.
     It can be useful in multi-agent environments where agents require unique identification.
     """
 
-    def __init__(self, env: Environment):
+    # This init isn't really needed as jumanji.Wrapper will forward the attributes,
```
WiemKhlifi, line 31: Since this is already started to improve Mava's generalization can't we also add a wrapper like MavaEnvWrapper or MavaEnv class? This might fix these issues once and for all in the future and reduce redundancy :thinking:

I believe this will resemble what Jumanji is doing if it's not then it can be dealt with later 👀

mava/wrappers/jaxmarl.py
```diff
@@ -168,7 +168,7 @@ def __init__(
         self,
         env: MultiAgentEnv,
         has_global_state: bool,
-        timelimit: int,
+        timelimit: int = 1000,
```
WiemKhlifi: This won't be unified for Jaxmarl wrapper? (time_limit attribute instead)

WiemKhlifi (changes requested):
Thank you @sash-a  it's encouraging to see Mava expanding its support for a broader range of environments. Until the error we encountered is resolved, here are my comments. 🙏

mava/wrappers/jaxmarl.py
sash-a: Had to do this because all children of MultiAgentEnv don't require timelimit to be passed (it's optional), so I just supplied a default. I can make it `-1` and then add an error if it's `< 0`

mava/utils/make_env.py
sash-a: Unfortunately we need this for mypy and doing it for jumanji is not so simple, because it is `Dict[Dict[str, Union[Any, JumanjiMarlWrapper]]]` which causes other issues because of the `Any` which we need for the generator

mava/utils/make_env.py
```diff
@@ -56,34 +57,31 @@
     "Cleaner-v0": {"generator": CleanerRandomGenerator, "wrapper": CleanerWrapper},
 }
 
-# Define a different registry for Matrax since it has no generator.
+# Define a different registry for other suites since they have no generator.
```
WiemKhlifi: One final suggestion on this, could we rename the `_jaxmarl_wrappers -> _jaxmarl_registry` like the others and maybe change the comment to:
 `# Registry mapping environment names directly to the corresponding wrapper classes.`

mava/wrappers/jaxmarl.py
```diff
@@ -354,11 +348,6 @@ def state_size(self) -> chex.Array:
         "Get the sate size of the global observation"
         return self._env.state_size
 
-    @cached_property
```
WiemKhlifi, line 361: Much better 💃

WiemKhlifi (approved):
LGMT 🥳  one minor comment but you can ignore it 👀

mava/wrappers/jaxmarl.py
```diff
@@ -167,35 +167,34 @@ def __init__(
...
-
+        - time_limit: The time limit for each episode.
         """
         # Check that all specs are the same as we only support homogeneous environments, for now ;)
         homogenous_error = (
             f"Mava only supports environments with homogeneous agents, "
             f"but you tried to use {env} which is not homogeneous."
         )
         assert is_homogenous(env), homogenous_error
+        assert time_limit > 0, f"Time limit must be greater than 0, got {time_limit}"
```
OmaymaMahjoub, line 189: why can't we just leave time_limit a mendatory attribute to fill and by that it will automatically raise an error without this assert?

sash-a: Good question. If you look in `make_env.py` when we create a jaxmarl env we don't pass the `time_limit` because each env has a set time limit - for mabrax its 1000 and so we set this in the mabrax constructor when we call `super` but we don't want it to be required to pass this in to the mabrax wrapper. So we need to keep all the JaxMarlWrappers the same that's why we have a default here (to make it optional) but we want all the children to set the value, that's why we have the assert because the default is -1, does that make sense?

OmaymaMahjoub: Yeah totally makes sense :raised_hands: can you just add a comment based on what you said now so it will not raise confusion

OmaymaMahjoub (approved):
Thanks @sash-a :fire:

WiemKhlifi (approved):
LGTM thanks @sash-a 🔥

RuanJohn (approved):

PR #1087, opened 2024-07-09
fix: remove weak types to prevent double compilation

### LINKED ISSUE #1086: [BUG] Double compilation of learn_fn
### Describe the bug
The learn function compiles twice.

### To Reproduce
1. Run any system

### Expected behavior
A single compilation

### Solution
The reason for double compilation stems from the timestep object in the learner_state. The datatypes in the reset versus the step. Additionally, the data types in the specs all need to be consistent. The problem stems from possible multiple locations. First the episode metrics wrapper needs to change and have explicit jax types as the variables for tracking rewards etc. Secondly, there may be problems stemming from the environment wrappers if types are not consistent. It can be a little hard to debug where the problem is coming from since multiple wrappers all on top of each other can cause the inconsistencies in different places therefore even though you fix one wrapper, it still might double compile leading to confusion.

To ensure double compilation is fixed, for testing wrap the learn function in chex.assert_max_traces(learn, n=1), additionally if you want to rule out other problems, instead of changing the learner_state after learning with learner_output.learner_state, rather changing the data systematically with learner_state._replace(params=learner_output.learner_state.params). If you do this step by step you will see that timestep is the issue. 

good luck ;)

EdanToledo:
## What?
Ensure record episode metrics is consistent typing in the reset vs step
## Why?
Fixes double compilation occuring due to timestep object in learner_state
## How?
types are consistent.
## Extra
I've only tested this with lbf and matrax and it fixes their double compilation. I believe smax double compilation is not fixed and is most likely due to the actual environment state which you dont have control over, or its due to the wrapper. good luck :)

### DIFF
--- a/mava/wrappers/episode_metrics.py
+++ b/mava/wrappers/episode_metrics.py
@@ -49,11 +49,18 @@ class RecordEpisodeMetrics(Wrapper):
         """Reset the environment."""
         key, reset_key = jax.random.split(key)
         state, timestep = self._env.reset(reset_key)
-        state = RecordEpisodeMetricsState(state, key, 0.0, 0, 0.0, 0)
+        state = RecordEpisodeMetricsState(
+            state,
+            key,
+            jnp.array(0.0, dtype=float),
+            jnp.array(0, dtype=int),
+            jnp.array(0.0, dtype=float),
+            jnp.array(0, dtype=int),
+        )
         timestep.extras["episode_metrics"] = {
-            "episode_return": 0.0,
-            "episode_length": 0,
-            "is_terminal_step": False,
+            "episode_return": jnp.array(0.0, dtype=float),
+            "episode_length": jnp.array(0, dtype=int),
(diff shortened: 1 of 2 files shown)


OmaymaMahjoub (dismissed):
Thanks @EdanToledo :fire:  Your solution resolved the issue for most environments, but the smax env still has the problem as you said in the PR description. 
I suggest merging this PR and keeping the issue open for further investigation on smax 🙏

sash-a:
I fixed it as much as possible for jaxmarl. MaBrax doesn't double compile anymore, but SMAX still does as the `time` field in their env state is a weak type, so it's as good as we can do I think :smile:

sash-a (approved):
Thanks for fixing this @EdanToledo :fire:

OmaymaMahjoub (approved):
:fire:

EdanToledo:
since its been approved, im just gonna merge for the fun of it

PR #1089, opened 2024-07-16
fix: unpack array shape when reshaping minibatches in mappo training

RuanJohn:
## What?
A really small fix where we weren't unpacking the shape of the array when reshaping into minibatches in ff_mappo. Without this, this system won't run.


sash-a (approved):
Must have happened on the ruff refactor, my bad :facepalm:

liamclarkza (approved):
LGTM

PR #1088, opened 2024-07-10
Feat: sebulba ff_ippo 

Louay-Ben-nessir:
## What?
Implemented the ff_ippo system for sebulba
## Why?
Integrate Sebulba's architecture due to its effectiveness in scenarios involving non-jitted/non-jax environments.
## How?
By implementing the ff_ippo system.
## Extra
The other PPO systems will be added once this PR is merged.
This PR was built on top of #1080

PR #1080, opened 2024-06-10
feat: gym wrapper for sebulba 

Louay-Ben-nessir:
## What?
Implemented a wrapper for gym environments to facilitate their use in sebulba.
## Why?
Integrate Sebulba's architecture due to its effectiveness in scenarios involving non-jitted/non-jax environments.
## How?
Add a wrapper for gym envs, the corresponding `create_env` function, and the sebulba yaml file.

PR #1090, opened 2024-07-16
Chore: anakin and sebulba folders

Louay-Ben-nessir:
## What?
Changed the systems folder structure. 
## Why?
To create a folder for Sebulba systems. 
## How?
copied all of the Anakin systems into their own folder.
## Extra
This PR must be reviewed after #1080  and #1088


mava/systems/anakin/sac/ff_isac.py
sash-a, line 1: The same thing needs to be done to the hydra config path in the q learning and SAC files and also the import of their types

mava/utils/checkpointing.py
```diff
@@ -24,7 +24,7 @@
 from jax.tree_util import tree_map
 from omegaconf import DictConfig, OmegaConf
 
-from mava.systems.ppo.types import HiddenStates, Params
+from mava.systems.anakin.ppo.types import HiddenStates, Params
```
sash-a: We definitely shouldn't be using a PPO type here, but I guess that is not the concern of this PR.

sash-a (changes requested):
Please make the changes to the sac and q learning files for the hydra config and types. Double check that it runs :pray:

sash-a:
Just wondering if it makes sense to have the anakin/sebulba split one folder deeper so instead of `anakin/ppo` it'll be `ppo/anakin`. The reason I say this is to have a shared types files, because I assume a lot of the types we already will be useful to sebulba. @Louay-Ben-nessir have you used the types we already have like `HiddenStates`/`LearnerState` etc?

RuanJohn (approved):
Thank you @Louay-Ben-nessir!

sash-a (approved):
Thanks @Louay-Ben-nessir :fire:

PR #1095, opened 2024-07-24
chore: change all jax tree_map to tree.map

WiemKhlifi:
## What?
Change all instances of `jax.tree_map` and ` jax.tree_util.tree_map ` to` jax.tree.map` in the code base.

## Why?
This change aligns our code base with the updated JAX library and ensures compatibility with the latest version of JAX. 

## How?
Replace all calls to jax.tree_map with jax.tree.map.


sash-a (changes requested):
Seems like there's still some `tree_util` calls in ff ippo and one tree_map in `checkpointing.py`. Otherwise looks good, thanks @WiemKhlifi :fire:

RuanJohn (approved):
Thank you for this @WiemKhlifi :fire: 
I cannot find anywhere else that still needs to change :evergreen_tree:

sash-a (approved):
Thanks @WiemKhlifi! :ship:

PR #1096, opened 2024-08-05
fix: sac key splitting

sash-a:
## What?
We weren't splitting the keys in the SAC rollout loop


RuanJohn (approved):
Good catch :call_me_hand:

OmaymaMahjoub (approved):
Thanks @sash-a!

PR #1097, opened 2024-08-06
feat: add grad clipping to sac

sash-a:
## What?
Add grad clipping to sac
## Why?
PPO has it and it's useful :shrug:


WiemKhlifi (approved):
Thanks @sash-a 🎉

RuanJohn (approved):
LGTM

PR #1099, opened 2024-09-16
feat: add ability to arbitrarily chain torso configs together

EdanToledo:
## What?
Add ability to chain different torsos together based on configs.
## Why?
Allows construction of complicated networks through configs alone.
## How?
Add a new class of network and a constructer function called by hydra.


sash-a (approved):
LGTM, would be cool to just have a network which is and this is the base of all networks in mava:

```
class Network(nn.Module)
    torso: CompositeNetwork
    head: nn.Module

    def __call__(x, ...):
        x = do some stuff to the input  # each different network would do different stuff e.g centralized/decentralized etc
        return head(torso(x))
```

But we can leave this for another PR

sash-a:
Thanks for this @EdanToledo we've decided to not add this for now because we don't really have mava setup to use it and instead create a PR using this to overhaul the whole network module and make it more flexible using this at the core.

We're making it an issue for now and will get to it after our next research push

EdanToledo:
😴

PR #1101, opened 2024-10-09
refactor: update configs folder

### LINKED ISSUE #1033: [FEATURE]: Reduce config files

WiemKhlifi:
## What?
Updating the configs folder.

## How?
* Storing all defaults in one folder for Anakin systems.
* Removing all system_name_logger and use one base_logger.
* Update the config paths in systems & test files.


test/integration_test.py
```diff
@@ -23,6 +23,8 @@
 # system run all envs, but each env and each system is run at least once.
 # For each system we select a random environment to run.
 # Then for each environment we select a random system to run.
+anakin_config_path = "../mava/configs/default/anakin"
+
 ppo_systems = [
     "ppo.anakin.ff_ippo",
```
WiemKhlifi: Here I'm thinking we could do something like the below when we have all the systems for the sebulba arch else we do them separately:
```suggestion
arch="anakin"
config_path = f"../mava/configs/default/{arch}"
ppo_systems = [
    f"ppo.{arch}.ff_ippo",
```

sash-a (changes requested):
Thanks Wiem!

Couple small things:
1. Just put the default configs in `configs/default` (no anakin) anakin/sebulba will share the same configs except for arch to start. If we find that params differ a lot then we can change it later. Because of this I think you can leave the integration tests as is
2. I think remove the `default_...` part of the config file name because they're now in the default folder.

WiemKhlifi:
Thank you Sasha for these suggestions! Without the `default_` is definitely cleaner I made all the necessary changes please check them 🙏

mava/advanced_usage/ff_ippo_store_experience.py
OmaymaMahjoub, line 1: if you can add `cfg.logger.system_name = "ff_ippo"` to this one as well

OmaymaMahjoub (changes requested):
Thanks @WiemKhlifi  for the changes, I added minor request 🙏

mava/advanced_usage/ff_ippo_store_experience.py
WiemKhlifi: Nice catch thanks 🙌

mava/advanced_usage/ff_ippo_store_experience.py
```diff
@@ -670,11 +670,12 @@ def _reshape_experience(experience: Dict[str, chex.Array]) -> Dict[str, chex.Arr
     logger.stop()
 
 
-@hydra.main(config_path="../configs", config_name="default_ff_ippo.yaml", version_base="1.2")
+@hydra.main(config_path="../configs/default", config_name="ff_ippo.yaml", version_base="1.2")
```
OmaymaMahjoub, line 673: default/  you need to add "/"

WiemKhlifi: It works with or without but nice catch it makes all files the same 🙌

WiemKhlifi: Well I guess I'll do the opposite and remove the '/' from all files 😅

OmaymaMahjoub: oh I see!

OmaymaMahjoub (dismissed):
LGTM! 🔥

sash-a (dismissed):
Thanks Wiem! Looks great :medal_sports:

OmaymaMahjoub (approved):
Thanks @WiemKhlifi 🙏

sash-a (approved):

PR #1102, opened 2024-10-15
Refactor networks

RuanJohn:
## What?
This is really just a quality of life PR to help us as we merge MAT and Sable. This will allows us to have for example a `mava/networks/attention.py` file instead of having to further crowd `mava/networks.py`
## Why?
Future proofing merging all the new systems.
## Extra 
The heads (continuous, discrete and recurrent) we also now only take in an embedded observation array and the action mask instead of an embedded observation and a observation object. This allows us to use the current heads inside of MAT and Sable's networks.


mava/advanced_usage/ff_ippo_store_experience.py
```diff
@@ -31,8 +31,8 @@
 from rich.pretty import pprint
 
 from mava.evaluator import get_eval_fn, make_ff_eval_act_fn
-from mava.networks import FeedForwardActor as Actor
-from mava.networks import FeedForwardValueNet as Critic
+from mava.networks.base import FeedForwardActor as Actor
+from mava.networks.base import FeedForwardValueNet as Critic
```
sash-a, line 35: Nit: maybe import these into `networks/__init__.py` then you don't need to go `networks.base`

sash-a (dismissed):
:1st_place_medal:

mava/advanced_usage/ff_ippo_store_experience.py
RuanJohn: Done :)

sash-a (approved):
:pray:

OmaymaMahjoub (approved):
Thanks @RuanJohn  🔥

PR #1103, opened 2024-10-16
Feat: MPE wrapper

WiemKhlifi:
## What?
Added a wrapper for MPE environment.
## Why?
This change allows to run MPE environments within the Mava ecosystem by supporting both `discrete` and `continuous` action spaces.
## Extra
Currently supports a fully cooperative `SimpleSpreadMPE` task as the observation and action spaces are homogenous across agents. Further modifications are needed to support other versions.


test/integration_test.py
```diff
@@ -34,9 +34,9 @@
 q_learning_systems = ["q_learning.anakin.rec_iql"]
 sac_systems = ["sac.anakin.ff_isac", "sac.anakin.ff_masac"]
 
-discrete_envs = ["gigastep", "lbf", "matrax", "rware", "smax"]
+discrete_envs = ["gigastep", "lbf", "matrax", "rware", "smax", "mpe"]
 cnn_envs = ["cleaner", "connector"]
-continuous_envs = ["mabrax"]
+continuous_envs = ["mabrax", "mpe"]  # todo: test mpe only in cont or disc envs?
```
sash-a: I think test only cont and make the default in the config cont as this is how we use it in benchmarks, then you should be able to remove all the if `env == "mpe"` from this file :smile:

mava/wrappers/jaxmarl.py
```diff
@@ -252,6 +253,12 @@ def _create_observation(
         return Observation(**obs_data)
 
     def observation_spec(self) -> specs.Spec:
+        if isinstance(self._env, SimpleSpreadMPE):
+            obs, _ = self._env.reset(jax.random.PRNGKey(0))
+            # The shape provided in the `observation_spaces` isn't always the same as
+            # that given after constructing the full agent's obs.
+            self._env.observation_spaces["agent_0"].shape = obs["agent_0"].shape
```
sash-a: I think this is fixed in jaxmarl 0.0.5, if that is the case could we please update and remove this :pray:

mava/wrappers/jaxmarl.py
```diff
@@ -403,3 +410,42 @@ def get_global_state(self, wrapped_env_state: BraxState, obs: Dict[str, Array])
...
+        if _is_discrete(self._env.action_space(self.agents[0])):
+            return self._env.action_space(self.agents[0]).n
+        return self._env.action_space(self.agents[0]).shape[0]
+
+    @cached_property
+    def state_size(self) -> chex.Array:
+        "Get the state size of the global observation"
+        # The shape provided in the `observation_spaces` isn't always the same as
+        # that given after constructing the full agent's obs
+        obs, _ = self._env.reset(jax.random.PRNGKey(0))
```
sash-a: Same here for updating to 0.0.5

mava/wrappers/jaxmarl.py
```diff
@@ -403,3 +410,42 @@ def get_global_state(self, wrapped_env_state: BraxState, obs: Dict[str, Array])
...
+    def action_mask(self, wrapped_env_state: BraxState) -> Array:
+        """Get action mask for each agent."""
+        return jnp.ones((self.num_agents, self.action_dim), dtype=bool)
+
+    def get_global_state(self, wrapped_env_state: BraxState, obs: Dict[str, Array]) -> Array:
+        """Get global state from observation and copy it for each agent."""
+
+        all_obs = jnp.array([obs[agent] for agent in self._env.agents]).flatten()
+        all_obs = jnp.expand_dims(all_obs, axis=0).repeat(self._env.num_agents, axis=0)
+        return all_obs
```
sash-a: Just double checking do they provide a global state?

mava/configs/default/rec_mappo.yaml
```diff
@@ -3,7 +3,7 @@ defaults:
   - arch: anakin
   - system: ppo/rec_mappo
   - network: rnn  # [rnn, rcnn]
-  - env: rware  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax]
+  - env: lbf  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax, mpe]
```
sash-a: lets leave the default as rware

```suggestion
  - env: rware  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax, mpe]
```

mava/configs/default/rec_ippo.yaml
```diff
@@ -3,7 +3,7 @@ defaults:
   - arch: anakin
   - system: ppo/rec_ippo
   - network: rnn  # [rnn, rcnn]
-  - env: rware  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax]
+  - env: lbf  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax, mpe]
```
sash-a:
```suggestion
  - env: rware  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax, mpe]
```

sash-a (changes requested):
Looking really good, couple minor things :smile:

mava/wrappers/jaxmarl.py
WiemKhlifi: Nope and they used the classic method in this wrapper ([world_state](https://github.com/FLAIROx/JaxMARL/blob/42e7d638aa5dab88757c95c22f0486af8ec85694/baselines/MAPPO/mappo_rnn_mpe.py#L61)) for their MAPPO baseline.

mava/wrappers/jaxmarl.py
WiemKhlifi: Yes this is fixed in that version but we should handle other issues if we fix it to 0.0.5 like respecting a specific version of Jax/ Flax else we'll have a dependencies conflict issue and should handle this error: https://github.com/FLAIROx/JaxMARL/issues/108.
It works for me if jax==0.4.25 and flax should be forced to 0.8 since the one forced by jaxmarl 0.0.5 (0.7.6) gives the same previous issue 😢

sash-a: This really sucks. Ye I think they just haven't upgraded flax for their algorithms, which seems to be why they pin to 0.4.25, I'm not really sure what to do here, because I'd really like to pin to the latest jaxmarl and definitely latest jax. @RuanJohn thoughts?

RuanJohn: Since we are using the forked version of JaxMARL now, do we still want to update this?

requirements/requirements.txt
```diff
@@ -8,7 +8,7 @@ hydra-core==1.3.2
 id-marl-eval @ git+https://github.com/instadeepai/marl-eval
 jax==0.4.30
 jaxlib==0.4.30
-jaxmarl
+jaxmarl @ git+https://github.com/RuanJohn/JaxMARL@unpin-jax
```
sash-a: Just so we're clear we don't change anything else

```suggestion
jaxmarl @ git+https://github.com/RuanJohn/JaxMARL@unpin-jax  # only unpins the Jax version
```

mava/configs/default/ff_masac.yaml
```diff
@@ -3,8 +3,8 @@ defaults:
   - logger: logger
   - arch: anakin
   - system: sac/ff_masac
-  - network: continuous_mlp  # [continuous_mlp]
-  - env: mabrax  # [mabrax, mpe]
+  - network: mlp
+  - env: mabrax  # [mabrax]
```
sash-a: Why remove mpe here and in ISAC?

mava/configs/default/rec_iql.yaml
```diff
@@ -4,7 +4,7 @@ defaults:
   - arch: anakin
   - system: q_learning/rec_iql
   - network: rnn  # [rnn, rcnn]
-  - env: smax  # [cleaner, connector, gigastep, lbf, matrax, rware, smax]
+  - env: smax  # [cleaner, connector, gigastep, lbf, matrax, rware, smax, mpe]
```
sash-a: Can we only add it to cont systems, I know it can be discrete, but lets treat it as one of our continuous envs

sash-a (commented):
Thanks @WiemKhlifi just a few very small things left, otherwise I'm happy with this

mava/wrappers/jaxmarl.py
WiemKhlifi: Yesss I merged the new changes already 🚀

mava/configs/default/ff_masac.yaml
WiemKhlifi: This was by mistake I guess when I was solving merge conflicts from develop 🤦‍♀️

mava/configs/default/rec_iql.yaml
WiemKhlifi: Ah right the default action space is set to continuous 😅

mava/configs/env/mpe.yaml
```diff
@@ -0,0 +1,18 @@
...
+# This metric is returned at the end of an experiment and can be used for hyperparameter tuning.
+eval_metric: episode_return
+
+implicit_agent_id: False
+# Whether or not to log the winrate of this environment. This should not be changed as not all
+# environments have a winrate metric.
+log_win_rate: False
+
+kwargs:
+  action_type: Continuous # Whether agent action spaces are "Continuous" or "Discrete".
```
OmaymaMahjoub, line 19: Are we sure discrete one works? if not please say we only support continuous at the moment

mava/configs/default/rec_ippo.yaml
```diff
@@ -3,7 +3,7 @@ defaults:
   - arch: anakin
   - system: ppo/rec_ippo
   - network: rnn  # [rnn, rcnn]
-  - env: rware  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax]
+  - env: rware  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax, mpe]
```
OmaymaMahjoub, line 6: can you add mpe to all the algorithms, I see you add it just for rec systems, is there a reason for that?

mava/wrappers/jaxmarl.py
```diff
@@ -407,3 +408,37 @@ def get_global_state(self, wrapped_env_state: BraxState, obs: Dict[str, Array])
...
+        # Adjusted automatically based on the action_type specified in the kwargs.
+        if _is_discrete(self._env.action_space(self.agents[0])):
+            return self._env.action_space(self.agents[0]).n
+        return self._env.action_space(self.agents[0]).shape[0]
+
+    @cached_property
+    def state_size(self) -> chex.Array:
+        "Get the state size of the global observation"
+        return self._env.observation_space(self.agents[0]).shape[0] * self.num_agents
+
```
OmaymaMahjoub, line 435:
```suggestion

    def observation_spec(self) -> specs.Spec:
        """Returns the observation spec."""
        # JaxMARL `env.observation_spaces` don't give the correct shape for MPE with more than 3 agent.
        # We need to get the correct shape by creating a mock state and getting the shape from it.
        mock_state=self._env.reset(jax.random.PRNGKey(0))[0]
        observation_shape = mock_state["agent_0"].shape
        observation_spaces = self._env.observation_spaces
        for i in range(self.num_agents):
            observation_spaces[f"agent_{i}"].shape = observation_shape
        agents_view = jaxmarl_space_to_jumanji_spec(merge_space(observation_spaces))

        action_mask = specs.BoundedArray(
            (self.num_agents, self.action_dim), bool, False, True, "action_mask"
        )
        step_count = specs.BoundedArray(
            (self.num_agents,), jnp.int32, 0, self.time_limit, "step_count"
        )

        if self.has_global_state:
            global_state = specs.Array(
                (self.num_agents, observation_spaces["agent_0"].shape[0]*self.num_agents),
                agents_view.dtype,
                "global_state",
            )

            return specs.Spec(
                ObservationGlobalState,
                "ObservationSpec",
                agents_view=agents_view,
                action_mask=action_mask,
                global_state=global_state,
                step_count=step_count,
            )

        return specs.Spec(
            Observation,
            "ObservationSpec",
            agents_view=agents_view,
            action_mask=action_mask,
            step_count=step_count,
        )


```

OmaymaMahjoub: MPE with more than 3 agents is broken because of the original generating of obs space so I suggested this function to fix the issue

OmaymaMahjoub: I removed the "state_size" fn since we don't need it anymore

OmaymaMahjoub (changes requested):
Thank you Wiem for the work! I made a suggestion to fix the issue to when it comes to run mpe with 5 or 10 agents, if you can accept the changes or adapt them :)

mava/configs/env/mpe.yaml
WiemKhlifi: Yes it's working fine I set that to the default at the beginning but @sash-a advised to do the inverse since in the paper we used only cont mpe

mava/configs/default/rec_ippo.yaml
WiemKhlifi: It was by mistake when I merged the main all was changed

mava/wrappers/jaxmarl.py
WiemKhlifi: We no longer need to to get a `mock_state` and `reset` env since this is fixed in the new jaxmarl version that we agreed to be using 😄

WiemKhlifi: And I don't get it why we remove the `state_size`? it's needed in this case since we're still be using the same `obs_spec `of the main `JaxMarlWrapper ` 🤔

mava/configs/env/mpe.yaml
OmaymaMahjoub: Since it is working fine, that sounds good!

mava/configs/default/ff_sable.yaml
```diff
@@ -3,7 +3,7 @@ defaults:
   - arch: anakin
   - system: sable/ff_sable
   - network: ff_retention
-  - env: rware  # [cleaner, connector, gigastep, lbf, rware, smax]
+  - env: rware  # [cleaner, connector, gigastep, lbf, rware, smax, mpe]
```
WiemKhlifi, line 6: todo: If mpe willl be merged before sable cont remove mpe from here.

mava/wrappers/jaxmarl.py
sash-a: Agreed we shouldn't need this because we are now using the newest jaxmarl and they should have fixed this

test/integration_test.py
```diff
@@ -162,7 +162,7 @@ def test_discrete_cnn_env(fast_config: dict, env_name: str) -> None:
 @pytest.mark.parametrize("env_name", continuous_envs)
 def test_continuous_env(fast_config: dict, env_name: str) -> None:
     """Test all continuous envs on random systems."""
-    system_path = random.choice(ppo_systems + sac_systems)
+    system_path = random.choice(ppo_systems + sac_systems + transformer_systems)
```
sash-a: Honestly these systems take long enough, I think we can probably leave them out of here especially because `test_transformer_system` has cont in it
```suggestion
    system_path = random.choice(ppo_systems + sac_systems)
```

sash-a (commented):
One final thing, but I am happy

test/integration_test.py
WiemKhlifi: Yep I agree I'll revert it 🙌

mava/wrappers/jaxmarl.py
OmaymaMahjoub: Ah valid I tested it now 😅

OmaymaMahjoub (dismissed):
Thanks @WiemKhlifi

sash-a (dismissed):

OmaymaMahjoub (approved):
Thanks @WiemKhlifi

sash-a (approved):
:fire:

PR #1104, opened 2024-10-18
feat: set the action head automatically 

WiemKhlifi:
## What?
Refactored action head selection to use a dictionary lookup instead of manually setting for each env and network.
## Why?
This change eases the process of selecting the appropriate action head for each environment.
## How?
Added the `get_action_head` function as a utility and removed the `action_head` from all yaml configs.


mava/utils/network_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def get_action_head(env_name: str) -> dict:
+    """Returns the appropriate action head config based on the environment name."""
+    action_head = action_head_per_env.get(env_name)
+
+    if action_head is None:
+        raise ValueError(f"Environment {env_name} is not recognized.")
+
+    return {"_target_": action_head}
```
sash-a: Could we rather do this with action specs :pray: 
So if action spec is a discrete/multi-discrete then return the discrete action head, otherwise the continuous action head

sash-a (changes requested):
Swap to using action specs instead of manually making the dictionary of env -> head :smile:

mava/utils/network_utils.py
WiemKhlifi: I got your point yeah that's much better I'll push the changes right away 🙌

mava/utils/network_utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+from mava.types import MarlEnv
+
+
+def get_action_head(env: MarlEnv) -> Dict[str, str]:
+    """Returns the appropriate action head config based on the environment action_spec."""
+    if isinstance(env.action_spec(), (DiscreteArray, MultiDiscreteArray)):
+        return {"_target_": "mava.networks.heads.DiscreteActionHead"}
+
+    return {"_target_": "mava.networks.heads.ContinuousActionHead"}
```
RuanJohn:
```suggestion
def get_action_head(env: MarlEnv) -> Dict[str, str], str:
    """Returns the appropriate action head config based on the environment action_spec."""
    if isinstance(env.action_spec(), (DiscreteArray, MultiDiscreteArray)):
        return {"_target_": "mava.networks.heads.DiscreteActionHead"}, DISCRETE

    return {"_target_": "mava.networks.heads.ContinuousActionHead"}, CONTINUOUS
```

mava/utils/network_utils.py
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Dict
+
+from jumanji.specs import DiscreteArray, MultiDiscreteArray
+
+from mava.types import MarlEnv
+
+
```
RuanJohn, line 23:
```suggestion
DISCRETE = "discrete"
CONTINUOUS = "continuous"

```

RuanJohn (changes requested):
This is a general suggestion so that we can also use this function for the transformer based systems which don't need a head config but need an action space type. 

You'd just also need to update the system run files to accommodate for this.

RuanJohn (approved):
Looks good! Thank you @WiemKhlifi :raised_hands:

sash-a (approved):
:fire:

PR #1105, opened 2024-10-21
feat: hasac

sash-a:
## What?
Heterogeneous agent soft actor critic: https://arxiv.org/pdf/2306.10715

Also moved the `step` function is all the SAC methods to a more sensible place


sash-a:
Easier to review the diff between masac and hasac: https://www.diffchecker.com/EKMo2FcC/

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,708 @@
...
+    actor_network = Actor(actor_torso, actor_action_head)
+    actor_params = jax.vmap(actor_network.init, in_axes=(0, None))(actor_keys, obs_single_batched)
+
+    # Making Q networks
+    critic_torso = hydra.utils.instantiate(cfg.network.critic_network.pre_torso)
+    q_network = QNetwork(critic_torso, centralised_critic=True)
+    q1_params = q_network.init(q1_key, obs_single_batched, concat_acts_batched)
+    q2_params = q_network.init(q2_key, obs_single_batched, concat_acts_batched)
+    q1_target_params = q_network.init(q1_target_key, obs_single_batched, concat_acts_batched)
+    q2_target_params = q_network.init(q2_target_key, obs_single_batched, concat_acts_batched)
```
RuanJohn: Just adding a note here to double check whether the critics should condition on a single agent's observation and the joint observation along with all the actions.

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,708 @@
...
+from mava.wrappers import episode_metrics
+
+
+# It is faster to do this with a vmap, but unfortunately that requires switching to numpyro.
+# This requires a lot of testing so there is currently an issue for it: #1098
+def get_actions(
+    actor_params: FrozenVariableDict,
+    actor_net: Actor,
+    keys: chex.PRNGKey,
+    env: MarlEnv,
```
RuanJohn: Passing in env just to get the number of agents seems a bit much. Can't we just set num_agents in the config and use it from there or something along those lines.

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,708 @@
...
+        params: SacParams, opt_states: OptStates, data: Transition, key: chex.PRNGKey
+    ) -> Tuple[SacParams, OptStates, Metrics]:
+        """Update the Q parameters."""
+        # Calculate Q target values.
+        act_keys = jax.random.split(key, env.num_agents)
+        next_action, next_log_prob = get_actions(
+            params.actor, actor_net, act_keys, env, data.next_obs
+        )
+
+        joint_next_actions = get_joint_action(next_action)
```
RuanJohn: A general note, but I think adding shapes in the comments could be super useful here.

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,708 @@
...
+    def update_actor_and_alpha(
+        params: SacParams, opt_states: OptStates, data: Transition, key: chex.PRNGKey
+    ) -> Tuple[SacParams, OptStates, Metrics]:
+        """Update the actor and alpha parameters. Compensated for the delay in policy updates."""
+        alpha_grad_fn = jax.value_and_grad(alpha_loss_fn)
+        actor_grad_fn = jax.value_and_grad(actor_loss_fn)
+
+        # compensate for the delay by doing `policy_frequency` updates instead of 1.
+        assert cfg.system.policy_update_delay > 0, "Need to have a policy update delay > 0."
+        for _ in range(cfg.system.policy_update_delay):
```
RuanJohn, line 390: Is it standard to do multiple policy updates after delaying like this?

RuanJohn (changes requested):
Overall looking very nice! 
Something that would be great is to add some comments on the shapes. And then just a questions and suggestions here and there.

mava/systems/sac/anakin/ff_hasac.py
sash-a: Ye it's a TD3 thing that people found useful in SAC, CleanRL does it [here](https://github.com/vwxyzjn/cleanrl/blob/master/cleanrl/sac_continuous_action.py#L265)

mava/systems/sac/anakin/ff_hasac.py
sash-a: So I just had a look and this actually will give us the state. The env is created with `global_state=True` always and so the critic will always pull out the global in favour of the obs.

Sorry for the confusion yesterday

RuanJohn: Ah okay, makes sense. Thanks for clarifying :raised_hands: 

Can you just add a comment here, that the obs object contains the global state please?

mava/systems/sac/anakin/ff_hasac.py
RuanJohn: Perfect, thanks for clarifying!

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,717 @@
...
+            key, act_key, agent_order_key = jax.random.split(key, 3)
+            act_keys = jax.random.split(act_key, env.num_agents)
+            if cfg.system.shuffle_agents:
+                agent_ids = jax.random.permutation(agent_order_key, env.num_agents)
+            else:
+                agent_ids = jnp.arange(env.num_agents)
+
+            joint_actions, log_probs = get_actions(
```
RuanJohn, line 400: Shapes for joint_actions and log_probs here would be great.

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,717 @@
...
+            )
+
+            # HASAC sequential update: run the normal actor update one at a time instead of batched.
+            # Update the joint actions after updating the actor and use the new joint actions.
+            for agent_id in agent_ids:
+                key, actor_key = jax.random.split(key)
+
+                agent_params = tree_slice(params.actor, agent_id)
+                agent_opt_state = tree_slice(opt_states.actor, agent_id)
+                agent_obs = tree_slice(data.obs, jnp.s_[:, agent_id])
```
RuanJohn, line 414: Similarly, a shape here would be nice. Also what is `jnp.s_`? I haven't seen it before.

RuanJohn (commented):
Just two other places where I think a shape would be nice. And then a question on `jnp.s_`

mava/systems/sac/anakin/ff_hasac.py
sash-a: Good question, it returns what's called an index expression, which is basically a way to pass an index eg `[:, 1]` around as a variable. So here it's convenient to pass it to `tree_slice` instead of writing `tree.map(lambda x: x[:, agent_id], data.obs)`

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,727 @@
...
+
+    full_action_shape = (cfg.arch.num_envs, *env.action_spec().shape)
+
+    # losses:
+    def q_loss_fn(
+        q_params: QVals, obs: Array, action: Array, target: Array
+    ) -> Tuple[Array, Metrics]:
+        q1_params, q2_params = q_params
+        # Concat all actions and tile them for num agents to create joint actions for all agents
+        joint_action = get_joint_action(action)  # (B, A, Act) -> (B, A, A * Act)
```
RuanJohn: Just a note to please update all the shape comments. Can we please do this everywhere? 
```suggestion
        joint_action = get_joint_action(action)  # (B, N, A) -> (N, N, N * A)
```

RuanJohn (changes requested):
Last minor nits on the shape comments.

RuanJohn (dismissed):
This is amazing. Great work @sash-a :fire:

mava/configs/arch/anakin.yaml
```diff
@@ -1,13 +1,13 @@
 # --- Anakin config ---
 
 # --- Training ---
-num_envs: 16  # Number of vectorised environments per device.
+num_envs: 32  # Number of vectorised environments per device.
```
OmaymaMahjoub: can you revert the changes here to keep it to the default 16 envs

mava/utils/jax_utils.py
```diff
@@ -14,13 +14,26 @@
...
 import jax
 import jax.numpy as jnp
 import numpy as np
 from jax import tree
+from typing_extensions import TypeAlias
+
+Indexer: TypeAlias = Union[int, slice, Tuple[slice, ...]]
+
+
+def tree_slice(pytree: chex.ArrayTree, i: Indexer) -> chex.ArrayTree:
```
OmaymaMahjoub, line 30: can you add documentation here to explain the need for this fn

mava/utils/jax_utils.py
```diff
@@ -14,13 +14,26 @@
...
+    return tree.map(lambda x: x[i], pytree)
+
+
+def tree_at_set(old_tree: chex.ArrayTree, i: Indexer, new_tree: chex.ArrayTree) -> chex.ArrayTree:
```
OmaymaMahjoub, line 35: same for this util function

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,728 @@
...
+def hydra_entry_point(cfg: DictConfig) -> float:
+    """Experiment entry point."""
+    # Allow dynamic attributes.
+    OmegaConf.set_struct(cfg, False)
+    cfg.logger.system_name = "ff_hasac"
+
+    # Run experiment.
+    final_return = run_experiment(cfg)
+
+    print(f"{Fore.CYAN}{Style.BRIGHT}MASAC experiment completed{Style.RESET_ALL}")
```
OmaymaMahjoub:
```suggestion
    print(f"{Fore.CYAN}{Style.BRIGHT}HASAC experiment completed{Style.RESET_ALL}")
```

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,728 @@
...
+    obs_single_batched = tree.map(lambda x: x[0][jnp.newaxis, ...], obs)
+
+    # Making actor network
+    actor_torso = hydra.utils.instantiate(cfg.network.actor_network.pre_torso)
+    action_head, _ = get_action_head(env)
+    actor_action_head = hydra.utils.instantiate(
+        action_head, action_dim=env.action_dim, independent_std=False
+    )
+    actor_network = Actor(actor_torso, actor_action_head)
+    actor_params = jax.vmap(actor_network.init, in_axes=(0, None))(actor_keys, obs_single_batched)
```
OmaymaMahjoub, line 162: if you can add comment here to explain that the vmap used to produce different params per agent

OmaymaMahjoub (changes requested):
Looks good from my side, just it lacks some  documentations here and there to make it more clear and easy for users to follow the difference between hasac and the other sac systems

mava/systems/sac/anakin/ff_hasac.py
```diff
@@ -0,0 +1,728 @@
...
+    key, eval_key = jax.random.split(key)
+
+    def eval_act_fn(
+        params: FrozenDict, timestep: TimeStep, key: chex.PRNGKey, actor_state: ActorState
+    ) -> Tuple[Action, Dict]:
+        keys = jax.random.split(key, eval_env.num_agents)
+        action, _ = get_actions(
+            params, actor, keys, eval_env.num_agents, eval_env.action_dim, timestep.observation
+        )
+        return action, {}
```
OmaymaMahjoub, line 625: why don't we do similar to ppo systems (not sure if it is the same for other sac systems), we import make_act_sac_fn from the evaluator and here we give it get actions as input

sash-a: Would be nice but it's not possible here because `get_actions` returns the actions and that function expects distributions. It would be possible if we had vmappable distributions, but I think that is for another PR.

RuanJohn (approved):
:fire:

OmaymaMahjoub (approved):
🚀

mava/systems/sac/anakin/ff_hasac.py
OmaymaMahjoub: we decided to not do this changes, since the eval_act_fn only used for HASAC so keeping it on the system side is fine

PR #1106, opened 2024-10-22
Add QMIX

RuanJohn:
## What?
Adds QMIX to Mava. Massive shoutout to @lbeyers and @jcformanek for all the work. This is just a refactor to get things to match the current Mava style! :fire: 
## Extra
* QMIX requires a more unique network config than other systems, so instead of further cluttering `mava/configs/network/rnn.yaml` I instead added `mava/configs/network/qmix_rnn.yaml`.
* The QMIX Mixer network also requires an MLP that doesn't have the outputs activated. I updated the `MLPTorso` to able to accommodate for this while keeping the default behaviour of the current `MLPTorso`.
* It has been benchmarked against the results from the Sable paper and everything checks out. Results are [here](https://app.neptune.ai/o/InstaDeep/org/Mava/runs/table?viewId=9d4c7d01-0183-496b-abcd-c27b1ac75cdd&detailsTab=charts&dash=table&type=run).
* Here is a [diff](https://www.diffchecker.com/IJWVHzfn/) between `rec_iql.py` and `rec_qmix.py` to hopefully make the reviewing easier.


mava/configs/env/smax.yaml
```diff
@@ -7,7 +7,7 @@ env_name: Smax  # Used for logging purposes and selection of the corresponding w
 # 2s3z | 3s5z | 5m_vs_6m | 10m_vs_11m | 27m_vs_30m | 3s5z_vs_3s6z | 3s_vs_5z | 6h_vs_8z | smacv2_5_units | smacv2_10_units | smacv2_20_units
 scenario:
   name: HeuristicEnemySMAX
-  task_name: 2s3z
+  task_name: 5m_vs_6m
```
sash-a:
```suggestion
  task_name: 2s3z
```

mava/configs/system/q_learning/rec_qmix.yaml
```diff
@@ -0,0 +1,35 @@
+# --- Defaults REC-QMIX ---
+total_timesteps: 1e7 # Set the total environment steps.
+# If unspecified, it's derived from num_updates; otherwise, num_updates adjusts based on this value.
+num_updates: 156250 # Number of updates.
```
sash-a: Some suggestions to bring this more in line with the less timesteps but more updates idea, feel free to reject these

```suggestion
total_timesteps: ~ # Set the total environment steps.
# If unspecified, it's derived from num_updates; otherwise, num_updates adjusts based on this value.
num_updates: 10000 # Number of updates.
```

mava/configs/system/q_learning/rec_qmix.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+# --- RL hyperparameters ---
+min_buffer_size: 32
+update_batch_size: 1 # Number of vectorised gradient updates per device.
+
+rollout_length: 2 # Number of environment steps per vectorised environment.
+epochs: 2 # Number of learn epochs per training data batch.
+
+# sizes
+buffer_size: 1000 # size of the replay buffer. Note: total size is this * num_devices
```
sash-a:
```suggestion
rollout_length: 8 # Number of environment steps per vectorised environment.
epochs: 32 # Number of learn epochs per training data batch.

# sizes
buffer_size: 10000 # size of the replay buffer. Note: total size is this * num_devices
```

mava/configs/system/q_learning/rec_qmix.yaml
```diff
@@ -0,0 +1,35 @@
...
+sample_batch_size: 128  # size of training data batch sampled from the buffer
+sample_sequence_length: 20 # 21 transitions are sampled, giving 20 complete data points
```
sash-a: Not sure if 21 is correct here, because you got rid of the +1 in the flashbax init

mava/configs/system/q_learning/rec_qmix.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+# learning rates
+q_lr: 3e-5  # the learning rate of the Q network network optimizer
+max_grad_norm: 10 # value used to clip optimiser - set big for no clipping
+
+# other
```
sash-a:
```suggestion
# Q Learning related
```

mava/networks/base.py
```diff
@@ -232,3 +233,73 @@ def __call__(
         eps_greedy_dist = MaskedEpsGreedyDistribution(q_values, eps, obs.action_mask)
 
         return hidden_state, eps_greedy_dist
+
+
+class QMixNetwork(nn.Module):
```
sash-a: I feel we should call this a mixing network, either `QMixingNetwork` or just `MixingNetwork` because at first glance it seems like the network we use in qmix for everything

mava/networks/torsos.py
```diff
@@ -35,11 +36,16 @@ def setup(self) -> None:
...
+        for i, layer_size in enumerate(self.layer_sizes):
             x = nn.Dense(layer_size, kernel_init=orthogonal(np.sqrt(2)))(x)
             if self.use_layer_norm:
                 x = nn.LayerNorm(use_scale=False)(x)
-            x = self.activation_fn(x)
+
+            if i != len(self.layer_sizes) - 1:
+                x = self.activation_fn(x)
+            elif i == len(self.layer_sizes) - 1 and self.activate_final:
+                x = self.activation_fn(x)
```
sash-a: Little cleaner I think
```suggestion
            should_activate = (i < len(self.layer_sizes) - 1) or self.activate_final
            x = self.activation_fn(x) if should_activate else x
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+    q_target_params = q_net.init(q_key, init_hidden_state, init_x)
+
+    # Make Mixer Network
+    dummy_agent_qs = jnp.zeros(
+        (
+            cfg.system.sample_batch_size,
+            cfg.system.sample_sequence_length - 1,
+            num_agents,
+        ),
+        "float32",
```
sash-a:
```suggestion
        dtype=float,
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+    global_env_state_shape = (
+        env.observation_spec().generate_value().global_state[0, :].shape
+    )  # NOTE: Env wrapper currently duplicates env state for each agent
+    dummy_global_env_state = jnp.zeros(
+        (
+            cfg.system.sample_batch_size,
+            cfg.system.sample_sequence_length - 1,
+            *global_env_state_shape,
+        ),
+        "float32",
```
sash-a:
```suggestion
        dtype=float,
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+    Tuple[MarlEnv, MarlEnv],
+    RecQNetwork,
+    QMixNetwork,
+    optax.GradientTransformation,
+    TrajectoryBuffer,
+    LearnerState,
+    MavaLogger,
+    chex.PRNGKey,
+]:
+    """Initialize system by creating the envs, networks etc."""
```
sash-a, line 72: Not too fussed if you remove this docstring, but if you remove it from here then please do the same for IQL

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+    # Distribute params, opt states and hidden states across all devices
+    params = replicate(params)
+    opt_state = replicate(opt_state)
+    init_hidden_state = replicate(init_hidden_state)
+
+    init_acts = env.action_spec().generate_value()
+
+    # NOTE: Term_or_trunc refers to the the joint done, ie. when all agents are done or when the
+    # episode has terminated. We use this exclusively in QMIX. Terminal refers to individual agent
+    # dones. We keep this in this file for consistency with IQL.
```
sash-a: nit
```suggestion
    # NOTE: term_or_trunc refers to the the joint done, ie. when all agents are done or when the
    # episode horizon has been reached. We use this exclusively in QMIX. 
    # Terminal refers to individual agent dones. We keep this here for consistency with IQL.
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        reward=jnp.zeros((1,), dtype=float),
+        terminal=jnp.zeros((1,), dtype=bool),
+        term_or_trunc=jnp.zeros((1,), dtype=bool),
+        next_obs=init_obs,
+    )
+
+    # Initialise trajectory buffer
+    rb = fbx.make_trajectory_buffer(
+        # n transitions gives n-1 full data points
+        sample_sequence_length=cfg.system.sample_sequence_length,
```
sash-a, line 179: For consistency either make this `cfg.system.sample_sequence_length` or change IQL also there's a mention of this in the docs

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+
+
+def make_update_fns(
+    cfg: DictConfig,
+    env: MarlEnv,
+    q_net: RecQNetwork,
+    mixer: QMixNetwork,
+    opt: optax.GradientTransformation,
+    rb: TrajectoryBuffer,
+) -> Any:
```
sash-a: Can we add a return type, I think it should be the same/similar to IQL

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        params,
+        first_keys,
+    )
+
+    return (env, eval_env), q_net, q_mixer, opt, rb, learner_state, logger, key
+
+
+def make_update_fns(
+    cfg: DictConfig,
+    env: MarlEnv,
```
sash-a, line 236: We seem to use the old Jumanji Environment type in IQL, can you update it to use `MarlEnv` :pray:

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+    mixer: QMixNetwork,
+    opt: optax.GradientTransformation,
+    rb: TrajectoryBuffer,
+) -> Any:
+    def select_eps_greedy_action(
+        action_selection_state: ActionSelectionState,
+        obs: Observation,
+        term_or_trunc: Array,
+    ) -> Tuple[ActionSelectionState, Array]:
+        """Select action to take in eps-greedy way. Batch and agent dims are included."""
```
sash-a, line 247: I see you removed the docstring from here, I agree with it, can you just do the same for IQL :pray:

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        )
+
+        obs = tree.map(lambda x: x[jnp.newaxis, ...], obs)
+        term_or_trunc = tree.map(lambda x: x[jnp.newaxis, ...], term_or_trunc)
+
+        next_hidden_state, eps_greedy_dist = q_net.apply(
+            params, hidden_state, (obs, term_or_trunc), eps
+        )
+
+        # Random key splitting
```
sash-a: No need for this comment, code is self evident
```suggestion
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        next_action_selection_state, action = select_eps_greedy_action(
+            action_selection_state, obs, term_or_trunc
+        )
+
+        next_env_state, next_timestep = jax.vmap(env.step)(env_state, action)
+
+        # Get reward
+        reward = jnp.mean(
+            next_timestep.reward, axis=-1, keepdims=True
+        )  # NOTE: Combine agent rewards, since QMIX is cooperative.
```
sash-a: nit
```suggestion
        # NOTE: Combine agent rewards, since QMIX is cooperative.
        reward = jnp.mean(next_timestep.reward, axis=-1, keepdims=True)
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        """
+        hidden_state = ScannedRNN.initialize_carry(
+            (cfg.system.sample_batch_size, obs.agents_view.shape[2]), cfg.network.hidden_state_dim
+        )
+        # the rb outputs (B, T, ... ) the RNN takes in (T, B, ...)
+        obs = switch_leading_axes(obs)  # (B, T) -> (T, B)
+        term_or_trunc = switch_leading_axes(term_or_trunc)  # (B, T) -> (T, B)
+        obs_term_or_trunc = (obs, term_or_trunc)
+
+        return hidden_state, obs_term_or_trunc
```
sash-a, line 328: Could move this to `utils/rnn` or `utils/qlearning` because it's used in both Q learning systems

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        q_online = switch_leading_axes(q_online)  # (T, B, ...) -> (B, T, ...)
+        # Get the q values of the taken actions and remove extra dim
+        q_online = jnp.squeeze(
+            jnp.take_along_axis(q_online, action[..., jnp.newaxis], axis=-1), axis=-1
+        )
+
+        q_online = mixer.apply(
+            online_mixer_params, q_online, obs.global_state[:, :, 0, ...]
+        )  # B,T,A,... -> B,T,1,...
+        # NOTE: States are replicated over agents so we take only take first one
```
sash-a: nit
```suggestion
        # NOTE: States are replicated over agents so we take only take first one
        q_online = mixer.apply(
            online_mixer_params, q_online, obs.global_state[:, :, 0, ...]
        )  # B,T,A,... -> B,T,1,...
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+
+        q_loss = jnp.mean((q_online - target) ** 2)
+
+        loss_info = {
+            "q_loss": q_loss,
+            "mean_q": jnp.mean(q_online),
+            "max_q_error": jnp.max(jnp.abs(q_online - target) ** 2),
+            "min_q_error": jnp.min(jnp.abs(q_online - target) ** 2),
+            "mean_target": jnp.mean(target),
```
sash-a: Doing this calculation twice, can you store it in an error variable and then min/max that :pray:

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+
+        return q_loss, loss_info
+
+    def update_q(
+        params: QMIXParams, opt_states: optax.OptState, data: Transition, t_train: int
+    ) -> Tuple[QMIXParams, optax.OptState, Metrics]:
+        """Update the Q parameters."""
+
+        # Get data aligned with current/next timestep
+        data_first = tree.map(lambda x: x[:, :-1, ...], data)  # (B, T, ...)
```
sash-a: Not a fan of `_first` here, I think we should just have `data` and `data_next` first implies it's only the first one. Similarly for `first_reward`

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+
+        first_reward = data_first.reward
+        next_done = data_next.term_or_trunc
+
+        # Get the greedy action using the distribution.
+        # Epsilon defaults to 0.
+        hidden_state, next_obs_term_or_trunc = prep_inputs_to_scannedrnn(
+            data.obs,
+            data.term_or_trunc,
+        )  # (T, B, ...)
```
sash-a: nit
```suggestion
        hidden_state, next_obs_term_or_trunc = prep_inputs_to_scannedrnn(
            data.obs, data.term_or_trunc
        )  # (T, B, ...)
```

mava/systems/q_learning/types.py
```diff
@@ -90,11 +90,21 @@ class ActionState(NamedTuple):
...
 
-class TrainState(NamedTuple):
+class QMIXParams(NamedTuple):
+    online: FrozenVariableDict
+    target: FrozenVariableDict
+    mixer_online: FrozenVariableDict
+    mixer_target: FrozenVariableDict
+
+
+QLearningParams = TypeVar("QLearningParams", QNetParams, QMIXParams)
```
sash-a, line 79: Can we also parameterize the Learner state with this :pray:

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+            final_action_state.term_or_trunc,
+            final_action_state.action_selection_state.hidden_state,
+            final_action_state.env_state,
+            final_action_state.action_selection_state.time_steps,
+            final_train_state.train_steps,
+            final_train_state.opt_state,
+            final_action_state.buffer_state,
+            final_train_state.params,
+            new_key,
+        )
```
sash-a, line 531: Seems to me like it would make a lot of sense to make the `LearnerState` contain the `ActionSelectionState`, because a lot of these params are just from the act state.

Possibly out of scope for this PR though

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+            term_or_trunc,
+            hidden_state,
+            env_state,
+            time_steps,
+            train_steps,
+            opt_state,
+            buffer_state,
+            params,
+            key,
+        ) = learner_state
```
sash-a, line 501: Half of this function is packing and unpacking params :joy:

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        final_action_state, metrics = scanned_act(action_state)
+
+        # Sample and learn
+        train_state = TrainState(
+            final_action_state.buffer_state,
+            params,
+            opt_state,
+            train_steps,
+            train_key,
+        )
```
sash-a: nit formatting to align with IQL
```suggestion
        train_state = TrainState(
            final_action_state.buffer_state, params, opt_state, train_steps, train_key
        )
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+    ) -> Tuple[chex.Array, ActorState]:
+        """The acting function that get's passed to the evaluator.
+        A custom function is needed for epsilon-greedy acting.
+        """
+        hidden_state = actor_state["hidden_state"]
+
+        term_or_trunc = timestep.last()
+        net_input = (timestep.observation, term_or_trunc[..., jnp.newaxis])
+        net_input = tree.map(lambda x: x[jnp.newaxis], net_input)  # add batch dim to obs
+        next_hidden_state, eps_greedy_dist = q_net.apply(params, hidden_state, net_input)
```
sash-a, line 581: You don't put 0 in here, but it's there in IQL, can you remove it from IQL :pray:

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+
+    # Create an initial hidden state used for resetting memory for evaluation
+    eval_batch_size = get_num_eval_envs(cfg, absolute_metric=False)
+    eval_hs = ScannedRNN.initialize_carry(
+        (jax.device_count(), eval_batch_size, cfg.system.num_agents),
+        cfg.network.hidden_state_dim,
+    )
+
+    max_episode_return = -jnp.inf
+    best_params = copy.deepcopy(unreplicate_batch_dim(learner_state.params.online))
```
sash-a, line 602: For some reason the diff showed this wasn't there in IQL, can you please add it :pray:

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        )
+        final_metrics, ep_completed = episode_metrics.get_final_step_metrics(metrics)
+        final_metrics["steps_per_second"] = steps_per_rollout / elapsed_time
+        loss_metrics = losses
+        logger.log(
+            {"timestep": t, "epsilon": eps},
+            t,
+            eval_idx,
+            LogEvent.MISC,
+        )
```
sash-a: nit formatting to align with IQL
```suggestion
        logger.log({"timestep": t, "epsilon": eps}, t, eval_idx, LogEvent.MISC)
```

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+        if ep_completed:
+            logger.log(final_metrics, t, eval_idx, LogEvent.ACT)
+        logger.log(loss_metrics, t, eval_idx, LogEvent.TRAIN)
+
+        # Evaluate:
+        key, eval_key = jax.random.split(key)
+        eval_keys = jax.random.split(eval_key, cfg.arch.n_devices)
+        eval_params = unreplicate_batch_dim(learner_state.params.online)
+        eval_metrics = evaluator(eval_params, eval_keys, {"hidden_state": eval_hs})
+        jax.block_until_ready(eval_metrics)
```
sash-a, line 631: Please add to IQL

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,697 @@
...
+            cfg.network.hidden_state_dim,
+        )
+
+        abs_metric_evaluator = get_eval_fn(eval_env, eval_act_fn, cfg, absolute_metric=True)
+        eval_metrics = abs_metric_evaluator(best_params, eval_keys, {"hidden_state": eval_hs})
+        logger.log(eval_metrics, t, eval_idx, LogEvent.ABSOLUTE)
+
+    logger.stop()
+
+    return float(episode_return)
```
sash-a: We should be returning eval performance like we do in other systems

sash-a (changes requested):
Thanks for this guys, it looks great! Lot of minor nits, tried to suggest them where I could. 2 slightly bigger things we need:

Please add a test, should be as simple as adding the qmix path in the test file :pray: 
Can you rename all "interact" stuff in IQL to just "act" like you have it here

mava/configs/env/smax.yaml
RuanJohn: Is there a reason you are updating all the configs? Right now it has the optimal configs for 5m_vs_6m which is quite nice in my opinion. Although it will be slower.

mava/systems/q_learning/anakin/rec_qmix.py
RuanJohn: I saw the comment, but opted not to change any of this since I thought this is probably logic from flashbax. @lbeyers can you please comment on this?

mava/configs/system/q_learning/rec_qmix.yaml
RuanJohn: @lbeyers can you please help here?

mava/systems/q_learning/anakin/rec_qmix.py
RuanJohn: Agree, it is out of scope. Can you please make an issue?

mava/configs/env/smax.yaml
sash-a: I just did it because I thought 2s3z was our default, happy to keep it as 5m_vs_6m :smile:

mava/systems/q_learning/anakin/rec_qmix.py
RuanJohn: It's true :sweat_smile:

mava/systems/q_learning/anakin/rec_qmix.py
RuanJohn: Sounds good. Removed. I felt like what was happening was explicit enough given the type hints.

mava/systems/q_learning/anakin/rec_qmix.py
RuanJohn: I know we reuse the code a bit between the two. But since this accesses the config without having it passed in, it takes a bit of work to move it to a util file. If you feel strongly, I am happy to do so?

mava/systems/q_learning/anakin/rec_qmix.py
RuanJohn: Since we use all the data later on (example on line 385), it won't work to call this data. But I agree data_first is a bad name.  Do you have another suggestion for the name?

mava/configs/env/smax.yaml
RuanJohn: Ah, good point. I am not too fussed about which one should be the default. I'll accept the suggestions since they make for faster runtimes :)

mava/configs/arch/anakin.yaml
```diff
@@ -1,7 +1,7 @@
 # --- Anakin config ---
 
 # --- Training ---
-num_envs: 16  # Number of vectorised environments per device.
+num_envs: 32  # Number of vectorised environments per device.
```
WiemKhlifi: Is this intended to be changed as well?

mava/configs/system/q_learning/rec_qmix.yaml
```diff
@@ -0,0 +1,35 @@
+# --- Defaults REC-QMIX ---
+total_timesteps: ~ # Set the total environment steps.
+# If unspecified, it's derived from num_updates; otherwise, num_updates adjusts based on this value.
+num_updates: 10000 # Number of updates.
+seed: 1
```
WiemKhlifi: Nit (just to keep up with the rest): 
``` suggestion
seed: 42
``` 
Maybe we should put this outside in future in the arch file or something? ( maybe this seed is better than the 42?)

mava/networks/base.py
```diff
@@ -232,3 +233,73 @@ def __call__(
...
 
         return hidden_state, eps_greedy_dist
+
+
+class QMixingNetwork(nn.Module):
+    num_actions: int
+    num_agents: int
+    hyper_hidden_dim: int = 64
+    embed_dim: int = 32
+    norm_env_states: bool = True
```
WiemKhlifi, line 243: I guess it's fine not to put default params since all of these are passed from the net config 👀

mava/networks/base.py
```diff
@@ -232,3 +233,73 @@ def __call__(
...
+
+        self.layer_norm: nn.Module = nn.LayerNorm()
+
+    @nn.compact
+    def __call__(
+        self,
+        agent_qs: chex.Array,
+        env_global_state: chex.Array,
+    ) -> chex.Array:
+        B, T = agent_qs.shape[:2]  # batch size #noqa: N806
```
WiemKhlifi, line 274: I'm not sure you agree with this ( I believe @lbeyers did this?) but I guess it will be nicer if we use some meaningful names like `batch_size` instead of `B` in the rest of the code (same for T etc..) 🙏

mava/configs/system/q_learning/rec_qmix.yaml
lbeyers: I agree with what @sash-a said. Currently, it looks like for QMix sample_sequence_length is 20, and since we don't add the +1 in the flashbax init, that gives us 19 full data points for a training sequence, since we need 20 transitions to compute 19 q-targets.

mava/systems/q_learning/anakin/rec_qmix.py
lbeyers: How you name/handle this depends on what you want the number you put in the config to mean

lbeyers: See my comment above; with Rec IQL I thought it was better to have sample_sequence_length mean "the number of effective training samples I get in the time dimension" rather than "the number of transitions I sample in the time dimension". The lengths of those two things differ by one. Whichever you think will give the user a better idea of what they're doing when they set the variable "sample_sequence_length" is probably the better choice

lbeyers: But if you want "sample_sequence_length" to mean "the number of effective training samples I get in the time dimension", then you have to add the +1 back to the flashbax init

mava/configs/network/qmix_rnn.yaml
```diff
@@ -0,0 +1,19 @@
+# ---Recurrent Structure Networks---
+hidden_state_dim: 256 # The size of the RNN hiddenstate for each agent.
+
+q_network:
+  pre_torso:
```
WiemKhlifi, line 5: Can't we just set the same config in `rnn.yaml` as this? Does it hurt to make things twice bigger since it is still a default setting that it's not meant to do benchmarking with?

mava/systems/q_learning/anakin/rec_qmix.py
WiemKhlifi: Maybe `current_data`? This is aligned with `next_data`

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,690 @@
...
+            cfg.network.hidden_state_dim,
+        )
+
+        abs_metric_evaluator = get_eval_fn(eval_env, eval_act_fn, cfg, absolute_metric=True)
+        eval_metrics = abs_metric_evaluator(best_params, eval_keys, {"hidden_state": eval_hs})
+        logger.log(eval_metrics, t, eval_idx, LogEvent.ABSOLUTE)
+
+    logger.stop()
+
+    return float(eval_performance)
```
WiemKhlifi: The float is already applied above we can return it as is:
```suggestion
    return eval_performance
```
I guess this comes from the iql 😅

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -0,0 +1,690 @@
...
+    """Experiment entry point."""
+    # Allow dynamic attributes.
+    OmegaConf.set_struct(cfg, False)
+    cfg.logger.system_name = "rec_qmix"
+    # Run experiment.
+    final_return = run_experiment(cfg)
+
+    print(f"{Fore.CYAN}{Style.BRIGHT}QMIX experiment completed{Style.RESET_ALL}")
+
+    return float(final_return)
```
WiemKhlifi: Nit + remove float:
```suggestion
    eval_performance = run_experiment(cfg)

    print(f"{Fore.CYAN}{Style.BRIGHT}QMIX experiment completed{Style.RESET_ALL}")

    return eval_performance
```

WiemKhlifi (changes requested):
Thanks @RuanJohn this is looking good so far, I just have few questions and some suggestions 🙏

mava/configs/system/q_learning/rec_qmix.yaml
RuanJohn: Thanks for the help :) 
Updated the comment and the buffer init for IQL to be consistent between systems.

mava/systems/q_learning/anakin/rec_qmix.py
RuanJohn: Renamed to `data_t0` and `data_t1`. Let me know your thoughts?

mava/networks/base.py
RuanJohn: We have agreed to use capital letters everywhere in Mava. It is more compact and lines up with our shape comments. 
We do this in MAT as well now and once that is merged we can remove this `noqa`

mava/configs/network/qmix_rnn.yaml
RuanJohn: I don't like that because you end up with q_network configs in PPO systems and vice versa. I actually think we should have an rnn_actor_critic.yaml, rnn_qlearning.yaml and rnn_qmix.yaml. But feel we can do this in another PR.

mava/networks/base.py
RuanJohn: It's just nice to still have them in my opinion. :shrug:

mava/configs/arch/anakin.yaml
RuanJohn: Reset to 16.

mava/configs/system/q_learning/rec_qmix.yaml
RuanJohn: Reset. If we want to restructure configs, I think we can do that in another PR.

mava/systems/q_learning/anakin/rec_qmix.py
RuanJohn: A way around this is to pass in the config every time we call this, but it feels a bit extra to me? Or we need a function that inits this which also feels a bit extra to me.

mava/networks/base.py
sash-a: Ye it's a bit of a balancing act to keep readability, but I quite like the single capital letter variables for shapes.

Maybe we can mention it in a readme or something

mava/systems/q_learning/anakin/rec_qmix.py
sash-a: I see, ye passing the config in would be better, but I think it's fine :smile:

mava/configs/network/qmix_rnn.yaml
WiemKhlifi: Totally agree!

mava/networks/base.py
WiemKhlifi: Yeah agree if we opt for this we should mention it somewhere instead of repeating those comments to clarify the meaning of each letter 👍

mava/systems/q_learning/anakin/rec_qmix.py
```diff
@@ -531,15 +532,15 @@ def update_step(
...
-        )
+    pmaped_update_step: Callable[
+        [LearnerState[QMIXParams]], Tuple[LearnerState[QMIXParams], Tuple[Metrics, Metrics]]
+    ] = jax.pmap(
+        jax.vmap(
+            lambda state: lax.scan(update_step, state, None, length=cfg.system.scan_steps),
+            axis_name="batch",
+        ),
+        axis_name="device",
+        donate_argnums=0,
```
sash-a: nit: I think with the new mypy rules in the pyproject PR you don't need typing for this variable, can we try get that in and then remove the typing here :pray:

sash-a (changes requested):
Sorry 1 super minor thing, but would be nice to try if we can get the pyproject PR in first

sash-a (approved):
:1st_place_medal: :2nd_place_medal: :3rd_place_medal: :medal_sports:

WiemKhlifi (approved):
🌟 🕺 🦸

PR #1107, opened 2024-10-22
Add MAT

RuanJohn:
## What?
Adds the Multi-agent Transformer to Mava.

## Extra
The transformer based systems, or any other system with only one set of parameters, do not require us to make a `Params` `NamedTuple` with only one element. I had to update the checkpointer to no longer check if the params that are restored are in a `FrozenDict`. All other systems can still checkpoint and reload with the change. We also no longer need the check since we pin to a version of Flax that is higher than 0.6.11.


mava/configs/network/transformer.yaml
```diff
@@ -0,0 +1,8 @@
+# --- Network params ---
+actor_network:
+  n_block: 3
+  n_embd: 128
+  n_head: 1
```
sash-a: Should the default not be a bit smaller?

mava/networks/attention.py
```diff
@@ -0,0 +1,93 @@
...
+
+class SelfAttention(nn.Module):
+    n_embd: int
+    n_head: int
+    n_agent: int
+    masked: bool = False
+
+    def setup(self) -> None:
+        # flax defualt is to only be defined using features out.
+        # bias init is zero in MAT.
```
sash-a: Not following what this means?

mava/networks/attention.py
```diff
@@ -0,0 +1,93 @@
...
+        self.query = nn.Dense(self.n_embd, kernel_init=orthogonal(0.01))
+        self.value = nn.Dense(self.n_embd, kernel_init=orthogonal(0.01))
+
+        # output projection
+        self.proj = nn.Dense(self.n_embd, kernel_init=orthogonal(0.01))
+
+        # causal mask to ensure that attention is only applied to the left in the input sequence
+        self.mask = jnp.tril(jnp.ones((self.n_agent + 1, self.n_agent + 1))).reshape(
+            1, 1, self.n_agent + 1, self.n_agent + 1
+        )
```
sash-a: makes it a little clearer what we are doing I think (may need to be formatted for line length)
```suggestion
        self.mask = jnp.tril(jnp.ones((self.n_agent + 1, self.n_agent + 1)))[jnp.newaxis, jnp.newaxis]
```

mava/networks/attention.py
```diff
@@ -0,0 +1,93 @@
...
+    def __call__(self, key: chex.Array, value: chex.Array, query: chex.Array) -> chex.Array:
+        # Shape names:
+        # B: batch size
+        # L: sequence length
+        # D: embedding dimension
+        # H: number of heads
+        # hs: head size
+        # nh: number of heads
+
+        batch, seq_len, embed_dim = key.shape
```
sash-a: Seeing as you've defined them in the comment above I think the code would read well if you used `B, L, D` as the variable names also

mava/networks/attention.py
```diff
@@ -0,0 +1,93 @@
...
+            .transpose((0, 2, 1, 3))
+        )
+        v = (
+            self.value(value)
+            .reshape(batch, seq_len, self.n_head, embed_dim // self.n_head)
+            .transpose((0, 2, 1, 3))
+        )
+
+        # causal attention: (B, nh, L, hs) x (B, nh, hs, L) -> (B, nh, L, L)
+        att = jnp.matmul(q, k.transpose((0, 1, 3, 2))) * (1.0 / jnp.sqrt(k.shape[-1]))
```
sash-a: Not sure if it's worth it because the comments are nice here, but this is where I'd say an einsum style `rearange` would read really well, it would probably mean we wouldn't even need the comments because shapes would be easy to spot

mava/networks/attention.py
```diff
@@ -0,0 +1,93 @@
...
+        att = nn.softmax(att, axis=-1)
+
+        y = jnp.matmul(att, v)  # (B, nh, L, L) x (B, nh, L, hs) -> (B, nh, L, hs)
+        # re-assemble all head outputs side by side
+        y = y.transpose((0, 2, 1, 3))
+        y = y.reshape(batch, seq_len, embed_dim)
+
+        # output projection
+        y = self.proj(y)
+        return y  # (B, L, D)
```
sash-a: nit
```suggestion
        return self.proj(y)  # (B, L, D)
```

mava/systems/mat/types.py
```diff
@@ -0,0 +1,43 @@
...
+from optax._src.base import OptState
+from typing_extensions import NamedTuple
+
+from mava.types import State
+
+
+class Params(NamedTuple):
+    """Parameters of a network with only an actor."""
+
+    actor_params: FrozenDict
```
sash-a: I wouldn't call these actor params, maybe `transformer_params` or `mat_params` but also if we've got a named tuple with only 1 thing in it then I think we don't need the named tuple, can just store the params as flax returns them

mava/systems/mat/types.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class OptStates(NamedTuple):
+    """OptStates of actor only learner."""
+
+    actor_opt_state: OptState
```
sash-a: Similarly here, I think removing these types will make it all a bit simpler

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+    action_dim: int,  # (, )
+    legal_actions: Union[chex.Array, None],
+    key: chex.PRNGKey,
+) -> Tuple[chex.Array, chex.Array]:
+    # We don't need legal_actions for continuous actions but keep it to keep the APIs consistent.
+    del legal_actions
+    batch_size, n_agent, _ = obs_rep.shape
+    shifted_action = jnp.zeros((batch_size, n_agent, action_dim))
+    output_action = jnp.zeros((batch_size, n_agent, action_dim))
+    output_action_log = jnp.zeros((batch_size, n_agent))
```
sash-a: Shouldn't continuous also get a start of sequence indicator?

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+    action_dim: int,  # (, )
+    legal_actions: chex.Array,  # (B, N, A)
+    key: chex.PRNGKey,
+) -> Tuple[chex.Array, chex.Array]:
+    batch_size, n_agent, _ = obs_rep.shape
+    one_hot_action = jax.nn.one_hot(action, action_dim)  # (B, A)
+    shifted_action = jnp.zeros((batch_size, n_agent, action_dim + 1))  # (B, N, A +1)
+    shifted_action = shifted_action.at[:, 0, 0].set(1)
+    shifted_action = shifted_action.at[:, 1:, 1:].set(one_hot_action[:, :-1, :])
+    logit = decoder(shifted_action, obs_rep)  # (B, N, A)
```
sash-a: I think this is the only thing blocking the use of heads? IMO cont actions should definitely get a start of sequence indicator and so the only difference would be the `one_hot` which I think is ok to have an `if` for either here or in MAT itself. Anyways can we please create the issue to use heads in MAT/Sable

Also might be useful to have a method for this part maybe `create_shifted_action` because sable will use it also

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+        output_action_log = output_action_log.at[:, i].set(action_log)
+
+        update_shifted_action = i + 1 < n_agent
+        shifted_action = jax.lax.cond(
+            update_shifted_action,
+            lambda action=action, i=i, shifted_action=shifted_action: shifted_action.at[
+                :, i + 1, 1:
+            ].set(jax.nn.one_hot(action, action_dim)),
+            lambda shifted_action=shifted_action: shifted_action,
+        )
```
sash-a: I think we can just do an `at[].set()` [with mode = drop](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html)

```suggestion
        # Adds all expect the last action to shifted_actions, as it is out of range
        shifted_action.at[:, i + 1, 1:].set(jax.nn.one_hot(action, action_dim), mode="drop")
```

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+        output_action_log = output_action_log.at[:, i].set(action_log)
+
+        update_shifted_action = i + 1 < n_agent
+        shifted_action = jax.lax.cond(
+            update_shifted_action,
+            lambda action=action, i=i, shifted_action=shifted_action: shifted_action.at[
+                :, i + 1, :
+            ].set(action),
+            lambda shifted_action=shifted_action: shifted_action,
+        )
```
sash-a: same here
```suggestion
        # Adds all expect the last action to shifted_actions, as it is out of range
        shifted_action.at[:, i + 1, 1:].set(action), mode="drop")
```

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+        distribution = tfd.Normal(loc=act_mean, scale=action_std)
+        distribution = tfd.Independent(
+            TanhTransformedDistribution(distribution),
+            reinterpreted_batch_ndims=1,
+        )
+        action = distribution.sample(seed=sample_key)  # (B, A)
+        action_log = distribution.log_prob(action)  # (B,)
+
+        output_action = output_action.at[:, i, :].set(action)
+        output_action_log = output_action_log.at[:, i].set(action_log)
```
sash-a, line 156: nit
```suggestion
        output_action = output_action.at[:, i].set(action)
        output_action_log = output_action_log.at[:, i].set(action_log)
```

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+) -> Tuple[chex.Array, chex.Array]:
+    # We don't need legal_actions for continuous actions but keep it to keep the APIs consistent.
+    del legal_actions
+    batch_size, n_agent, _ = obs_rep.shape
+    shifted_action = jnp.zeros((batch_size, n_agent, action_dim))
+    output_action = jnp.zeros((batch_size, n_agent, action_dim))
+    output_action_log = jnp.zeros((batch_size, n_agent))
+
+    for i in range(n_agent):
+        act_mean = decoder(shifted_action, obs_rep)[:, i, :]  # (B, A)
```
sash-a, line 142: nit
```suggestion
        act_mean = decoder(shifted_action, obs_rep)[:, i]  # (B, A)
```

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+    output_action = jnp.zeros((batch_size, n_agent))
+    output_action_log = jnp.zeros_like(output_action)
+
+    for i in range(n_agent):
+        logit = decoder(shifted_action, obs_rep)[:, i, :]  # (B, A)
+        masked_logits = jnp.where(
+            legal_actions[:, i, :],
+            logit,
+            jnp.finfo(jnp.float32).min,
+        )
```
sash-a, line 109: nit
```suggestion
        logit = decoder(shifted_action, obs_rep)[:, i]  # (B, A)
        masked_logits = jnp.where(
            legal_actions[:, i],
            logit,
            jnp.finfo(jnp.float32).min,
        )
```

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+    action_dim: int,  # (, )
+    legal_actions: chex.Array,  # (B, N, A)
+    key: chex.PRNGKey,
+) -> Tuple[chex.Array, chex.Array]:
+    # We don't need legal_actions for continuous actions but keep it to keep the APIs consistent.
+    del legal_actions
+    batch_size, n_agent, _ = obs_rep.shape
+    shifted_action = jnp.zeros((batch_size, n_agent, action_dim))
+
+    shifted_action = shifted_action.at[:, 1:, :].set(action[:, :-1, :])
```
sash-a, line 74: nit
```suggestion
    shifted_action = shifted_action.at[:, 1:].set(action[:, :-1])
```

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+    action: chex.Array,  # (B, N)
+    action_dim: int,  # (, )
+    legal_actions: chex.Array,  # (B, N, A)
+    key: chex.PRNGKey,
+) -> Tuple[chex.Array, chex.Array]:
+    batch_size, n_agent, _ = obs_rep.shape
+    one_hot_action = jax.nn.one_hot(action, action_dim)  # (B, A)
+    shifted_action = jnp.zeros((batch_size, n_agent, action_dim + 1))  # (B, N, A +1)
+    shifted_action = shifted_action.at[:, 0, 0].set(1)
+    shifted_action = shifted_action.at[:, 1:, 1:].set(one_hot_action[:, :-1, :])
```
sash-a, line 45: nit
```suggestion
    shifted_action = shifted_action.at[:, 1:, 1:].set(one_hot_action[:, :-1])
```

mava/networks/utils/mat/decode.py
```diff
@@ -0,0 +1,171 @@
...
+from flax import linen as nn
+
+from mava.networks.distributions import IdentityTransformation, TanhTransformedDistribution
+
+# General shapes legend:
+# B: batch size
+# N: number of agents
+# O: observation dimension
+# A: action dimension
+# E: model embedding dimension
```
sash-a, line 30: Really like this, like I said in transformer though I think we should use them as variable names

sash-a (commented):
Looking really nice so far! Will give it a proper look tomorrow, haven't looked at the two big files yet

Made some mega nits with the extra `:` when slicing, feel free to reject those if you feel they are useful :smile:

mava/networks/mat_network.py
```diff
@@ -0,0 +1,317 @@
...
+    ffn_dim: int
+    embed_dim: int
+
+    def setup(self) -> None:
+        self.W_1 = self.param("W_1", nn.initializers.zeros, (self.embed_dim, self.ffn_dim))
+        self.W_G = self.param("W_G", nn.initializers.zeros, (self.embed_dim, self.ffn_dim))
+        self.W_2 = self.param("W_2", nn.initializers.zeros, (self.ffn_dim, self.embed_dim))
+
+    def __call__(self, x: chex.Array) -> chex.Array:
+        return (jax.nn.swish(x @ self.W_G) * (x @ self.W_1)) @ self.W_2
```
sash-a: Should we add this `torsos.py`?

mava/networks/mat_network.py
```diff
@@ -0,0 +1,317 @@
...
+        if self.use_swiglu:
+            self.mlp = SwiGLU(self.n_embd, self.n_embd)
+        else:
+            self.mlp = nn.Sequential(
+                [
+                    nn.Dense(self.n_embd, kernel_init=orthogonal(jnp.sqrt(2))),
+                    nn.gelu,
+                    nn.Dense(self.n_embd, kernel_init=orthogonal(0.01)),
+                ],
+            )
```
sash-a: Can we extract this to a method in this file, something like `_make_mlp` (we use it in decode block also)

mava/networks/mat_network.py
```diff
@@ -0,0 +1,317 @@
...
+        return (jax.nn.swish(x @ self.W_G) * (x @ self.W_1)) @ self.W_2
+
+
+class EncodeBlock(nn.Module):
+    n_embd: int
+    n_head: int
+    n_agent: int
+    masked: bool = False
+    use_swiglu: bool = False
+    use_rmsnorm: bool = False
```
sash-a: Does sable just pass a network config for some of this stuff? If so we should follow that - maybe also make a simple type `MATNetConf` just so we don't lose info from doing this

mava/networks/mat_network.py
```diff
@@ -0,0 +1,317 @@
...
+        )
+
+    def __call__(self, obs: chex.Array) -> Tuple[chex.Array, chex.Array]:
+        obs_embeddings = self.obs_encoder(obs)
+        x = obs_embeddings
+
+        rep = self.blocks(self.ln(x))
+        v_loc = self.head(rep)
+
+        return jnp.squeeze(v_loc, axis=-1), rep
```
sash-a: Never understood why this was call v_loc, I think value is more explicit?
```suggestion
        value = self.head(rep)

        return jnp.squeeze(value, axis=-1), rep
```

mava/networks/mat_network.py
```diff
@@ -0,0 +1,317 @@
...
+            use_swiglu=self.use_swiglu,
+            use_rmsnorm=self.use_rmsnorm,
+        )
+
+        if self.action_space_type == "discrete":
+            self.act_function = discrete_autoregressive_act
+            self.train_function = discrete_parallel_act
+        elif self.action_space_type == "continuous":
+            self.act_function = continuous_autoregressive_act
+            self.train_function = continuous_parallel_act
```
sash-a: Just need to import these from the file where we get the action space type

```suggestion
        if self.action_space_type == _DISCRETE:
            self.act_function = discrete_autoregressive_act
            self.train_function = discrete_parallel_act
        elif self.action_space_type == _CONTINUOUS:
            self.act_function = continuous_autoregressive_act
            self.train_function = continuous_parallel_act
```

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+)
+from mava.utils.logger import LogEvent, MavaLogger
+from mava.utils.network_utils import get_action_head
+from mava.utils.total_timestep_checker import check_total_timesteps
+from mava.utils.training import make_learning_rate
+from mava.wrappers.episode_metrics import get_final_step_metrics
+
+
+def get_learner_fn(
+    env: Environment,
```
sash-a: Can you do a find and replace all of these with MarlEnv
```suggestion
    env: MarlEnv,
```

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+        def _env_step(learner_state: LearnerState, _: Any) -> Tuple[LearnerState, PPOTransition]:
+            """Step the environment."""
+            params, opt_state, key, env_state, last_timestep = learner_state
+
+            # SELECT ACTION
+            key, policy_key = jax.random.split(key)
+            action, log_prob, value = actor_action_select_fn(  # type: ignore
+                params.actor_params,
+                last_timestep.observation.agents_view,
+                last_timestep.observation.action_mask,
```
sash-a: Can we take in the whole observation here? Just keeps it inline with the rest of mava

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+        )
+
+        # CALCULATE ADVANTAGE
+        params, opt_state, key, env_state, last_timestep = learner_state
+
+        key, last_val_key = jax.random.split(key)
+        _, _, last_val = actor_action_select_fn(  # type: ignore
+            params.actor_params,
+            last_timestep.observation.agents_view,
+            last_timestep.observation.action_mask,
```
sash-a: Same here

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+            last_val_key,
+        )
+
+        def _calculate_gae(
+            traj_batch: PPOTransition,
+            last_val: chex.Array,
+        ) -> Tuple[chex.Array, chex.Array]:
```
sash-a: nit:
```suggestion
        def _calculate_gae(
            traj_batch: PPOTransition, last_val: chex.Array
        ) -> Tuple[chex.Array, chex.Array]:
```

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+                _get_advantages,
+                (jnp.zeros_like(last_val), last_val),
+                traj_batch,
+                reverse=True,
+                unroll=16,
+            )
+            return (
+                advantages,
+                advantages + traj_batch.value,
+            )
```
sash-a: nit
```suggestion
            return advantages, advantages + traj_batch.value
```

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+
+        def _update_epoch(update_state: Tuple, _: Any) -> Tuple:
+            """Update the network for a single epoch."""
+
+            def _update_minibatch(train_state: Tuple, batch_info: Tuple) -> Tuple:
+                """Update the network for a single minibatch."""
+
+                # UNPACK TRAIN STATE AND BATCH INFO
+                params, opt_state, key = train_state
+                (traj_batch, advantages, targets) = batch_info
```
sash-a: nit
```suggestion
                traj_batch, advantages, targets = batch_info
```

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+                    entropy_key: chex.PRNGKey,
+                ) -> Tuple:
+                    """Calculate the actor loss."""
+                    # RERUN NETWORK
+
+                    log_prob, value, entropy = actor_apply_fn(  # type: ignore
+                        actor_params,
+                        traj_batch.obs.agents_view,
+                        traj_batch.action,
+                        traj_batch.obs.action_mask,
```
sash-a: Same here

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+                return (new_params, new_opt_state, key), loss_info
+
+            (
+                params,
+                opt_state,
+                traj_batch,
+                advantages,
+                targets,
+                key,
+            ) = update_state
```
sash-a: nit
```suggestion
            params, opt_state, traj_batch, advantages, targets, key = update_state
```

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+            )
+
+            update_state = (
+                params,
+                opt_state,
+                traj_batch,
+                advantages,
+                targets,
+                key,
+            )
```
sash-a:
```suggestion
            update_state = params, opt_state, traj_batch, advantages, targets, key
```

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+) -> Tuple[LearnerFn[LearnerState], Any, LearnerState]:
+    """Initialise learner_fn, network, optimiser, environment and states."""
+    # Get available TPU cores.
+    n_devices = len(jax.devices())
+
+    # Get number of agents.
+    config.system.num_agents = env.num_agents
+
+    # PRNG keys.
+    key, actor_net_key, _ = keys
```
sash-a: Can we just pass in 1 less key?

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+        obs_dim=init_x.agents_view.shape[-1],
+        action_dim=env.action_dim,
+        n_block=config.network.actor_network.n_block,
+        n_head=config.network.actor_network.n_head,
+        n_embd=config.network.actor_network.n_embd,
+        n_agent=config.system.num_agents,
+        use_rmsnorm=config.network.use_rmsnorm,
+        use_swiglu=config.network.use_swiglu,
+        action_space_type=action_space_type,
+    )
```
sash-a: Definitely think initializing with a config would be nice here

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,647 @@
...
+
+        # Separately log timesteps, actoring metrics and training metrics.
+        logger.log({"timestep": t}, t, eval_step, LogEvent.MISC)
+        if ep_completed:  # only log episode metrics if an episode was completed in the rollout.
+            logger.log(episode_metrics, t, eval_step, LogEvent.ACT)
+        logger.log(learner_output.train_metrics, t, eval_step, LogEvent.TRAIN)
+
+        # Prepare for evaluation.
+        start_time = time.time()
+
```
sash-a: We don't measure eval time out here anymore
```suggestion
```

sash-a (changes requested):
Looking super good! I think main question is the net config :fire:

mava/systems/mat/anakin/mat.py
RuanJohn: We would need to agree on what this config looks like, because Sable doesn't use a network config like you'd expect either.

mava/networks/attention.py
RuanJohn: These were just old notes I made for myself. Will remove.

mava/configs/network/transformer.yaml
RuanJohn: Ye for sure. For now, I am just keeping the default configs for small-4ag so I can easily test that we don't regress with any changes.

mava/networks/attention.py
RuanJohn: Done.

mava/networks/utils/mat/decode.py
RuanJohn: No, it doesn't. Please see the original implementation [here](https://github.com/PKU-MARL/Multi-Agent-Transformer/blob/be3ff49c8264d454c1fe2c41582aa2bfc98498c8/mat/algorithms/utils/transformer_act.py#L47) and [here](https://github.com/PKU-MARL/Multi-Agent-Transformer/blob/be3ff49c8264d454c1fe2c41582aa2bfc98498c8/mat/algorithms/utils/transformer_act.py#L73).

mava/networks/utils/mat/decode.py
RuanJohn: It goes a bit beyond just star tof sequence tokens or not and it is also not really the parallel versions that are affected. It is because when we use heads we cannot pass in a single agent's embedded observation at a time nor can we mask a single agents actions when initialising the distributions for action selection. When using heads we are forced to pass all the agent's observations and legal action masks through at the same time. This had a non-zero effect on the performance when I did it to test if using heads was possible and for this reason I think this is something we should skip for now. Make and issue and consider looking at in the future. 

I don't see this as a major issue since these transformer based systems are already so different from other systems we have in Mava that being slightly different here for the sake of performance is okay in my opinion.

mava/networks/utils/mat/decode.py
RuanJohn: Since there are slight shape differences between continuous and discrete I quite like the extra `:` since it shows the dim more explicitly. During development we have also had small bugs creep in because of this very reason.

mava/networks/utils/mat/decode.py
RuanJohn: Same as above.

mava/networks/mat_network.py
RuanJohn: Totally agree! Moved.

mava/networks/mat_network.py
RuanJohn: No, Sable still passes everything through but has an extra Net config for setting whether it has memory or not as far as I am aware. Maybe @OmaymaMahjoub can expand a bit on this?

mava/networks/mat_network.py
RuanJohn: Just got it from the original implementation :sweat_smile:
Renamed.

mava/networks/mat_network.py
RuanJohn: Ah yes, I forgot to do this :sweat_smile: 
Done.

mava/systems/mat/anakin/mat.py
RuanJohn: Done

mava/systems/mat/anakin/mat.py
RuanJohn: Done.

mava/systems/mat/anakin/mat.py
RuanJohn: Done

mava/systems/mat/anakin/mat.py
RuanJohn: Yep. Done.

mava/networks/attention.py
RuanJohn: Added a todo about einops since the results didn't match up nicely between the reshapes and einops on a quick test. We can investigate in the future.

mava/networks/mat_network.py
RuanJohn: Done :)

mava/networks/utils/mat/decode.py
RuanJohn: Done.

mava/systems/mat/types.py
RuanJohn: Done. Also updated the checkpointer to accommodate for this.

mava/networks/mat_network.py
RuanJohn: Updated to now have a net config

mava/systems/mat/anakin/mat.py
RuanJohn: Please see the net_config update :)

mava/networks/attention.py
```diff
@@ -36,47 +36,34 @@ def setup(self) -> None:
         self.proj = nn.Dense(self.n_embd, kernel_init=orthogonal(0.01))
 
         # causal mask to ensure that attention is only applied to the left in the input sequence
-        self.mask = jnp.tril(jnp.ones((self.n_agent + 1, self.n_agent + 1))).reshape(
-            1, 1, self.n_agent + 1, self.n_agent + 1
-        )
+        self.mask = jnp.tril(jnp.ones((self.n_agent + 1, self.n_agent + 1)))[
+            jnp.newaxis, jnp.newaxis
+        ]
```
sash-a: mega nit, not a fan of the indexing on a new line:
```suggestion
        self.mask = jnp.tril(jnp.ones((self.n_agent + 1, self.n_agent + 1)))
        self.mask = self.mask[jnp.newaxis, jnp.newaxis]
```

mava/networks/mat_network.py
```diff
@@ -160,53 +138,48 @@ def __call__(self, x: chex.Array, rep_enc: chex.Array) -> chex.Array:
...
+        ln = nn.RMSNorm if self.net_config.use_rmsnorm else nn.LayerNorm
 
-        if self.action_space_type == "discrete":
+        if self.action_space_type == _DISCRETE:
             self.action_encoder = nn.Sequential(
                 [
-                    nn.Dense(self.n_embd, use_bias=False, kernel_init=orthogonal(jnp.sqrt(2))),
+                    nn.Dense(
+                        self.net_config.n_embd, use_bias=False, kernel_init=orthogonal(jnp.sqrt(2))
+                    ),
```
sash-a: Does a bias make such a big difference for cont vs discrete systems, seems like everything else is the same?

If so then maybe `use_bais=self.action_space_type == _CONTINUOUS` is cleaner

mava/systems/mat/anakin/mat.py
```diff
@@ -420,13 +382,8 @@ def learner_setup(
...
 
     # Initialise actor params and optimiser state.
-    actor_params = actor_network.init(
-        actor_net_key, init_x.agents_view, init_action, init_x.action_mask, jax.random.PRNGKey(0)
-    )
-    actor_opt_state = actor_optim.init(actor_params)
-
-    # Pack params.
-    params = Params(actor_params)
+    params = actor_network.init(actor_net_key, init_x, init_action, jax.random.PRNGKey(0))
```
sash-a, line 384: Nice reminder
```suggestion
    # `PRNGKey(0)` is used for action selection, so doesn't matter during init
    params = actor_network.init(actor_net_key, init_x, init_action, jax.random.PRNGKey(0))
```

mava/systems/mat/anakin/mat.py
```diff
@@ -490,14 +446,10 @@ def run_experiment(_config: DictConfig) -> float:
     env, eval_env = environments.make(config)
 
     # PRNG keys.
-    key, key_e, actor_net_key, critic_net_key = jax.random.split(
-        jax.random.PRNGKey(config.system.seed), num=4
-    )
+    key, key_e, actor_net_key, _ = jax.random.split(jax.random.PRNGKey(config.system.seed), num=4)
```
sash-a:
```suggestion
    key, key_e, actor_net_key = jax.random.split(jax.random.PRNGKey(config.system.seed), num=3)
```

test/integration_test.py
```diff
@@ -101,6 +102,19 @@ def test_sac_system(fast_config: dict, system_path: str) -> None:
     _run_system(system_path, cfg)
 
 
+@pytest.mark.parametrize("system_path", transformer_systems)
+def test_transformer_system(fast_config: dict, system_path: str) -> None:
+    """Test transformer systems on random envs."""
+    _, _, system_name = system_path.split(".")
+    env = random.choice(continuous_envs)
```
sash-a:
```suggestion
    env = random.choice(discrete_envs + continuous_envs)
```

sash-a (changes requested):
Couple small things

mava/configs/arch/anakin.yaml
```diff
@@ -1,13 +1,13 @@
 # --- Anakin config ---
 
 # --- Training ---
-num_envs: 16  # Number of vectorised environments per device.
+num_envs: 64  # Number of vectorised environments per device.
```
WiemKhlifi: Spamming just to revert before merging if not aiming to update these 👀 
```suggestion
num_envs: 16  # Number of vectorised environments per device.
num_evaluation: 200  # Number of evenly spaced evaluations to perform during training.
```

mava/configs/env/rware.yaml
```diff
@@ -1,7 +1,7 @@
 # ---Environment Configs---
 defaults:
   - _self_
-  - scenario: tiny-2ag # [tiny-2ag, tiny-4ag, tiny-4ag-easy, small-4ag]
+  - scenario: small-4ag # [tiny-2ag, tiny-4ag, tiny-4ag-easy, small-4ag]
```
WiemKhlifi: Spam: revert before merge
```suggestion
  - scenario: tiny-2ag # [tiny-2ag, tiny-4ag, tiny-4ag-easy, small-4ag]
```

mava/configs/system/mat/mat.yaml
```diff
@@ -0,0 +1,25 @@
...
+total_timesteps: ~ # Set the total environment steps.
+# If unspecified, it's derived from num_updates; otherwise, num_updates adjusts based on this value.
+num_updates: 1220 # Number of updates
+seed: 42
+
+# --- Agent observations ---
+add_agent_id: True
+
+# --- RL hyperparameters ---
+actor_lr: 0.0005 # Learning rate for actor network
```
WiemKhlifi, line 12: Just raising this for future maybe we should change `actor_lr` for both MAT and Sable to `network_lr` or something else since it's not really an actor network 💭

mava/networks/attention.py
```diff
@@ -0,0 +1,77 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import chex
+import jax.numpy as jnp
+from flax import linen as nn
+from flax.linen.initializers import orthogonal
+
+# TODO: Use einops for all the reshapes and matrix multiplications
```
WiemKhlifi, line 20: Is this still a todo or will be ignored?

sash-a (dismissed):
:robot: :robot: :robot: :robot:

mava/networks/mat_network.py
```diff
@@ -0,0 +1,274 @@
...
+from mava.networks.torsos import SwiGLU
+from mava.networks.utils.mat.decode import (
+    continuous_autoregressive_act,
+    continuous_parallel_act,
+    discrete_autoregressive_act,
+    discrete_parallel_act,
+)
+from mava.systems.mat.types import MATNetworkConfig
+from mava.types import MavaObservation
+from mava.utils.network_utils import _CONTINUOUS, _DISCRETE
```
WiemKhlifi, line 32: NIT:
```suggestion
from mava.utils.network_utils import CONTINUOUS, DISCRETE
```

WiemKhlifi (changes requested):
Some nits else the rest is good to go 🛥️ 🥅

mava/configs/default/mat.yaml
```diff
@@ -0,0 +1,11 @@
+defaults:
+  - logger: logger
+  - arch: anakin
+  - system: mat/mat
+  - network: transformer
+  - env: rware
```
OmaymaMahjoub:
```suggestion
  - env: rware  # [cleaner, connector, gigastep, lbf, mabrax, matrax, rware, smax]
```

test/integration_test.py
```diff
@@ -101,6 +102,19 @@ def test_sac_system(fast_config: dict, system_path: str) -> None:
...
+    with initialize(version_base=None, config_path=config_path):
+        cfg = compose(config_name=f"{system_name}", overrides=[f"env={env}"])
+        cfg = _get_fast_config(cfg, fast_config)
+
+    _run_system(system_path, cfg)
+
+
 @pytest.mark.parametrize("env_name", discrete_envs)
 def test_discrete_env(fast_config: dict, env_name: str) -> None:
     """Test all discrete envs on random systems."""
```
OmaymaMahjoub, line 121: if you can add MAT to the random choice here

mava/configs/network/transformer.yaml
```diff
@@ -0,0 +1,6 @@
+# --- Network params ---
+n_block: 1 # Transformer blocks
+n_embd: 64 # Transformer embedding dimension
```
OmaymaMahjoub: for sake of unification maybe we can rename this to embed_dim similar to Sable

mava/networks/torsos.py
```diff
@@ -76,6 +76,19 @@ def __call__(self, observation: chex.Array) -> chex.Array:
...
+    embed_dim: int
+
+    def setup(self) -> None:
+        self.W_1 = self.param("W_1", nn.initializers.zeros, (self.embed_dim, self.ffn_dim))
+        self.W_G = self.param("W_G", nn.initializers.zeros, (self.embed_dim, self.ffn_dim))
+        self.W_2 = self.param("W_2", nn.initializers.zeros, (self.ffn_dim, self.embed_dim))
+
+    def __call__(self, x: chex.Array) -> chex.Array:
+        return (jax.nn.swish(x @ self.W_G) * (x @ self.W_1)) @ self.W_2
+
```
OmaymaMahjoub:
```suggestion
class SwiGLU(nn.Module):
    """SiwGLU module for Sable's Network.

    Implements the SwiGLU feedforward neural network module, which is a variation
    of the standard feedforward layer using the Swish activation function combined
    with a Gated Linear Unit (GLU).
    """

    hidden_dim: int
    input_dim: int

    def setup(self) -> None:
        # Initialize the weights for the SwiGLU layer
        self.W_linear = self.param(
            "W_linear", nn.initializers.zeros, (self.input_dim, self.hidden_dim)
        )
        self.W_gate = self.param("W_gate", nn.initializers.zeros, (self.input_dim, self.hidden_dim))
        self.W_output = self.param(
            "W_output", nn.initializers.zeros, (self.hidden_dim, self.input_dim)
        )

    def __call__(self, x: chex.Array) -> chex.Array:
        """Applies the SwiGLU mechanism to the input tensor `x`."""
        # Apply Swish activation to the gated branch and multiply with the linear branch
        gated_output = jax.nn.swish(x @ self.W_gate) * (x @ self.W_linear)
        # Transform the result back to the input dimension
        return gated_output @ self.W_output

```

OmaymaMahjoub: I added documentation and updated vars naming in swiglu, i added this suggestion here since MAT PR will be first to be merged, also adding it to torsos.py is better than in the outside utils folder (i will update sable PR based on that)

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,597 @@
...
+from mava.utils.logger import LogEvent, MavaLogger
+from mava.utils.network_utils import get_action_head
+from mava.utils.total_timestep_checker import check_total_timesteps
+from mava.utils.training import make_learning_rate
+from mava.wrappers.episode_metrics import get_final_step_metrics
+
+
+def get_learner_fn(
+    env: MarlEnv,
+    apply_fns: Tuple[ActorApply, CriticApply],
```
OmaymaMahjoub: if you can create type ExecutionApply and TrainApply instead of using actor critic one

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,597 @@
...
+
+    # PRNG keys.
+    key, key_e, actor_net_key = jax.random.split(jax.random.PRNGKey(config.system.seed), num=3)
+
+    # Setup learner.
+    learn, actor_network, learner_state = learner_setup(env, (key, actor_net_key), config)
+
+    eval_keys = jax.random.split(key_e, n_devices)
+
+    def eval_act_fn(
```
OmaymaMahjoub, line 455: If we can here follow the ppo systems where we call a maker function from the evaluator instead of creating it here

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,597 @@
...
+        logger.log(learner_output.train_metrics, t, eval_step, LogEvent.TRAIN)
+
+        trained_params = unreplicate_batch_dim(learner_state.params)
+        key_e, *eval_keys = jax.random.split(key_e, n_devices + 1)
+        eval_keys = jnp.stack(eval_keys)
+        eval_keys = eval_keys.reshape(n_devices, -1)
+
+        # Evaluate.
+        eval_metrics = evaluator(trained_params, eval_keys, {})
+        jax.block_until_ready(eval_metrics)
```
OmaymaMahjoub, line 537: do we need these block_until_ready? We never added them to other on-policy systems

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,597 @@
...
+    if action_space_type == "discrete":
+        init_action = jnp.zeros((1, config.system.num_agents), dtype=jnp.int32)
+    elif action_space_type == "continuous":
+        init_action = jnp.zeros((1, config.system.num_agents, env.action_dim), dtype=jnp.float32)
+    else:
+        raise ValueError("Invalid action space type")
+
+    # Define network and optimiser.
+    actor_network = MultiAgentTransformer(
+        obs_dim=init_x.agents_view.shape[-1],
```
OmaymaMahjoub: we don't need obs_dim as input

mava/systems/mat/anakin/mat.py
```diff
@@ -0,0 +1,597 @@
...
+
+    if action_space_type == "discrete":
+        init_action = jnp.zeros((1, config.system.num_agents), dtype=jnp.int32)
+    elif action_space_type == "continuous":
+        init_action = jnp.zeros((1, config.system.num_agents, env.action_dim), dtype=jnp.float32)
+    else:
+        raise ValueError("Invalid action space type")
+
+    # Define network and optimiser.
+    actor_network = MultiAgentTransformer(
```
OmaymaMahjoub, line 368: can we rename this to mat_network (very optional)

OmaymaMahjoub (changes requested):
Thanks Ruan for all the work on adding MAT, I left some comments but most of them nit and minor suggestions

mava/networks/mat_network.py
RuanJohn: This import won't work since the variables are called `_CONTINUOUS` and `_DISCRETE` in `mava.utils.network_utils`.

mava/networks/attention.py
RuanJohn: We decided to leave it as a todo instead of making an issue.

mava/configs/arch/anakin.yaml
RuanJohn: Done.

mava/configs/env/rware.yaml
RuanJohn: Done.

mava/configs/default/mat.yaml
RuanJohn: Done. I didn't include cleaner and connector since MAT doesn't have a CNN and these envs don't have vector obs in Mava yet.

mava/configs/system/mat/mat.yaml
RuanJohn: My suggestion is that we keep it as it for now.

test/integration_test.py
RuanJohn: I spoke to Sasha about this and since the goal is to just test the envs, it makes more sense to omit MAT here, since it will be the slowest. I think we should do the same for Sable where we only test the algo but exclude it from the env tests.

mava/systems/mat/anakin/mat.py
RuanJohn: The reason PPO and SAC defines it in the evaluator is because the same function is used across many systems. Since MAT uses a different `act_fn` and is the only one that uses it it makes more sense to me to define it in the system run file.

mava/systems/mat/anakin/mat.py
RuanJohn: I don't think it hurts to have them.

mava/systems/mat/anakin/mat.py
RuanJohn: Good catch. Removed.

mava/systems/mat/anakin/mat.py
RuanJohn: I think `MultiAgentTransformer` is a bit nicer. So will keep as is.

mava/configs/network/transformer.yaml
RuanJohn: Done.

mava/systems/mat/anakin/mat.py
RuanJohn: Done.

mava/networks/torsos.py
RuanJohn: Thank you for the suggestions! I implemented some of them and then didn't add some because I feel the code is self-explanatory enough. 
I also think naming the one variable `embed_dim` instead of `input_dim` is more consistent with the rest of our code.

test/integration_test.py
OmaymaMahjoub: Sure, i will update sable PR based on this

mava/systems/mat/anakin/mat.py
OmaymaMahjoub: That's convincing, thanks for the explanation

OmaymaMahjoub (approved):
Thanks @RuanJohn  🔥

mava/configs/system/mat/mat.yaml
WiemKhlifi: Yeah in a future PR we can do a cleaning for both Sable and MAT 👍

mava/networks/attention.py
WiemKhlifi: Understood thanks!

mava/networks/mat_network.py
WiemKhlifi: Yeah I was thinking of changing that but nvm 😅

WiemKhlifi (approved):
Thanks @RuanJohn 🥳 🎆 🌟

PR #1109, opened 2024-10-24
feat: switch to pyproject.toml

sash-a:
## What?
Switch to pyproject.toml over setup.py and update mypy rules
## Why?
pyproject.toml is the modern approach and mypy rules where a bit strict, switched to defaults where appropriate


mava/advanced_usage/ff_ippo_store_experience.py
```diff
@@ -548,11 +548,9 @@ def run_experiment(_config: DictConfig) -> None:
     def _reshape_experience(experience: Dict[str, chex.Array]) -> Dict[str, chex.Array]:
         """Reshape experience to match buffer."""
         # Swap the T and NE axes (D, NU, UB, T, NE, ...) -> (D, NU, UB, NE, T, ...)
-        experience: Dict[str, chex.Array] = tree.map(lambda x: x.swapaxes(3, 4), experience)
```
RuanJohn, line 551: I really like that we can take this away. What did you change?

mava/utils/make_env.py
```diff
@@ -147,7 +147,7 @@ def make_jaxmarl_env(
         add_global_state,
     )
 
-    train_env, eval_env = add_extra_wrappers(train_env, eval_env, config)
+    train_env, eval_env = add_extra_wrappers(train_env, eval_env, config)  # type: ignore
```
RuanJohn: Why do we have to `# type: ignore` here now? Would be great if we could fix this :pray:

mava/utils/make_env.py
```diff
@@ -208,7 +208,7 @@ def make_gigastep_env(
     train_env = wrapper(scenario.make(**kwargs), has_global_state=add_global_state)
     eval_env = wrapper(scenario.make(**kwargs), has_global_state=add_global_state)
 
-    train_env, eval_env = add_extra_wrappers(train_env, eval_env, config)
+    train_env, eval_env = add_extra_wrappers(train_env, eval_env, config)  # type: ignore
```
RuanJohn: Same here for the type ignore.

pyproject.toml
```diff
@@ -1,39 +1,54 @@
...
+requires-python=">=3.9"  # would be nice to upgrade this and jumanji at some point
+keywords=["multi-agent", "reinforcement learning", "python", "jax", "anakin", "sebulba"]
+classifiers=[
+    "Environment :: Console",
+    "Intended Audience :: Science/Research",
+    "Intended Audience :: Developers",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.12",
```
RuanJohn, line 25: Not sure about this, but can we include 3.11 as well since we support it.

pyproject.toml
```diff
@@ -1,39 +1,54 @@
...
-disallow_untyped_calls = false
-disallow_untyped_defs = true
-disallow_incomplete_defs = true
-check_untyped_defs = true
-disallow_untyped_decorators = false
-strict_optional = true
-strict_equality = true
-explicit_package_bases = true
-follow_imports = "skip"
-ignore_missing_imports = true
```
RuanJohn, line 20: It looks like we are removing quite a few mypy args. Is there a reason for this?

pyproject.toml
```diff
@@ -1,39 +1,54 @@
...
-    "numpy.*",
-    "optax.*",
-    "neptune.*",
-    "hydra.*",
-    "omegaconf.*",
+[project]
+name="id-mava"
+authors=[{name="InstaDeep Ltd"}]
+dynamic=["version", "dependencies", "optional-dependencies"]
+license={file="LISCENSE"}
```
RuanJohn: Just a small typo.

```suggestion
license={file="LICENSE"}
```

RuanJohn (changes requested):
This is super nice! :fire: 

A few questions and a typo fix.

mava/advanced_usage/ff_ippo_store_experience.py
sash-a: It was one of the mypy rules that I felt was a bit strict and also wasn't a default

mava/utils/make_env.py
sash-a: So this is from changing one of the mypy rules to a default rule. Honestly it's an us problem and should be an error, it's because `train_env` and `eval_env` for some reason don't conform to the `MarlEnv` spec, I'll have a quick look but it might not be easily fixable

pyproject.toml
sash-a: Ye a lot of them were either already set by the defaults, or very strict and not the default. I kept only the super necessary ones :smile:

mava/utils/make_env.py
sash-a: I tried but I couldn't find a fix for this. Added missing method so the wrappers should now comply with `MarlEnv` but mypy is complaining :shrug:

pyproject.toml
```diff
@@ -1,39 +1,55 @@
...
+version={attr="mava.__version__"}
+dependencies={file="requirements/requirements.txt"}
+optional-dependencies={dev={file=["requirements/requirements-dev.txt"]}}
+
+[project.urls]
+"Homepage"="https://github.com/instadeep/Mava"
+"Bug Tracker"="https://github.com/instadeep/Mava/issues"
+
+[tool.mypy]
+python_version="3.11"
```
WiemKhlifi, line 40: Why not `python_version="3.12"`?

mava/networks/torsos.py
```diff
@@ -59,7 +59,9 @@ def setup(self) -> None:
     def __call__(self, observation: chex.Array) -> chex.Array:
         """Forward pass."""
         x = observation
-        for channel, kernel, stride in zip(self.channel_sizes, self.kernel_sizes, self.strides):
+        for channel, kernel, stride in zip(
+            self.channel_sizes, self.kernel_sizes, self.strides, strict=False
+        ):
```
WiemKhlifi: The default is `False `but maybe it's good to set it to `True` to avoid mismatch?
```suggestion
        for channel, kernel, stride in zip(
            self.channel_sizes, self.kernel_sizes, self.strides, strict=True
        ):
```

mava/utils/logger.py
```diff
@@ -296,7 +296,7 @@ def log_dict(self, data: Dict, step: int, eval_step: int, event: LogEvent) -> No
         for value in data.values():
             value = value.item() if isinstance(value, jax.Array) else value
             values.append(f"{value:.3f}" if isinstance(value, float) else str(value))
-        log_str = " | ".join([f"{k}: {v}" for k, v in zip(keys, values)])
+        log_str = " | ".join([f"{k}: {v}" for k, v in zip(keys, values, strict=False)])
```
WiemKhlifi: Same here as mentioned above? set True?

mava/types.py
```diff
@@ -103,7 +102,7 @@ def discount_spec(self) -> specs.BoundedArray:
         ...
 
     @property
-    def unwrapped(self) -> Environment:
+    def unwrapped(self) -> Any:
```
WiemKhlifi, line 105: 🔥

RuanJohn (dismissed):
:1st_place_medal:

WiemKhlifi (changes requested):
Just some basic comments /suggestion 👀

pyproject.toml
sash-a: I think we want to keep it compatible with our lowest version that we support

mava/networks/torsos.py
sash-a: This is a weird one, `ruff` just added this for me during pre-commit, so I think it's one of it's rules :shrug: 

I think true is good if we're having the keyword there anyways

mava/types.py
sash-a: Ye because the unwrapped isn't necessarily a jumanji env, it could be anything e.g smax

WiemKhlifi: Yeeesss that's why I like it we're detaching from relaying on jumanji 💃

WiemKhlifi (approved):
Thanks @sash-a 🌟 🦸

RuanJohn (approved):
:fire:

PR #1111, opened 2024-10-25
feat: smaller networks for tests

sash-a:
## What?
Make test run faster by using tiny networks
## Why?
Faster is better?

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -12,37 +12,67 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict
+from typing import Any, Dict, List, TypeAlias
 
 import pytest
 
+_ConfSingleValue: TypeAlias = bool | int | float
+ConfigValue: TypeAlias = _ConfSingleValue | List[_ConfSingleValue] | Dict[str, _ConfSingleValue]
+
 
 @pytest.fixture
-def fast_config() -> Dict[str, Dict[str, bool | int | float]]:
+def fast_config_modifications() -> Dict[str, ConfigValue]:
     return {
-        "system": {
-            # common
-            "num_updates": 2,
-            "rollout_length": 1,
-            "num_minibatches": 1,
-            "update_batch_size": 1,
-            # ppo:
-            "ppo_epochs": 1,
-            # sac:
-            "explore_steps": 1,
-            "epochs": 1,  # also for iql
-            "policy_update_delay": 1,
-            "buffer_size": 8,  # also for iql
-            "batch_size": 1,
-            # iql
-            "min_buffer_size": 4,
-            "sample_batch_size": 1,
-            "sample_sequence_length": 1,
-        },
-        "arch": {
-            "num_envs": 1,
-            "num_eval_episodes": 1,
-            "num_evaluation": 1,
-            "absolute_metric": False,
-        },
+        # ---------- system config ---------
+        # common
+        "num_updates": 2,
+        "rollout_length": 1,
+        "num_minibatches": 1,
+        "update_batch_size": 1,
+        # ppo
+        "ppo_epochs": 1,
+        # sac
+        "explore_steps": 1,
+        "epochs": 1,  # also for iql
+        "policy_update_delay": 1,
+        "buffer_size": 8,  # also for iql
+        "batch_size": 1,
+        # iql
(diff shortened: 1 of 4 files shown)


sash-a:
Looks like it's ~1 minute faster :racehorse:

test/conftest.py
```diff
@@ -12,37 +12,67 @@
...
 
 import pytest
 
+_ConfSingleValue: TypeAlias = bool | int | float
+ConfigValue: TypeAlias = _ConfSingleValue | List[_ConfSingleValue] | Dict[str, _ConfSingleValue]
+
 
 @pytest.fixture
-def fast_config() -> Dict[str, Dict[str, bool | int | float]]:
+def fast_config_modifications() -> Dict[str, ConfigValue]:
```
WiemKhlifi: A stupid question: why `fast_config_modifications` instead of  `fast_config` 🙄

test/conftest.py
```diff
@@ -12,37 +12,67 @@
...
+
+    Args:
+        d: Dictionary to search through
+        replacements: The keys and values to replace
+    """
+
+    def _find_replace_recursive(current_dict: Dict[str, ConfigValue]) -> Dict[str, ConfigValue]:
+        """Helper function that recursively searches and replaces values."""
+        x = {}
+        for k, v in list(current_dict.items()):
```
WiemKhlifi: Few things I remember from my PS classes is that converting these to a list slows down your code a bit 👀 
```suggestion
        for k, v in current_dict.items():
```

test/conftest.py
```diff
@@ -12,37 +12,67 @@
...
+            if isinstance(v, dict):
+                v = _find_replace_recursive(v)
+            elif k in replacements:
+                v = replacements[k]
+
+            x[k] = v
+
+        return x
```
WiemKhlifi: Can't we just replace elements in the `current_dict`? ( tested  & it works)
```suggestion
            if isinstance(v, dict):
                current_dict[k] = _find_replace_recursive(v)
            elif k in replacements:
                current_dict[k] = replacements[k]

        return current_dict
```

test/integration_test.py
```diff
@@ -19,6 +19,8 @@
 from hydra import compose, initialize
 from omegaconf import DictConfig, OmegaConf
 
+from test.conftest import find_replace
```
WiemKhlifi: I'm not sure if this is good practice to import from `conftest` can't we just move this function into the `test/integration_test.py` or create a file for helper fcts?

WiemKhlifi (changes requested):
Thanks @sash-a few suggestions but all LGTM 🥳

RuanJohn (approved):
Thanks for this! :raised_hands:

test/integration_test.py
sash-a: Ye agreed, it felt a bit weird to me, I'll move it :smile:

test/conftest.py
```diff
@@ -12,37 +12,67 @@
...
+        # sac
+        "explore_steps": 1,
+        "epochs": 1,  # also for iql
+        "policy_update_delay": 1,
+        "buffer_size": 8,  # also for iql
+        "batch_size": 1,
+        # iql
+        "min_buffer_size": 4,
+        "sample_batch_size": 1,
+        "sample_sequence_length": 1,
```
RuanJohn: Just so it plays nice with `rec_iql` in the tests. 

```suggestion
        "sample_sequence_length": 2,
```

RuanJohn (changes requested):
Just the update to the sample sequence length in the testing config for iql please.

test/conftest.py
sash-a: switched back

test/conftest.py
```diff
@@ -12,37 +12,41 @@
...
-from typing import Dict
+from typing import Dict, List, TypeAlias
+from test.utils import ConfigValue
 
 import pytest
 
 
 @pytest.fixture
-def fast_config() -> Dict[str, Dict[str, bool | int | float]]:
+def fast_config() -> Dict[str, ConfigValue]:
```
WiemKhlifi, line 23: I like this packing 😅

RuanJohn (approved):
:ship:

WiemKhlifi (approved):
Thanks @sash-a 🥳

PR #1112, opened 2024-10-25
feat: uv for github actions

sash-a:
## What?
Use UV for github actions
## Why?
Should make them a bit faster


RuanJohn (approved):
Need for speed! :sunglasses:

WiemKhlifi (approved):
You're a star 🌟 😎

PR #1114, opened 2024-10-28
Fix lbf and rware obs spec types

RuanJohn:
## What?
The spec was not getting correctly extended when the environment had a global state. It was instead only updating the spec of `global_state` while leaving `agents_view` as the default. 

## Extra
Taking the spec casting away everywhere does not fix the LLVM error. So this is the best thing to do for now.


WiemKhlifi (approved):
LGTM 🎆

sash-a (approved):
:shield:

PR #1115, opened 2024-10-28
feat: updated dockerfile

sash-a:
## What?
Updated the docker file and makefile
## Why?
Needed a fresh coat of paint :smile: 
## How?
Using a different method to check if a GPU is available (thanks Claude) and using uv to install everything


Dockerfile
```diff
@@ -1,45 +1,23 @@
...
+COPY . .
 
-RUN echo "Installing requirements..."
-RUN pip install --quiet --upgrade pip setuptools wheel &&  \
-    pip install -e .
+RUN uv pip install -e .
 
-# Need to use specific cuda versions for jax
-ARG USE_CUDA=true
+ARG USE_CUDA=false
```
WiemKhlifi, line 18: I'm not an expert in dockerfiles but I believe this should be updated using the makefile right?😅

Makefile
```diff
@@ -1,19 +1,8 @@
...
-# For Windows use CURDIR
-ifeq ($(PWD),)
-PWD := $(CURDIR)
-endif
+# Check if GPU is available - if `nvidia-smi` works then use GPUs
+GPUS := $(shell command -v nvidia-smi > /dev/null && nvidia-smi > /dev/null 2>&1 && echo "--gpus all" || echo "")
 
 # Set flag for docker run command
-BASE_FLAGS=-it --rm  -v ${PWD}:/home/app/mava -w /home/app/mava
+BASE_FLAGS=-it --rm -v ${PWD}:/home/app/mava -w /home/app/mava
```
WiemKhlifi: A suggestion for running an example is to set a default example like the ff_ippo, but it can be ignored  (tested & works) 👀 
```python
example="mava/systems/ppo/anakin/ff_ippo.py"
run:
	@if [ -z "$(example)" ]; then echo "Please set a correct 'example'"; exit 1; fi;
	$(DOCKER_RUN) python $(example)
```

Dockerfile
```diff
@@ -1,45 +1,23 @@
...
 RUN if [ "$USE_CUDA" = true ] ; \
-    then pip install "jax[cuda12]==0.4.30" ; \
+    then uv pip install jax[cuda12]==0.4.30 ; \
```
WiemKhlifi, line 20: I have a bad network to test this now but I believe this didn't give any issue on your cuda? 😢

Dockerfile
```diff
@@ -1,45 +1,23 @@
...
-COPY ./requirements ./requirements
-COPY setup.py .
-COPY README.md .
-COPY mava/version.py mava/version.py
+COPY . .
 
-RUN echo "Installing requirements..."
-RUN pip install --quiet --upgrade pip setuptools wheel &&  \
-    pip install -e .
+RUN uv pip install -e .
```
WiemKhlifi, line 16: This `uv` is really awesome 😎 🔥

WiemKhlifi (changes requested):
Thanks @sash-a this really looks good 🔥  I only have few basic comments as newbie 🤓

Dockerfile
```diff
@@ -1,45 +1,23 @@
-FROM nvidia/cuda:11.8.0-cudnn8-runtime-ubuntu22.04
+# Stage 1: Build environment
+FROM python:3.12-slim as core
```
WiemKhlifi: NIT based on [this recommendation](https://docs.docker.com/reference/build-checks/from-as-casing/): 
```suggestion
FROM python:3.12-slim AS core
```

Dockerfile
sash-a: Yup, but we are basically setting a default here, because you could run this without the Makefile and then forget to pass `USE_CUDA`

Dockerfile
sash-a: Ye I've double checked and it seems to work

Makefile
sash-a: I think we rather just add an example in the docs for how to use the dockerfile

RuanJohn: I think we don't have to use volumes. 

```suggestion
BASE_FLAGS=-it --rm
```

Dockerfile
arnupretorius: But wouldn't this overwrite the argument being passed in by the makefile?

sash-a: As long as you pass it in it will overwrite it, this is just a default, so:
```
docker build --build-arg USE_CUDA=<true/false>
```
will overwrite it

RuanJohn (approved):
:motor_boat:

Dockerfile
WiemKhlifi: Yeah I tested this on my wsl and it seems to install the cuda version of Jax correctly 👍

WiemKhlifi (approved):
Thanks @sash-a for this new coat of paint 🚋 🕺 💃

PR #1116, opened 2024-10-28
fix: fix logging during evaluation for JaxMARL envs

WiemKhlifi:
## What?
Fix the logging of the metrics during the evaluation phase. This is affecting MPE in the first place but also slightly affected SMAX's `win_rate`.
## Why?
Based on the offline conservation this avoids getting stuck at 0 `episode_return`/ `win_rate` when using the `RecordEpisodeMetrics` wrapper for evaluation 🚒


RuanJohn (approved):
Thanks @WiemKhlifi :raised_hands:

sash-a (approved):
:100:

PR #1120, opened 2024-10-30
Add vector connector wrapper

RuanJohn:
## What?
Adds a vector observation wrapper for connector. 
## Extra
* It turns out Mava is already using MaConnector. 
* This PR takes away the need for passing `env_name` to the env make functions in `mava/utils/make_env.py`. We now also use `config.env.env_name` to select the wrapper but will always instantiate environments with `config.env.scenario.env_name`. This allows us to use `config.env.env_name` for wrapper selection and logging while keeping track of exact env versions via the scenario configs.


mava/configs/env/connector.yaml
```diff
@@ -10,6 +10,9 @@ env_name: MaConnector # Used for logging purposes.
 # This metric is returned at the end of an experiment and can be used for hyperparameter tuning.
 eval_metric: episode_return
 
+# Whether the environment observations will be a feature vector or 2-dimensional
+observation_type: vector # [pixel, vector]
```
RuanJohn: Note: I am not sure that `pixel` is the correct word here. Perhaps `grid_based` is better? But keen to hear what people think.

mava/utils/make_env.py
```diff
@@ -112,6 +121,8 @@ def make_jumanji_env(
     train_env = wrapper(train_env, add_global_state=add_global_state)
     eval_env = wrapper(eval_env, add_global_state=add_global_state)
 
+    jax.debug.print("Implicit Agent ID: {x}", x=config.env.implicit_agent_id)
+
```
sash-a:
```suggestion
```

mava/wrappers/jumanji.py
```diff
@@ -311,6 +312,143 @@ def observation_spec(
...
+        def create_agents_view(grid: chex.Array) -> chex.Array:
+            positions = jnp.where(grid % TARGET == POSITION, True, False)
+            targets = jnp.where((grid % TARGET == 0) & (grid != EMPTY), True, False)
+            paths = jnp.where(grid % TARGET == PATH, True, False)
+
+            # group positions and paths
+            blockers = jnp.where(positions, 1, jnp.where(paths, -1, 0))
+
+            position_per_agent = jnp.where(grid == POSITION, True, False)
+            target_per_agent = jnp.where(grid == TARGET, True, False)
```
sash-a: No need for the `where`

```suggestion
            position_per_agent = grid == POSITION
            target_per_agent = grid == TARGET
```

mava/wrappers/jumanji.py
```diff
@@ -311,6 +312,143 @@ def observation_spec(
...
+        step_count = specs.BoundedArray(
+            (self.num_agents,),
+            int,
+            jnp.zeros(self.num_agents, dtype=int),
+            jnp.repeat(self.time_limit, self.num_agents),
+            "step_count",
+        )
+
+        tiles_in_fov = (self.fov * 2 + 1) ** 2
+        single_agent_obs = 4 + tiles_in_fov * 2
```
sash-a, line 426:
```suggestion
        # 2 sets of tiles in fov (blockers and targets) + xy position of agent and target
        tiles_in_fov = (self.fov * 2 + 1) ** 2
        single_agent_obs = 4 + tiles_in_fov * 2
```

mava/utils/make_env.py
```diff
@@ -103,7 +105,14 @@ def make_jumanji_env(
...
     generator = generator(**config.env.scenario.task_config)
-    wrapper = _jumanji_registry[env_name]["wrapper"]
+
+    # If observation_type is in the config use the corresponding wrapper and make sure
+    # implicit_agent_id is set to False.
+    if hasattr(config.env, "observation_type"):
+        wrapper = _jumanji_registry[env_name]["wrapper"][config.env.observation_type]
+        config.env.implicit_agent_id = False
+    else:
+        wrapper = _jumanji_registry[env_name]["wrapper"]
```
sash-a: Not a fan of this, I think I would prefer to just make this a separate env as we discussed this morning. So all envs only ever have a single wrapper. 

If you do this we should re-use the same scenario, just have another env config and then just add that env config to the tests :pray:

sash-a (changes requested):
Mostly minor things, big one is to have a separate env for vector connector

mava/utils/make_env.py
RuanJohn: Reworked a bit to now have a separate env config for vector connector. I also added vector connector to the integration tests.

sash-a (dismissed):
:star:

mava/utils/make_env.py
```diff
@@ -46,23 +46,24 @@
...
+        "wrapper": VectorConnectorWrapper,
     },
-    "Cleaner-v0": {"generator": CleanerRandomGenerator, "wrapper": CleanerWrapper},
+    "Cleaner": {"generator": CleanerRandomGenerator, "wrapper": CleanerWrapper},
 }
 
 # Registry mapping environment names directly to the corresponding wrapper classes.
 _matrax_registry = {"Matrax": MatraxWrapper}
-_jaxmarl_registry: Dict[str, Type[JaxMarlWrapper]] = {"Smax": SmaxWrapper, "MaBrax": MabraxWrapper}
+_jaxmarl_registry = {"Smax": SmaxWrapper, "MaBrax": MabraxWrapper}
```
WiemKhlifi, line 66: 💃

mava/wrappers/jumanji.py
```diff
@@ -311,6 +312,144 @@ def observation_spec(
...
+
+# get location coordinates from 2D grid
+def _get_location(grid: chex.Array) -> chex.Array:
+    row_len = grid.shape[-1]
+    index = jnp.argmax(grid)
+    return jnp.asarray((jnp.floor(index / row_len), jnp.remainder(index, row_len)), dtype=int)
+
+
+class VectorConnectorWrapper(JumanjiMarlWrapper):
+    """Multi-agent wrapper for the MA Connector environment.
```
WiemKhlifi:
```suggestion
    """Multi-agent wrapper for the MaConnector environment that transforms multi-dimensional observation into vector-based one.

WiemKhlifi (changes requested):
Thanks @RuanJohn, I gave a suggestion for doc else rest is good to go 🥳

mava/wrappers/jumanji.py
WiemKhlifi: This is what came to my mind as a succinct idea to say that 💭

sash-a (approved):
:loudspeaker:

WiemKhlifi (approved):
🥳 🌟

PR #1121, opened 2024-10-30
More rware scenario configs

RuanJohn:
## What?
Simple PR that adds all the RWARE scenario configs from the Sable paper.


sash-a (approved):
:handshake:

WiemKhlifi (approved):
🤖

PR #1123, opened 2024-11-01
Chore: Add asserts for number of envs divisible by number of minibatches

SimonDuToit:
Add an assert in each PPO system to check that the number of envs is divisible by the number of minibatches.


sash-a (changes requested):
Could you also add this to MAT please :pray:

sash-a (approved):
:star2:

RuanJohn (approved):
:raised_hands:

PR #1124, opened 2024-11-01
chore: PPO system cleanup

sash-a:
## What?
Did a couple small clean ups to the PPO systems that are long overdue:
* Removed all the capital comments
* Removed opt states from the losses
* Removed tree map around the done reshape during acting (it's just an array)
* Removed metrics from the traj batch because they aren't really part of the batch
* Cleaner unpacking of the loss metrics

@RuanJohn should we add this for MAT?


mava/systems/mat/anakin/mat.py
```diff
@@ -83,51 +84,35 @@ def _update_step(learner_state: LearnerState, _: Any) -> Tuple[LearnerState, Tup
...
-            )
+            done = timestep.last().repeat(env.num_agents).reshape(config.arch.num_envs, -1)
             transition = PPOTransition(
-                done,
-                action,
-                value,
-                timestep.reward,
-                log_prob,
-                last_timestep.observation,
-                info,
```
RuanJohn, line 120: A slight concern I have with not storing info in the transition is that we lose the ability to quickly edit systems that could need it in the future. What do you think?

mava/systems/mat/anakin/mat.py
```diff
@@ -208,28 +190,27 @@ def _loss_fn(
                         )
                         * gae
                     )
-                    loss_actor = -jnp.minimum(loss_actor1, loss_actor2)
-                    loss_actor = loss_actor.mean()
+                    actor_loss = -jnp.minimum(loss_actor1, loss_actor2)
```
RuanJohn: If we change from `loss_actor` -> `actor_loss` can we please do the same for `loss_actor1` and `loss_actor2`

mava/systems/mat/anakin/mat.py
```diff
@@ -248,15 +229,11 @@ def _loss_fn(
...
-                # UPDATE ACTOR PARAMS AND OPTIMISER STATE
+                # Update params and optimiser state
                 actor_updates, new_opt_state = actor_update_fn(actor_grads, opt_state)
                 new_params = optax.apply_updates(params, actor_updates)
 
-                # PACK LOSS INFO
-                total_loss = actor_loss_info[0]
-                value_loss = actor_loss_info[1][2]
-                actor_loss = actor_loss_info[1][0]
-                entropy = actor_loss_info[1][1]
```
RuanJohn, line 259: Why do we no longer need to use this?

mava/systems/ppo/anakin/ff_ippo.py
```diff
@@ -253,39 +232,36 @@ def _critic_loss_fn(
...
-                entropy = actor_loss_info[1][1]
+                actor_loss, (raw_actor_loss, entropy) = actor_loss_info
+                critic_loss, raw_critic_loss = critic_loss_info
+
+                total_loss = actor_loss + critic_loss
                 loss_info = {
                     "total_loss": total_loss,
-                    "value_loss": value_loss,
-                    "actor_loss": actor_loss,
+                    "value_loss": raw_critic_loss,
```
RuanJohn: If we change value -> critic I think we should be consistent with the naming everywhere. Are you keeping it as value to still be comparable with old experiments?

mava/systems/ppo/anakin/ff_ippo.py
```diff
@@ -253,39 +232,36 @@ def _critic_loss_fn(
...
-                # PACK NEW PARAMS AND OPTIMISER STATE
                 new_params = Params(actor_new_params, critic_new_params)
                 new_opt_state = OptStates(actor_new_opt_state, critic_new_opt_state)
 
-                # PACK LOSS INFO
-                total_loss = actor_loss_info[0] + critic_loss_info[0]
-                value_loss = critic_loss_info[1]
-                actor_loss = actor_loss_info[1][0]
-                entropy = actor_loss_info[1][1]
+                actor_loss, (raw_actor_loss, entropy) = actor_loss_info
```
RuanJohn: No so sure about appending `raw_` to the loss values. I think `actor_loss` and `raw_actor_loss` are the same in this case as well. So we could always just catch one of them with a `_` and used the other.

mava/systems/ppo/anakin/ff_mappo.py
```diff
@@ -74,39 +73,36 @@ def _update_step(learner_state: LearnerState, _: Any) -> Tuple[LearnerState, Tup
...
+            # Step environment
             env_state, timestep = jax.vmap(env.step, in_axes=(0, 0))(env_state, action)
 
-            # LOG EPISODE METRICS
-            done = tree.map(
-                lambda x: jnp.repeat(x, config.system.num_agents).reshape(config.arch.num_envs, -1),
-                timestep.last(),
-            )
-            info = timestep.extras["episode_metrics"]
+            done = timestep.last().repeat(env.num_agents).reshape(config.arch.num_envs, -1)
```
RuanJohn, line 92: I like this!

mava/systems/ppo/anakin/ff_mappo.py
```diff
@@ -144,23 +140,21 @@ def _update_epoch(update_state: Tuple, _: Any) -> Tuple:
 
             def _update_minibatch(train_state: Tuple, batch_info: Tuple) -> Tuple:
                 """Update the network for a single minibatch."""
-                # UNPACK TRAIN STATE AND BATCH INFO
                 params, opt_states, key = train_state
                 traj_batch, advantages, targets = batch_info
 
                 def _actor_loss_fn(
                     actor_params: FrozenDict,
-                    actor_opt_state: OptState,
```
RuanJohn, line 153: Long overdue!

RuanJohn (changes requested):
Thank you @sash-a! I think these are some very needed changes. I just have a few suggestions and questions. Any comment applies to all systems.

sash-a:
todo: same fixes for sable and switch sable to using the PPOTransition

README.md
```diff
@@ -184,10 +184,6 @@ python mava/systems/ff_ippo.py env=rware env/scenario=tiny-4ag
 
 Additionally, we also have a [Quickstart notebook][quickstart] that can be used to quickly create and train your first Multi-agent system.
 
-## Advanced Usage 👽
```
WiemKhlifi, line 187: Are we giving up on this because it's no longer maintained?

sash-a: Ya @RuanJohn and I discussed it. I think if we want to bring it back at some point it would be nice, but right now it's not maintained and I'm not even sure if it works right now

WiemKhlifi: Indeed I agree on that it has some issues within the evaluator as well, hpefully we can restore it in future 🤞

OmaymaMahjoub (dismissed):
LGTM! Let's just wait for the benchmark before we merge this 🙏

RuanJohn (dismissed):
:trophy:

OmaymaMahjoub (approved):
Thanks @sash-a 🙌

RuanJohn (approved):
:star2:

PR #1130, opened 2024-11-08
fix: limit timestep-pos-encoding to rec-Sable

OmaymaMahjoub:
## What?
Minor fix to limit the timestep positional encoding to only rec-sable.

## How?
Add a timestep flag checker in recurrent apply call

### DIFF
--- a/mava/networks/retention.py
+++ b/mava/networks/retention.py
@@ -301,7 +301,8 @@ class MultiScaleRetention(nn.Module):
         B, S, _ = value_n.shape
 
         # Positional encoding of the current step
-        key_n, query_n, value_n = self.pe(key_n, query_n, value_n, step_count)
+        if self.memory_config.timestep_positional_encoding:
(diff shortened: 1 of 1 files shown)


lattaai4:
Hello, I tried to fix the issue

This is what I did:

Modify the SableNetwork class to include a flag for timestep positional encoding and only apply it in the recurrent version. Update the __call__ and recurrent methods accordingly.

I also created pull request: https://github.com/instadeepai/Mava/pull/1131


> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

sash-a (approved):
Nice catch :smile:

RuanJohn (approved):
:fire:

PR #1131, opened 2024-11-08
Autofix: Fix: Limit timestep-pos-encoding to Rec-Sbale

lattaai4:
Modify the SableNetwork class to include a flag for timestep positional encoding and only apply it in the recurrent version. Update the __call__ and recurrent methods accordingly. 
> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

### DIFF
--- a/mava/networks/sable_network.py
+++ b/mava/networks/sable_network.py
@@ -43,6 +43,8 @@ class EncodeBlock(nn.Module):
     n_agents: int
 
     def setup(self) -> None:
+        if self.use_timestep_encoding:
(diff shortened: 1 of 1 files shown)


sash-a:
Not quite sure what this is, but we already have a PR out for this :smile: 

If you want to contribute please take a look at our issues

PR #1134, opened 2024-11-11
Feat: Support latest Jumanji version

WiemKhlifi:
## What?
Upgrade to the latest Jumanji version of `1.0.1` instead of `0.3.1` and pin to the original and latest `Jumanji` and `Matrax`.
## How?
- Change the `requirements.txt` to use original versions instead of a fork.
- Adapt all the wrappers and systems to use cached specs similar to jumanji wrappers and envs.

# Extra:
- Note that this PR will be merged after pushing [this PR](https://github.com/instadeepai/jumanji/pull/263) into jumanji for Connector env updates.
- Note that in some wrappers, since the specs outputs will be cached, some attributes can't be retrieved if defined after the `super().__init__(env)`  ( The `self.__getattr__(env,name)` in parent class can't get the attribute from env if it's defined with different name in the env wrapper class). 
For example:

```python
# Should work:
super().__init__(env)
self.time_limit = self._env.time_limit 

# Shouldn't work:
super().__init__(env)
self.time_limit = self._env.max_episode_length
```


mava/configs/env/connector.yaml
```diff
@@ -4,7 +4,7 @@ defaults:
   - scenario: con-5x5x3a # [con-5x5x3a, con-7x7x5a, con-10x10x10a, con-15x15x23a]
 # Further environment config details in "con-10x10x5a" file.
 
-env_name: MaConnector # Used for logging purposes.
+env_name: Connector # Used for logging purposes.
```
RuanJohn, line 7: Don't we use MaConnector?

mava/wrappers/jumanji.py
```diff
@@ -222,15 +227,20 @@ class ConnectorWrapper(JumanjiMarlWrapper):
...
         super().__init__(env, add_global_state)
-        self._env: MaConnector
+        self._env: Connector
 
     def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
         """Modify the timestep for the Connector environment."""
 
         # TARGET = 3 = The number of different types of items on the grid.
         def create_agents_view(grid: chex.Array) -> chex.Array:
+            # add switch_perspective to change grid (@sasha-a what do you think?):
```
RuanJohn: Please remove this comment. Also @sash-a can you please comment here?

mava/wrappers/jumanji.py
```diff
@@ -245,18 +255,9 @@ def create_agents_view(grid: chex.Array) -> chex.Array:
...
-        def aggregate_rewards(
-            timestep: TimeStep,
-        ) -> TimeStep[Observation]:
-            """Aggregate individual rewards and discounts across agents."""
-            team_reward = jnp.sum(timestep.reward)
-            reward = jnp.repeat(team_reward, self.num_agents)
-            return timestep.replace(reward=reward)
-
-        timestep = aggregate_rewards(timestep)
-
```
RuanJohn, line 259: If we were aggregating before, we should probably be aggregating again now to keep the logic consistent. Unless Jumanji now does something differently?

RuanJohn (changes requested):
Thanks for this @WiemKhlifi! Just a few questions, but it looks mostly good to me.
As a sanity check can you please do a few test runs to just check that the system performance is unaffected?

mava/wrappers/jumanji.py
WiemKhlifi: Yeah Jumanji already returns the sum of rewards and the diffrence between Connector and Maconnector is the individual reward the observation or the thing that @sash-a added for `switch_perspective` [function](https://github.com/sash-a/jumanji/blob/f102941b0b5eb4302dbe624048617803a07fe87a/jumanji/environments/routing/connector/utils.py#L182)

WiemKhlifi: @sash-a probably need some help on deciding whether or not to discard this utility since the paper version uses it and if not should it be included in mava only?

mava/wrappers/jumanji.py
WiemKhlifi: @sash-a I explained it all below 😅

requirements/requirements.txt
```diff
@@ -10,9 +10,9 @@ id-marl-eval @ git+https://github.com/instadeepai/marl-eval
 jax==0.4.30
 jaxlib==0.4.30
 jaxmarl @ git+https://github.com/RuanJohn/JaxMARL@unpin-jax  # This only unpins the version of Jax.
-jumanji @ git+https://github.com/sash-a/jumanji@old_jumanji  # Includes a few extra MARL envs
+jumanji @ git+https://github.com/instadeepai/jumanji
```
sash-a: Jumanji is on pypi, so we don't need to point to a git repo
```suggestion
jumanji
```

requirements/requirements.txt
```diff
@@ -10,9 +10,9 @@ id-marl-eval @ git+https://github.com/instadeepai/marl-eval
...
 lbforaging
-matrax @ git+https://github.com/instadeepai/matrax@4c5d8aa97214848ea659274f16c48918c13e845b
+matrax @ git+https://github.com/instadeepai/matrax
```
sash-a: Similarly for matrax, although I think we still need to release that one
```suggestion
matrax
```

requirements/requirements.txt
sash-a, line 1: Can you also remove the git stuff from marl-eval, because it should also be on pypi

mava/wrappers/jumanji.py
```diff
@@ -266,25 +291,23 @@ def aggregate_rewards(
...
     ) -> specs.Spec[Union[Observation, ObservationGlobalState]]:
         """Specification of the observation of the environment."""
-        step_count = specs.BoundedArray(
-            (self.num_agents,),
-            int,
-            jnp.zeros(self.num_agents, dtype=int),
-            jnp.repeat(self.time_limit, self.num_agents),
-            "step_count",
-        )
+        obs_spec = self._env.observation_spec
```
sash-a: Don't we need to repeat here?

mava/systems/sable/anakin/ff_sable.py
```diff
@@ -372,17 +372,13 @@ def learner_setup(
...
-
     # PRNG keys.
     key, net_key = keys
 
     # Get number of agents and actions.
     action_dim = env.action_dim
-    n_agents = env.action_spec().shape[0]
+    n_agents = env.num_agents
     config.system.num_agents = n_agents
-    config.system.num_actions = action_dim
```
sash-a, line 341: Why remove this from the config?

mava/systems/sable/anakin/rec_sable.py
```diff
@@ -417,9 +417,8 @@ def learner_setup(
 
     # Get number of agents and actions.
     action_dim = env.action_dim
-    n_agents = env.action_spec().shape[0]
+    n_agents = env.num_agents
     config.system.num_agents = n_agents
-    config.system.num_actions = action_dim
```
sash-a, line 368: Same here?

sash-a (changes requested):
Thanks Wiem, couple small things, mostly removing the git stuff from requirements.txt where possible

requirements/requirements.txt
WiemKhlifi: Amazing! I'll update it 🙌

mava/systems/sable/anakin/ff_sable.py
WiemKhlifi: This was from a copy of old ppo but now since we call `action_dim` only in the `learner_setup` and we're not using it in other parts of code like the `num_agents` I removed it from here

requirements/requirements.txt
WiemKhlifi: Yep with the newer version since now it's 0.0.4 which depends on old jumanji it will be the same issue :/
Can we do that or Arnu should?

mava/wrappers/jumanji.py
WiemKhlifi: Nice catch! When I removed this forgot that Sable uses the `step_count` but other systems don't and none vec connector isn't for sable so didn't get any issues 🫠 
I'll revert (but we're not using this xd)!

requirements/requirements.txt
WiemKhlifi: Same issue I guess but a little bit different from matrax, this is given by the CI:
image

WiemKhlifi: _**Running pytest locally in new env**_ 🫠 
image

mava/configs/env/vector-connector.yaml
```diff
@@ -4,7 +4,10 @@ defaults:
   - scenario: con-5x5x3a # [con-5x5x3a, con-7x7x5a, con-10x10x10a, con-15x15x23a]
 # Further environment config details in "con-10x10x5a" file.
 
-env_name: VectorMaConnector # Used for logging purposes.
+env_name: VectorConnector # Used for logging purposes.
+
+# Choose whether to aggregate the list of individual rewards and use the team reward (default setting) OR use_individual_rewards=True.
```
RuanJohn:
```suggestion
# Choose whether to aggregate individual rewards into a shared team reward or not.
```

mava/wrappers/jumanji.py
```diff
@@ -39,12 +40,25 @@
...
+    """Aggregate individual rewards across agents."""
+    if use_individual_rewards:
+        # Returns a list of individual rewards that will be used as is.
+        return reward
+
+    # Aggregate the list of individual rewards and use a single team_reward.
+    team_reward = jnp.sum(reward)
+    return jnp.repeat(team_reward, num_agents)
+
+
```
RuanJohn:
```suggestion
def aggregate_rewards(
    reward: chex.Array, num_agents: int
) -> chex.Array:
    """Aggregate individual rewards across agents."""
    team_reward = jnp.sum(reward)
    return jnp.repeat(team_reward, num_agents)
```

mava/wrappers/jumanji.py
```diff
@@ -197,40 +203,67 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
             action_mask=timestep.observation.action_mask,
             step_count=jnp.repeat(timestep.observation.step_count, self.num_agents),
         )
-        if self._use_individual_rewards:
-            # The environment returns a list of individual rewards and these are used as is.
-            return timestep.replace(observation=modified_observation)
+        # Whether or not aggregate the list of individual rewards.
+        reward = aggregate_rewards(timestep.reward, self.num_agents, self._use_individual_rewards)
```
RuanJohn: I must admit I am not a massive fan of the `not` here. But I prefer it over having the conditional in the aggregation function. What do you think of just having the config option be `aggregate_rewards` instead of `use_individual_rewards`? Then we could change the conditional here to `if self._aggregate_rewards`.
```suggestion
        if not self._use_individual_rewards:
            reward = aggregate_rewards(timestep.reward, self.num_agents)
```

mava/wrappers/jumanji.py
RuanJohn: I think it is a bit more clear to have the conditional in the wrapper and then to just let this function do the aggregation.

mava/wrappers/jumanji.py
```diff
@@ -266,14 +289,17 @@ def aggregate_rewards(
         # The episode is won if all agents have connected.
         extras = timestep.extras | {"won_episode": timestep.extras["ratio_connections"] == 1.0}
 
-        return timestep.replace(observation=Observation(**obs_data), extras=extras)
+        # Whether or not aggregate the list of individual rewards.
+        reward = aggregate_rewards(timestep.reward, self.num_agents, self._use_individual_rewards)
```
RuanJohn: Same about the`not` as in my previous comment. 
```suggestion
        if not self._use_individual_rewards:
            reward = aggregate_rewards(timestep.reward, self.num_agents, self._use_individual_rewards)
```

mava/wrappers/jumanji.py
```diff
@@ -197,40 +203,67 @@ def modify_timestep(self, timestep: TimeStep) -> TimeStep[Observation]:
...
-        inner_spec = super().observation_spec()
+        inner_spec = super().observation_spec
         spec = inner_spec.replace(agents_view=inner_spec.agents_view.replace(dtype=float))
         if self.add_global_state:
             spec = spec.replace(global_state=inner_spec.global_state.replace(dtype=float))
 
         return spec
 
 
+def switch_perspective(grid: chex.Array, agent_id: int, num_agents: int) -> chex.Array:
```
RuanJohn, line 219: Nice docstring!

mava/wrappers/jumanji.py
```diff
@@ -407,8 +439,11 @@ def _create_one_agent_view(i: int) -> chex.Array:
         # The episode is won if all agents have connected.
         extras = timestep.extras | {"won_episode": timestep.extras["ratio_connections"] == 1.0}
 
-        return timestep.replace(observation=Observation(**obs_data), extras=extras)
+        # Whether or not aggregate the list of individual rewards.
+        reward = aggregate_rewards(timestep.reward, self.num_agents, self._use_individual_rewards)
```
RuanJohn:
```suggestion
        if not self._use_individual_rewards:
            reward = aggregate_rewards(timestep.reward, self.num_agents, self._use_individual_rewards)
```

RuanJohn (changes requested):
Thanks @WiemKhlifi. Some suggestions from my side.

mava/wrappers/jumanji.py
sash-a: eh either way I prefer this :smile:

WiemKhlifi: I chose the second option to use `aggregate_rewards`  instead of the not which is less confusing 😄

sash-a (dismissed):
:star:

requirements/requirements.txt
```diff
@@ -6,13 +6,13 @@ flax>=0.8.1
 gigastep @ git+https://github.com/mlech26l/gigastep
 gymnasium
 hydra-core==1.3.2
-id-marl-eval @ git+https://github.com/instadeepai/marl-eval
+id-marl-eval
```
RuanJohn: Since marl-eval hasn't had a release since recent changes can we please keep it as is. 
```suggestion
id-marl-eval @ git+https://github.com/instadeepai/marl-eval
```

RuanJohn (changes requested):
Just a marl-eval dependency suggestion.

requirements/requirements.txt
sash-a: It has, I did a release about a week ago

WiemKhlifi: @sash-a do you mean matrax?

sash-a: ah sorry I do mean matrax :facepalm:

sash-a (approved):

RuanJohn (approved):
:joystick:

PR #1140, opened 2024-11-13
chore: only run integration tests after merge to dev

sash-a:
Integration tests are taking too long :smile:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v3
         with:
-          version: "0.4.26"
+          version: "0.5.4"
           enable-cache: true
           cache-dependency-glob: "requirements/requirements**.txt"  # invalidate cache when requirements file changes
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,12 +1,12 @@
-name: Tests and Linters 🧪
+name: CI 🧪
```
RuanJohn, line 1: Just to keep with us called the intergration one `Integration Tests: 🧪`
```suggestion
name: Linters 🖌️
```

RuanJohn (changes requested):
Thanks @sash-a! Ironically the linters are failling :sweat_smile: 
And then just one small naming suggestion.

.github/workflows/ci.yaml
sash-a: I quite like CI, I took inspiration from [ruff](https://github.com/astral-sh/ruff/blob/main/.github/workflows/ci.yaml). They have all tests/builds/linting in `ci.yaml` which is what this can be if we every add unit tests for things like utils or other continuous integration things

RuanJohn: Ah, makes sense. Okay, let's keep CI in that case.

RuanJohn (approved):
:clap:

OmaymaMahjoub (approved):
Thanks @sash-a 🔥

PR #1147, opened 2024-11-29
fix: update ret_output shape

### LINKED ISSUE #1146: [BUG] output from multi-scale retention only contains output from the first head
### Describe the bug
[Here](https://github.com/instadeepai/Mava/blob/2074e7eccd3a7d34ee02e3fbc8c02915c859bb92/mava/networks/retention.py#L265-L291), ret_output is only filled with the output of the first retention head.

`ret_output` is declared with shape `(batch_size, chunk_size, head_size)`  but for every y of shape `(batch_size, chunk_size, head_size)` obtained from `self.retention_heads[head`] , we update ret_output  at slice `:, :, self.head_size * head : self.head_size * (head + 1)`.

Since ret_output  is of shape `(batch_size, chunk_size, head_size)` , won't it just update the entire tensor for `head = 0` but leave `ret_output`  unchanged for the subsequent head operations `head = {1, ..., n_head-1}` ?

### To Reproduce
run any sable (feed-forward and recurrent) system

### Expected behavior
`ret_output` contains output from all retention heads

ArnolFokam:
Closes #1146

### DIFF
--- a/mava/networks/retention.py
+++ b/mava/networks/retention.py
@@ -242,7 +242,7 @@ class MultiScaleRetention(nn.Module):
         self.w_o = self.param(
             "w_o",
             nn.initializers.normal(stddev=1 / self.embed_dim),
-            (self.head_size, self.embed_dim),
+            (self.embed_dim, self.embed_dim),
         )
         self.group_norm = nn.GroupNorm(num_groups=self.n_head)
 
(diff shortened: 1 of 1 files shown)


sash-a (dismissed):
Great spot, thanks @ArnolFokam :fire:

OmaymaMahjoub (approved):
Thanks @ArnolFokam  🔥

sash-a (approved):
:pray:

PR #1148, opened 2024-12-04
Feat Sebulba recurrent IQL

Louay-Ben-nessir:
## What?
A recurrent IQL implementation using the Sebulba architecture.
## Why?
Offline Sebulba base and non-jax envs in Mava.
## How?
Mixed the Sebulba structure from PPO with the learner code from Anakin IQL.


mava/configs/system/q_learning/rec_iql.yaml
```diff
@@ -11,13 +11,13 @@ add_agent_id: True
 min_buffer_size: 32
 update_batch_size: 1 # Number of vectorised gradient updates per device.
 
-rollout_length: 2 # Number of environment steps per vectorised environment.
+rollout_length:  2 # Number of environment steps per vectorised enviro²nment.
```
sash-a: Not sure how the ^2 got in there :joy: 
```suggestion
rollout_length: 2 # Number of environment steps per vectorised environment.
```

mava/systems/q_learning/types.py
```diff
@@ -21,22 +21,22 @@
 from jumanji.env import State
 from typing_extensions import NamedTuple, TypeAlias
 
-from mava.types import Observation
+from mava.types import MavaObservation, Observation
```
sash-a: Where do we still use `Observation`? Can we use `MavaObservation` everywhere?

mava/utils/config.py
```diff
@@ -34,10 +34,13 @@ def check_sebulba_config(config: DictConfig) -> None:
...
-        f"Number of training samples per evaluator ({num_eval_samples})"
-        + f"must be divisible by num_minibatches ({config.system.num_minibatches})."
-    )
+
+    # PPO specifique check
+    if "num_minibatches" in config.system:
+        assert num_eval_samples % config.system.num_minibatches == 0, (
+            f"Number of training samples per evaluator ({num_eval_samples})"
+            + f"must be divisible by num_minibatches ({config.system.num_minibatches})."
+        )
```
sash-a: A thought on this, maybe we can split these up into multiple methods e.g `check_num_updates`, `check_num_envs` etc. Then have a `check_sebulba_config_ppo`, `check_anakin_config_ppo` and a `check_sebulba_config_iql` which will use the relevant methods?

mava/utils/sebulba.py
```diff
@@ -16,17 +16,21 @@
...
 import jax.numpy as jnp
+import numpy as np
 from colorama import Fore, Style
+from flashbax import make_trajectory_buffer
 from jax import tree
 from jax.sharding import Sharding
 from jumanji.types import TimeStep
+from omegaconf import DictConfig
 
-# todo: remove the ppo dependencies when we make sebulba for other systems
```
sash-a, line 27: This is a good point though, maybe there's something we can do about it :thinking:

Maybe a protocol like that has `action, obs, reward`, not sure if there's any other common attributes?

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
     def __exit__(self, *args: Any) -> None:
         end = time.monotonic()
         self.to.append(end - self.start)
+
+
+# from https://github.com/EdanToledo/Stoix/blob/feat/sebulba-dqn/stoix/utils/rate_limiters.py
+class RateLimiter:
```
sash-a: This file is getting quite long, can we structure it like this:
```
-utils/
--sebulba/
---utils.py
---rate_limiters.py
---pipelines.py
```

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
...
+    """
+
+    def __init__(
+        self,
+        config: DictConfig,
+        learner_sharding: Sharding,
+        key: jax.random.PRNGKey,
+        rate_limiter: RateLimiter,
+        init_transition: Transition,
+        lifetime: ThreadLifetime,
```
sash-a: While we're here, I think we should actually replace the lifetime with `threading.Event`, [it's a python builtin](https://docs.python.org/3/library/threading.html#event-objects) that does exactly what we want. Although maybe that should be a separate PR, I'll make an issue

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
...
+            max_length_time_axis=config.system.buffer_size,
+            min_length_time_axis=config.system.min_buffer_size,
+        )
+        self.buffer_states = [rb.init(init_transition) for _ in range(self.num_buffers)]
+        self.buffer_adds_count = [0] * self.num_buffers
+
+        # Setup functions
+        self.buffer_add = jax.jit(rb.add, device=self.cpu)
+        self.buffer_sample = jax.jit(rb.sample, device=self.cpu)
+        self.move_to_device = lambda tree: jax.tree.map(lambda x: jax.device_put(x, self.cpu), tree)
```
sash-a: Seems like this is unused, would just be `device_get` anyway

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
...
+            self.rate_limiter.await_can_insert(timeout=QUEUE_PUT_TIMEOUT)
+        except TimeoutError:
+            print(
+                f"{Fore.RED}{Style.BRIGHT}Actor has timed out on insertion, "
+                f"this should not happen. A deadlock might be occurring{Style.RESET_ALL}"
+            )
+
+        # [Transition(num_envs)] * rollout_len -> Transition[done=(num_envs, rollout_len, ...)]
+        traj = _stack_trajectory(traj)
+        traj = jax.device_get(traj)
```
sash-a: Maybe be better to do the device get first so we block the device as little as possible? Although it's probably a tiny difference

```suggestion
        traj = jax.device_get(traj)
        traj = _stack_trajectory(traj)
```

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
...
+                f"{Fore.RED}{Style.BRIGHT}Learner has timed out on sampling, "
+                f"this should not happen. A deadlock might be occurring{Style.RESET_ALL}"
+            )
+
+        # Sample the data
+        # Potential deadlock risk here. Although it hasn't occurred during testing.
+        # if an unexplained deadlock happens, it is likely due to this section.
+        sampled_batch: List[Transition] = [
+            self.buffer_sample(state, sample_key).experience for state in self.buffer_states
+        ]
```
sash-a: Not sure I follow why this could deadlock?

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
...
+        transitions: Transition = jax.tree_map(lambda *x: np.concatenate(x), *sampled_batch)
```
sash-a: nit
```suggestion
        transitions: Transition = tree.map(lambda *x: np.concatenate(x), *sampled_batch)
```

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
...
+        transitions = jax.device_put(transitions, device=self.sharding)
+
+        self.rate_limiter.sample()
+
+        if not self._queue.empty():
```
sash-a: Can we rename this to `metrics_queue`, wasn't clear what this was storing

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
...
+        assert samples_per_insert > 0, "samples_per_insert must be greater than 0"
+
+        self.samples_per_insert = samples_per_insert
+        self.min_diff = min_diff
+        self.max_diff = max_diff
+        self.min_size_to_sample = min_size_to_sample
+
+        self.inserts = 0.0
+        self.samples = 0
+        self.deletes = 0
```
sash-a: Wonder if we need deletes here?

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
     def __exit__(self, *args: Any) -> None:
         end = time.monotonic()
         self.to.append(end - self.start)
+
+
+# from https://github.com/EdanToledo/Stoix/blob/feat/sebulba-dqn/stoix/utils/rate_limiters.py
+class RateLimiter:
+    def __init__(
+        self, samples_per_insert: float, min_size_to_sample: int, min_diff: float, max_diff: float
+    ):
```
sash-a: Can we please add a good doc string here please :pray:

mava/utils/sebulba.py
```diff
@@ -197,3 +203,357 @@ def __enter__(self) -> None:
...
+                min_size_to_sample * samples_per_insert + error_buffer
+              )
+            The offset is added so that the error tracked is for the insert/sample
+            ratio only takes into account operations occurring AFTER stage 1. If a
+            range (two float tuple) then the values are used without any offset.
+
+        Raises:
+          ValueError: If error_buffer is smaller than max(1.0, samples_per_inserts).
+        """
+        if isinstance(error_buffer, float) or isinstance(error_buffer, int):
```
sash-a: nit

```suggestion
        if isinstance(error_buffer, (int, float)):
```

mava/wrappers/gym.py
```diff
@@ -271,21 +272,27 @@ def step(self, action: list) -> TimeStep:
 
         ep_done = np.logical_or(terminated, truncated)
         step_type = np.where(ep_done, StepType.LAST, StepType.MID)
+        terminated = np.repeat(
+            terminated[..., np.newaxis], repeats=self.num_agents, axis=-1
+        )  # (B,) --> (B, N)
```
sash-a, line 277: Does this already happen for smax and lbf?

sash-a (changes requested):
I've looked through everything except the system file and it looks good, Sebulba utils especially! Just some relatively minor style changes

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+
+        obs = tree.map(lambda x: x[jnp.newaxis, ...], obs)
+        term_or_trunc = tree.map(lambda x: x[jnp.newaxis, ...], term_or_trunc)
+
+        next_hidden_state, eps_greedy_dist = q_net.apply(
+            params, hidden_state, (obs, term_or_trunc), eps
+        )
+
+        action = eps_greedy_dist.sample(seed=key)
+        action = action[0, ...]  # (1, B, A) -> (B, A)
```
sash-a: Bit safer as this will silently error if actions 0 dim is ever larger than 1

```suggestion
        action = eps_greedy_dist.sample(seed=key).squeeze(0)  # (B, A)
```

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+                        params, hstate_tpu, obs_tpu, last_dones, act_key, step_count
+                    )
+                    cpu_action = jax.device_get(action)
+
+                # Step environment
+                with RecordTimeTo(actor_timings["env_step_time"]):
+                    next_timestep = env.step(cpu_action)
+
+                # Prepare the transation
+                terminal = (1 - timestep.discount[..., 0, jnp.newaxis]).astype(bool)
```
sash-a, line 167: Are you sure we want to remove the agent dim here?

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+
+            def q_loss_fn(
+                q_online_params: FrozenVariableDict,
+                obs: Array,
+                term_or_trunc: Array,
+                action: Array,
+                target: Array,
+            ) -> Tuple[Array, Metrics]:
+                # axes switched here to scan over time
+                hidden_state, obs_term_or_trunc = prep_inputs_to_scannedrnn(obs, term_or_trunc)
```
sash-a, line 253: A general comment, I think this would be a lot easier to read if we used `done` to mean `term_or_trunc` which I think is a reasonable thing. Would have to make the change in anakin also though :/

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+                for source in params_sources:
+                    source.update(params.online)
+
+        # Pass all the metrics and  params to the main thread (evaluator) for logging and evaluation
+        if ep_metrics:
+            ep_metrics = tree.map(lambda *x: np.asarray(x), *ep_metrics)
+            train_metrics = tree.map(lambda *x: np.asarray(x), *train_metrics)
+
+        timing_dict = tree.map(lambda *x: np.mean(x), *rollout_times) | learn_times
+        timing_dict = tree.map(np.mean, timing_dict, is_leaf=lambda x: isinstance(x, list))
```
sash-a: 2 things, can we call this time_metrics and can you add a shape explainer comment as it's a bit hard to work out what is happening here

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+            term_or_trunc: Termination or truncation flag.
+            key: PRNG key for sampling.
+            t: Current timestep (used for epsilon decay).
+
+        Returns:
+            Tuple containing the chosen action, next hidden state, and updated timestep.
+        """
+
+        eps = jnp.maximum(
+            config.system.eps_min, 1 - (t / config.system.eps_decay) * (1 - config.system.eps_min)
```
sash-a, line 114: Would be nice if we could set a different decay per actor, although I think that's out of scope for this PR. Maybe if you could make an issue to add in some of the ape-X DQN features that would be great

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+        actor_timings: Dict[str, List[float]] = defaultdict(list)
+        with RecordTimeTo(actor_timings["rollout_time"]):
+            for _ in range(config.system.rollout_length):
+                with RecordTimeTo(actor_timings["get_params_time"]):
+                    params = params_source.get()  # Get the latest parameters from the learner
+
+                timestep = next_timestep
+                obs_tpu = tree.map(move_to_device, timestep.observation)
+
+                last_dones = tree.map(move_to_device, dones)
```
sash-a: `device_put` will automatically `tree_map`
```suggestion
                obs_tpu = move_to_device(timestep.observation)
                last_dones = move_to_device(dones)
```

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+) -> Tuple[
+    SebulbaLearnerFn[LearnerState, Transition],
+    RecQNetwork,
+    LearnerState,
+    Sharding,
+    Transition,
+]:
+    """Initialise learner_fn, network and learner state."""
+
+    # create temporory envoirnments.
```
sash-a: nit
```suggestion
    # create temporary environments.
```

mava/configs/system/q_learning/rec_iql.yaml
```diff
@@ -31,3 +31,7 @@ gamma: 0.99  # discount factor
 
 eps_min: 0.05
 eps_decay: 1e5
+
+# --- Sebulba parameters ---
+data_sample_mean: 150  # Average number of times the learner should sample each item from the replay buffer.
```
sash-a: Can we rather call this: `mean_data_sample_rate`. Wasn't clear to me what it was when I read it in the system file

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+    config.sample_per_insert = config.system.data_sample_mean * insert_to_sample_ratio
+    config.tolerance = config.sample_per_insert * config.system.error_tolerance
+
+    min_num_inserts = max(
+        config.system.sample_sequence_length // config.system.rollout_length,
+        config.system.min_buffer_size // config.system.rollout_length,
+        1,
+    )
+
+    rate_limiter = SampleToInsertRatio(config.sample_per_insert, min_num_inserts, config.tolerance)
```
sash-a: We should probably put them in the system config so it's easier to find on things like neptune

```suggestion
    config.system.sample_per_insert = config.system.data_sample_mean * insert_to_sample_ratio
    config.system.tolerance = config.sample_per_insert * config.system.error_tolerance

    min_num_inserts = max(
        config.system.sample_sequence_length // config.system.rollout_length,
        config.system.min_buffer_size // config.system.rollout_length,
        1,
    )

    rate_limiter = SampleToInsertRatio(config.system.sample_per_insert, min_num_inserts, config.system.tolerance)
```

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+            episode_metrics, ep_completed = get_final_step_metrics(episode_metrics)
+            episode_metrics["steps_per_second"] = steps_per_rollout / time_metrics["rollout_time"]
+            if ep_completed:
+                logger.log(episode_metrics, t, eval_step, LogEvent.ACT)
+
+            train_metrics["learner_step"] = (eval_step + 1) * config.system.num_updates_per_eval
+            train_metrics["learner_steps_per_second"] = (
+                config.system.num_updates_per_eval
+            ) / time_metrics["learner_time_per_eval"]
+            logger.log(train_metrics, t, eval_step, LogEvent.TRAIN)
```
sash-a: I think we should always log train metrics even if an episode hasn't finished yet, what do you think?

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+        episode_return = np.mean(eval_metrics["episode_return"])
+
+        if save_checkpoint:  # Save a checkpoint of the learner state
+            checkpointer.save(
+                timestep=steps_per_rollout * (eval_step + 1),
+                unreplicated_learner_state=learner_state_cpu,
+                episode_return=episode_return,
+            )
+
+        if config.arch.absolute_metric and max_episode_return <= episode_return:
```
sash-a: I never know what order the bools will be evaluated in so I always add brackets, because it might be doing `(config.arch.absolute_metric and max_episode_return) <= episode_return`
```suggestion
        if config.arch.absolute_metric and (max_episode_return <= episode_return):
```

sash-a (changes requested):
Great work here! Really minor changes required. Happy to merge this pending some benchmarks

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,782 @@
...
+                    )
+                    cpu_action = jax.device_get(action)
+
+                # Step environment
+                with RecordTimeTo(actor_timings["env_step_time"]):
+                    next_timestep = env.step(cpu_action)
+
+                # Prepare the transation
+                terminal = (1 - timestep.discount[..., 0, jnp.newaxis]).astype(bool)
+                dones = next_timestep.last()[..., jnp.newaxis]
```
sash-a: Can we call this `next_done` just so it's clear which step it's aligned with

mava/utils/config.py
Louay-Ben-nessir: I split it into base_sebulba_checks and ppo_sebulba_checks. Any more splits feel excessive 🤔

mava/utils/sebulba.py
Louay-Ben-nessir: nope, just removed it and the related functions!

mava/systems/q_learning/sebulba/rec_iql.py
Louay-Ben-nessir: The dones flag is removed here and added back in the scannedRNN. We could either modify the RNN to handle the dones flag with or without the agent dimension or standardize it by keeping the agent dimension across all scripts. 🤔

mava/utils/sebulba/pipelines.py
```diff
@@ -0,0 +1,300 @@
...
+        """Clear the pipeline."""
+        while not self._queue.empty():
+            try:
+                self._queue.get(block=False)
+            except queue.Empty:
+                break
+
+    def stop(self) -> None:
+        """Signal the thread to stop."""
+        self._stop_event.set()
```
sash-a: This is a nice change, although I don't think it needs to be an event in this case, it can just be a bool. Because it's internal to this thread and there's no chance of a deadlock here since once you call stop nothing else will start the thread again :+1:

mava/wrappers/gym.py
```diff
@@ -251,18 +252,18 @@ def __init__(self, env: gymnasium.vector.VectorEnv):
...
         obs, info = self.env.reset(seed=seed, options=options)  # type: ignore
 
-        num_agents = len(self.env.single_action_space)  # type: ignore
         num_envs = self.env.num_envs
 
         step_type = np.full(num_envs, StepType.FIRST)
-        rewards = np.zeros((num_envs, num_agents), dtype=float)
-        teminated = np.zeros(num_envs, dtype=float)
+        rewards = np.zeros((num_envs, self.num_agents), dtype=float)
+        terminated = np.zeros((num_envs, self.num_agents), dtype=float)
```
SimonDuToit, line 264: I assume with this change we would also need to change sebulba PPO? Since currently it does this same operation. We should decide if generally its better doing this in the system or the wrapper.

mava/utils/sebulba/pipelines.py
```diff
@@ -0,0 +1,300 @@
...
+            except queue.Empty:
+                break
+
+    def qsize(self) -> int:
+        """Returns the number of trajectories in the pipeline."""
+        return self.metrics_queue.qsize()
+
+    def stop(self) -> None:
+        """Signal the thread to stop."""
+        self._stop_event.set()
```
SimonDuToit: Just to check: the Pipeline class is unchanged from the original one in utils/sebulba, right?

SimonDuToit (commented):
Great work Louay! Just two questions from my side as you can see in the comments.

mava/utils/sebulba/pipelines.py
```diff
@@ -0,0 +1,300 @@
...
+        with start_condition:
+            self.tickets_queue.put((start_condition, end_condition))
+            start_condition.wait()  # wait to be allowed to start
+
+        # [Transition(num_envs)] * rollout_len -> Transition[done=(num_envs, rollout_len, ...)]
+        traj = _stack_trajectory(traj)
+        traj, timestep = jax.device_put((traj, timestep), device=self.sharding)
+
+        time_dict, episode_metrics = metrics
+        # [{'metric1' : value1, ...} * rollout_len -> {'metric1' : [value1, value2, ...], ...}
```
sash-a, line 100: While we're here, this comment has been wrong for a while :joy: 
```suggestion
        # [{'metric1' : value1}, ...] * rollout_len -> {'metric1' : [value1, value2, ...], ...}
```

mava/utils/sebulba/pipelines.py
```diff
@@ -0,0 +1,300 @@
...
+            sample from the buffers.
+            init_transition : A sample trasition used to initialize the buffers.
+            lifetime: A `ThreadLifetime` which is used to stop this thread.
+        """
+        super().__init__(name="Pipeline")
+        self.cpu = jax.devices("cpu")[0]
+
+        self.tickets_queue: queue.Queue = queue.Queue()
+        self.metrics_queue: queue.Queue = queue.Queue(maxsize=100)
+        self._stop_event = threading.Event()
```
sash-a: Similar here, this can just be a bool

mava/utils/sebulba/pipelines.py
```diff
@@ -0,0 +1,300 @@
...
+            rate_limiter: A `RateLimiter` Used to manage how often we are allowed to
+            sample from the buffers.
+            init_transition : A sample trasition used to initialize the buffers.
+            lifetime: A `ThreadLifetime` which is used to stop this thread.
+        """
+        super().__init__(name="Pipeline")
+        self.cpu = jax.devices("cpu")[0]
+
+        self.tickets_queue: queue.Queue = queue.Queue()
+        self.metrics_queue: queue.Queue = queue.Queue(maxsize=100)
```
sash-a, line 180:
```suggestion
        # Only keep the latest 100 metrics, otherwise too many metrics are added we risk an OOM
        self.metrics_queue: queue.Queue = queue.Queue(maxsize=100)
```

mava/utils/sebulba/pipelines.py
```diff
@@ -0,0 +1,300 @@
...
+        traj = jax.device_get(traj)
+        # [Transition(num_envs)] * rollout_len -> Transition[done=(num_envs, rollout_len, ...)]
+        traj = _stack_trajectory(traj)
+
+        time_dict, episode_metrics = metrics
+        # [{'metric1' : value1, ...} * rollout_len -> {'metric1' : [value1, value2, ...], ...}
+        episode_metrics = _stack_trajectory(episode_metrics)
+
+        self.buffer_states[actor_id] = self.buffer_add(self.buffer_states[actor_id], traj)
+        self.buffer_adds_count[actor_id] += 1
```
sash-a: Do we still need this, seem it isn't used?

I think there's also a property of the buffer state we can use if we want to get the size

mava/configs/system/q_learning/rec_iql.yaml
```diff
@@ -31,3 +31,7 @@ gamma: 0.99  # discount factor
 
 eps_min: 0.05
 eps_decay: 1e5
+
+# --- Sebulba parameters ---
+replay_ratio : 4  # Average number of times the learner should sample each item from the replay buffer.
+error_tolerance:  ~ # Tolerance for how much the learner/actor can sample/insert before being blocked. Must be greater than 2 to avoid deadlocks.
```
sash-a: I think this is a slightly better explainer

```suggestion
error_tolerance: float  # Maximum allowed deviation from the target sampling ratio. Controls how far the actual samples can vary from the ideal samples-per-insert ratio. Must be > 2 to prevent deadlocks.
```

Also let's give this a value, because I think the blocking one might be too blocking so would be nice to make this the default. A value of ~4 seems good to me what did you find in your experiments?

mava/utils/sebulba/rate_limiters.py
sash-a, line 1: This is really great!

mava/utils/sebulba/utils.py
```diff
@@ -0,0 +1,76 @@
...
+    """A `ParamSource` is a component that allows networks params to be passed from a
+    `Learner` component to `Actor` components.
+    """
+
+    def __init__(self, init_value: Params, device: jax.Device):
+        super().__init__(name=f"ParamsSource-{device.id}")
+        self.value: Params = jax.device_put(init_value, device)
+        self.device = device
+        self.new_value: queue.Queue = queue.Queue()
+        self._stop_event = threading.Event()
```
sash-a: Same here for being a bool

mava/wrappers/gym.py
sash-a: Good point, I think we do this in the wrappers for the anakin systems

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,803 @@
...
+    Args:
+        key: The PRNG key for stochasticity.
+        env: The environment to interact with.
+        config: Configuration settings for rollout and environment.
+        rollout_queue: Queue for sending collected trajectories to the learner.
+        params_source: Provides the latest network parameters from the learner.
+        q_net: The Q-network.
+        actor_device: Index of the actor device to use for rollout.
+        seeds: Seeds for environment initialization.
+        thread_lifetime: Controls the thread's lifecycle.
```
sash-a:
```suggestion
        stop_event: used to inform this thread that it is time to stop.
```

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,803 @@
...
+from mava.utils.sebulba.utils import ParamsSource, RecordTimeTo
+from mava.wrappers.episode_metrics import get_final_step_metrics
+from mava.wrappers.gym import GymToJumanji
+
+
+def rollout(
+    key: chex.PRNGKey,
+    env: GymToJumanji,
+    config: DictConfig,
+    rollout_queue: Pipeline,
```
sash-a: In this case it's not a queue, because we're using a replay buffer. Can we rename this to pipeline

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,803 @@
...
+
+                # Update all the params sources so all actors can get the latest params
+                params = jax.block_until_ready(learner_state.params)
+                for source in params_sources:
+                    source.update(params.online)
+
+        # Pass all the metrics and  params to the main thread (evaluator) for logging and evaluation
+
+        if ep_metrics_list:
+            # [{metric1 : (num_envs, ...), ...} * n_rollouts] -->
```
sash-a:
```suggestion
            # [{metric1 : (num_envs, ...)}, ...] * n_rollouts -->
```

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,803 @@
...
+            # {metric1 : (n_rollouts, num_envs, ...), ...]
+            ep_metrics = tree.map(lambda *x: np.asarray(x), *ep_metrics_list)
+
+            # [{metric1: value1, ...} * n_rollouts] -->
```
sash-a:
```suggestion
            # [{metric1: value1}, ...] * n_rollouts -->
```

mava/systems/q_learning/sebulba/rec_iql.py
```diff
@@ -0,0 +1,803 @@
...
+        actor.join()
+        print(f"{Fore.RED}{Style.BRIGHT}{actor.name} stopped{Style.RESET_ALL}")
+    print(f"{Fore.RED}{Style.BRIGHT}Stopping pipeline...{Style.RESET_ALL}")
+    pipe.stop()
+    pipe.join()
+    print(f"{Fore.RED}{Style.BRIGHT}Stopping params sources...{Style.RESET_ALL}")
+    for params_source in params_sources:
+        params_source.stop()
+        params_source.join()
+    print(f"{Fore.RED}{Style.BRIGHT}All threads stopped...{Style.RESET_ALL}")
```
sash-a: Can we make this a function in sebulba/utils? And then use it in PPO also because it should be the same right?

sash-a (changes requested):
Looks great to me, just a few minor things :pray:

mava/utils/sebulba/pipelines.py
Louay-Ben-nessir: small change, removed the Threadlifetime. Everything else in the same

mava/utils/sebulba.py
Louay-Ben-nessir: I don't remember exactly 😅 
it has something to do with the states inside the buffer array possibly changing while we loop over them 🤔

mava/systems/q_learning/sebulba/rec_iql.py
Louay-Ben-nessir: I can easily add this in this PR 👀

mava/utils/sebulba/pipelines.py
Louay-Ben-nessir: Yep, just an artifact. Was used to track when we should call `self.rate_limiter.insert()` instead we insert fractions now `self.rate_limiter.insert(1 / self.num_buffers)`

mava/configs/system/q_learning/rec_iql.yaml
Louay-Ben-nessir: I think 4 is reasonable, Most of the top SMAC runs had 2 but some had 16

mava/systems/q_learning/sebulba/rec_iql.py
Louay-Ben-nessir: definitely, I've been meaning to also move this to before the  `absolute_metric` to avoid having the actor running in the background

sash-a (dismissed):
:fire:

mava/systems/q_learning/sebulba/rec_iql.py
sash-a: I think rather leave it for now, no need to make this more complex

mava/utils/sebulba/pipelines.py
```diff
@@ -0,0 +1,299 @@
...
+        self.tickets_queue: queue.Queue = queue.Queue()
+        self._queue: queue.Queue = queue.Queue(maxsize=max_size)
+        self._should_stop = False
+
+    def run(self) -> None:
+        """This function ensures that trajectories on the queue are consumed in the right order. The
+        start_condition and end_condition are used to ensure that only 1 thread is processing an
+        item from the queue at one time, ensuring predictable memory usage.
+        """
+        while not self._should_stop:
```
SimonDuToit, line 73: Please update the (online) pipeline with the changes recently merged into dev, which adds support for hstates.

SimonDuToit (changes requested):
Looks good! Just one request and one question.

mava/wrappers/gym.py
SimonDuToit: Also: if terminated is repeated for num_agents in the wrapper, should we still be repeating dones for num_agents in the systems? Because I see you didn't remove this from sebulba PPO. Or does this not actually effect the dones?

Louay-Ben-nessir: Yes, this doesn't affect the dones, it's only used for discount calculations. The step_type is still of shape (num_envs,)
I double checked and the TS chapes are the same in Anakin and Sebulba

SimonDuToit (approved):
LGTM

sash-a (approved):
:fire: :fire:

PR #1158, opened 2025-02-12
Use timestep positional encoding by default for Rec Sable

RuanJohn:
## Why?
It is important for Rec Sable to have timestep positional encodings, so I think it should be the default config setting.


OmaymaMahjoub (approved):
Thanks @RuanJohn 🙏

WiemKhlifi (approved):
Thanks @RuanJohn 🤓

PR #1160, opened 2025-02-27
feat: ff_ippo refactor

sash-a:
## What?
A nice refactor to reduce the nesting used in PPO and generally make the code more understandable.
## Why?
I wanted to test Claude code and am pleasantly surprised with the results.
## How?
I asked Claude code to refactor the code and reduce the nesting and it's done well
## Note:
This doesn't need to go in now, we should decide how quickly we want to merge this considering all the other PPO based systems.
Any style changes you've wanted should be added to this PR - so check carefully!


sash-a:
If we also pull out the `_update_step` function we could import that into the sebulba systems and use it as the learning function reducing some code duplication, but making the sebulba systems less single file. However I feel those files are getting so long that this might be a good option - it's kinda a question of where do we draw the line on importing logic :thinking:

SimonDuToit:
It seems to have deleted a bunch of the docstrings when it moved the functions around :sweat_smile:

Issue #1163, opened 2025-03-03
[BUG] Issue due to newest gymnasium==1.1.0


WiemKhlifi:
### Describe the bug
The latest working version of Mava with Gymnasium is `1.0.0,` while version `1.1.0` has errors.

### To Reproduce
1. Simply create new conda env and do all the installations.
2. Run `pip list | grep gymnasium`, the expected installed version will be 1.1.0.
3. Run PPO sebulba system and you'll get the 'expected behavior' below.
4. To fix, do `pip install gymnasium==1.0.0` and rerun PPO sebulba.

### Expected behavior
In the `make_env` function for Sebulba environments, we encountered an error from the `async_multiagent_worker`, as shown in the screenshot.
![Image](https://github.com/user-attachments/assets/a7d832ce-7f60-4462-8f67-b2f27cd8d82f)

### Possible Solution
Fix the bug given by the newer version or freeze the working version.

PR #1164, opened 2025-03-03
Fix: pin gymansium and flax versions

SimonDuToit:
Pin `gymnasium` version to `1.0.0` and restrict flax version to '<0.10.4'.
Both packages recently had updates which break mava. The gymnasium update breaks sebulba systems and the flax update breaks recurrent systems. The gymnasium issue is mentioned in #1163


sash-a (approved):

RuanJohn (approved):
Thanks for the fix 🙌

PR #1165, opened 2025-03-04
feat: rec sable sebulba

Louay-Ben-nessir:
What?
A recurrent sable implementation using the Sebulba architecture.


mava/utils/sebulba/pipelines.py
```diff
@@ -84,7 +84,7 @@ def put(
         self,
         traj: Sequence[MavaTransition],
         metrics: Tuple[Dict, List[Dict]],
-        final_timestep: Tuple[TimeStep, Optional[HiddenStates]],
+        final_timestep: Tuple[TimeStep, Any],
```
SimonDuToit, line 87: Whats the reason for this change? Doesn't rec sable also use HiddenStates?

SimonDuToit (commented):
Looks good, just one question from my side.

mava/utils/sebulba/pipelines.py
Louay-Ben-nessir: I am using the pipeline to transfer both the first hidden state used in the rollout and the last hidden state generated.

The pipeline should work with any PyTree structure that contains JAX arrays, but I couldn't find a clean way to annotate it

sash-a: I've been wanting to fix this for a while we should annotate this class with 3 generic types `B, T` B is the type for the batch and T is the type for the timestep. Then we can specify different types for ff/rec, however this would also need other code changes so I'm in favour of just leaving it as is for now and keeping it any and changing it in a future PR

mava/systems/sable/sebulba/rec_sable.py
```diff
@@ -0,0 +1,787 @@
...
+    ) -> Tuple[LearnerState, Metrics]:
+        """A single update of the network.
+
+        This function calculates advantages and targets based on the trajectories
+        from the actor and updates the actor and critic networks based on the losses.
+
+        Args:
+            learner_state (LearnerState): contains all the items needed for learning.
+            traj_batch (PPOTransition): the batch of data to learn with.
+        """
```
sash-a: This was quite confusing to me at first glance, can we change it to this (please also change where ever we use these names e.g: rollout and minibatching -  might be useful to change it in the anakin system also :pray:)

```suggestion
        initial_hstate: HiddenStates,
        final_hstate: HiddenStates,
    ) -> Tuple[LearnerState, Metrics]:
        """A single update of the network.

        This function calculates advantages and targets based on the trajectories
        from the actor and updates the actor and critic networks based on the losses.

        Args:
            learner_state (LearnerState): contains all the items needed for learning.
            traj_batch (PPOTransition): the batch of data to learn with.
            initial_hstate (HiddenState): the hidden state from the start of the rollout.
            final_hstate (HiddenState): the last hidden state of the rollout.
        """
```

mava/systems/sable/sebulba/rec_sable.py
```diff
@@ -0,0 +1,787 @@
...
+        timing_dict = rollout_times | learn_times
+        timing_dict = tree.map(np.mean, timing_dict, is_leaf=lambda x: isinstance(x, list))
+
+        eval_queue.put((ep_metrics, train_metrics, learner_state, timing_dict))
+
+
+def learner_setup(
+    key: chex.PRNGKey, config: DictConfig, learner_devices: List
+) -> Tuple[
+    Callable[[LearnerState, Transition, HiddenStates, HiddenStates], Tuple[LearnerState, Metrics]],
```
sash-a: This is annoying but not sure there is a way around it, seeing as you use it above can you store it in a variable please and use that instead

mava/systems/sable/sebulba/rec_sable.py
```diff
@@ -0,0 +1,787 @@
...
+    Callable,
```
sash-a, line 474: I think this is the same callable as above

mava/systems/sable/sebulba/rec_sable.py
```diff
@@ -0,0 +1,787 @@
...
+        shard_map(
+            learn,
+            mesh=mesh,
+            in_specs=(learn_state_spec, data_spec, data_spec, data_spec),
+            out_specs=(learn_state_spec, data_spec),
+        )
+    )
+
+    # Load model from checkpoint if specified.
+    # TODO: check if this is working
```
sash-a: Did you check?

mava/systems/sable/sebulba/rec_sable.py
```diff
@@ -0,0 +1,787 @@
...
+
+        Args:
+            learner_state (LearnerState): contains all the items needed for learning.
+            traj_batch (PPOTransition): the batch of data to learn with.
+        """
+
+        # Calculate advantage
+        params, opt_states, key, _, final_timestep = learner_state
+        key = jnp.squeeze(key, axis=0)
+        _, _, current_val, _ = sable_action_select_fn(  # type: ignore
```
sash-a: Instead of current_val/current_done can we change to final_val/final_done

mava/systems/sable/sebulba/rec_sable.py
```diff
@@ -0,0 +1,787 @@
...
+                # Step environment
+                with RecordTimeTo(actor_timings["env_step_time"]):
+                    timestep = env.step(cpu_action)
+
+                # Updated the dones and Hstates
+                dones = timestep.last()
+                dones = jnp.expand_dims(dones, (1, 2, 3, 4))
+                hstates = tree.map(
+                    lambda hs, dones=dones: jnp.where(dones, jnp.zeros_like(hs), hs), hstates
+                )
```
sash-a: np instead of jnp

```suggestion
                dones = np.expand_dims(dones, (1, 2, 3, 4))
                hstates = tree.map(
                    lambda hs, dones=dones: np.where(dones, np.zeros_like(hs), hs), hstates
                )
```

mava/systems/sable/sebulba/rec_sable.py
```diff
@@ -0,0 +1,787 @@
...
+        prev_hstates = hstates
+
+        with RecordTimeTo(actor_timings["rollout_time"]):
+            for _ in range(config.system.rollout_length):
+                with RecordTimeTo(actor_timings["get_params_time"]):
+                    params = params_source.get()  # Get the latest parameters from the learner
+
+                obs_tpu = tree.map(move_to_device, timestep.observation)
+
+                prev_dones = jnp.repeat(timestep.last(), config.system.num_agents).reshape(
```
sash-a: np instead of jnp
```suggestion
                prev_dones = np.repeat(timestep.last(), config.system.num_agents).reshape(
```

sash-a (changes requested):
Looks good just a few small changes

mava/systems/sable/sebulba/rec_sable.py
Louay-Ben-nessir: I've changed the names to  last_val/last_done as that's more in line with the GAE function, happy to make it final_val/final_done if you think that's better

sash-a (approved):
:partying_face: :balloon:

SimonDuToit (approved):
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣤⣤⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠸⣿⣿⣿⣷⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣿⣧⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣾⣿⣿⣿⣿⡟⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⣰⣿⣿⣿⣿⣿⣿⣷⣶⣶⣶⣤⣤⣀⡀⠀⠀
⠀⣀⣀⣀⣀⡀⠀⢀⣾⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡆⠀
⢰⣿⣿⣿⣿⡇⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣅⠀
⢸⣿⣿⣿⣿⡇⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠄
⣿⣿⣿⣿⣿⡇⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠋⠀
⣿⣿⣿⣿⣿⡇⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠀⠀
⢸⣿⣿⣿⣿⡇⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡟⠋⠀⠀
⢸⣿⣿⣿⣿⡇⠀⠀⠙⠿⠿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡿⠿⠃⠀⠀⠀
⠀⠛⠛⠛⠛⠃⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀

PR #1166, opened 2025-03-13
feat: logger refactor

sash-a:
## What?
Updated the logging to be a bit cleaner and much more extendable
* Now can log any custom metrics added to `extras["env_metrics"]` at both eval and act time (not just win rate at eval time)
* Instantiation of loggers has been cleaned up as well as the config
* Added a `log_config` method to pretty print the config (for console logger) and put it on neptune (for neptune logger)
* I think that everything :shrug: 
## Why?
I have some plans for improving our sweeping workflow and it involves making the loggers more extendable and also I wanted to test more AI coding tools, a lot of the code here is courtesy of [aider](https://aider.chat/)

### DIFF
--- a/mava/configs/logger/logger.yaml
+++ b/mava/configs/logger/logger.yaml
@@ -8,7 +8,7 @@ loggers:
   neptune:
     _target_: mava.utils.logger.NeptuneLogger
     enabled: True
-    
+
     # If specified will resume the run with this run ID.
     # This is useful for resuming runs and logging from multiple processes.
     # NOTE: it will overwrite the run unless you set the timestep correctly.
--- a/mava/configs/logger/logger.yaml
+++ b/mava/configs/logger/logger.yaml
@@ -1,22 +1,36 @@
 # --- Logging options ---
+base_exp_path: results/
 
-base_exp_path: results # Base path for logging.
-use_console: True # Whether to log to stdout.
-use_tb: False # Whether to use tensorboard logging.
-use_json: False # Whether to log marl-eval style to json files.
-use_neptune: False  # Whether to log to neptune.ai.
+loggers:
+  console:
+    _target_: mava.utils.logger.ConsoleLogger
+    enabled: true
+  neptune:
+    _target_: mava.utils.logger.NeptuneLogger
+    enabled: true
 
-# --- Other logger kwargs ---
-kwargs:
-  neptune_project: Instadeep/Mava
-  neptune_tag: [rware]
-  detailed_neptune_logging: False  # having mean/std/min/max can clutter neptune so we make it optional
-  json_path: ~ # If set, json files will be logged to a set path so that multiple experiments can
-    # write to the same json file for easy downstream aggregation and plotting with marl-eval.
-  upload_json_data: False # Whether JSON file data should be uploaded to Neptune for downstream
+    project: Instadeep/Mava
+    tag: [delete]
+    detailed_logging: False  # having mean/std/min/max can clutter neptune so we make it optional
+    architecture_name: ${arch.architecture_name}
+    # Whether JSON file data should be uploaded to Neptune for downstream
     # aggregation and plotting of data from multiple experiments. Note that when uploading JSON files,
     # `json_path` must be unset to ensure that uploaded json files don't continue getting larger
-    # over time. Setting both will break.
+    # over time. Setting both will raise an error.
+    upload_json_data: False  
+  json:
+    _target_: mava.utils.logger.JsonLogger
+    enabled: true
+    
+    # If set, json files will be logged to a set path so that multiple experiments can
+    # write to the same json file for easy downstream aggregation and plotting with marl-eval.
+    path: ~ 
+    task_name: ${env.scenario.task_name}
+    env_name: ${env.env_name}
+    seed: ${system.seed}
+  tensorboard:
+    _target_: mava.utils.logger.TensorboardLogger
+    enabled: true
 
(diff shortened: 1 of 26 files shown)


mava/configs/logger/logger.yaml
```diff
@@ -1,23 +1,41 @@
...
-use_tb: False # Whether to use tensorboard logging.
-use_json: False # Whether to log marl-eval style to json files.
-use_neptune: False  # Whether to log to neptune.ai.
+loggers:
+  console:
+    _target_: mava.utils.logger.ConsoleLogger
+    enabled: True
+  neptune:
+    _target_: mava.utils.logger.NeptuneLogger
+    enabled: True
```
OmaymaMahjoub, line 15: Can you set the default to false

mava/utils/logger.py
```diff
@@ -41,19 +45,73 @@ class LogEvent(Enum):
     MISC = "misc"
 
 
-class MavaLogger:
-    """The main logger for Mava systems.
+def winrate_custom_metric(metrics: Dict, config: DictConfig) -> Dict:
```
OmaymaMahjoub: we can remove config here it is not used

mava/configs/logger/logger.yaml
```diff
@@ -1,23 +1,41 @@
...
+    # If specified will resume the run with this run ID.
+    # This is useful for resuming runs and logging from multiple processes.
+    # NOTE: it will overwrite the run unless you set the timestep correctly.
+    run_id: ~
+    project: Instadeep/mava-benchmark
+    tag: [delete]
+    group_tag: [delete]
+    detailed_logging: False  # having mean/std/min/max can clutter neptune so we make it optional
+    architecture_name: ${arch.architecture_name}
+    # Whether JSON file data should be uploaded to Neptune for downstream
```
RuanJohn: I really like that the arguments for each logger are now directly associated with it. 
Just a note, can you please add a comment saying that we need the arch name for the neptune logger?

RuanJohn: Quite a nit point as well, but I find it easier if multi-line comments start on the same line as the config setting they are describing.

mava/utils/logger.py
```diff
@@ -41,19 +45,73 @@ class LogEvent(Enum):
...
+
+    return metrics
+
 
-    def __init__(self, config: DictConfig) -> None:
+class MavaLogger:
+    def __init__(
+        self,
+        config: DictConfig,
+        custom_metrics_fn: Callable[[Dict, DictConfig], Dict] = winrate_custom_metric,
```
RuanJohn: Doesn't this put us in a bit of tricky situation moving forward if environments have many custom metrics since this looks like it assumes we will only have one custom metric at a time?

Or if the custom metric need to extra processing like win rate does, it should just log those with no problem?

mava/configs/logger/logger.yaml
```diff
@@ -1,23 +1,41 @@
 # --- Logging options ---
```
RuanJohn, line 1: A comment at the top of this config that explains how things work would be really nice.

mava/utils/logger.py
```diff
@@ -148,29 +199,51 @@ def stop(self) -> None:
...
+            run_id: ID of the run you wish to resume - None if you don't want to resume the run.
+                Note this will overwrite the run if you restart the step from 0.
+        """
+        # async logging leads to deadlocks in sebulba
+        mode = "async" if architecture_name == "anakin" else "sync"
 
-        self.logger["config"] = stringify_unsupported(cfg)
-        self.logger["sys/group_tags"].add(list(cfg.logger.kwargs.neptune_group_tag))
+        if run_id is not None:
+            self.logger = neptune.init_run(with_id=run_id, project=project, mode=mode)
```
RuanJohn, line 246: Do we ever use this?

RuanJohn (changes requested):
Thanks @sash-a! Overall, this is a super nice update! 
Just a few comments and questions.

mava/utils/logger.py
sash-a: So if you had more custom metrics that need processing you just create a function that handles all of them, the function just takes all the metrics so it would modify the ones that it cares about. 

But that's only for custom metrics that need to be modified many metrics are probably fine just as is

sash-a: It's sort of only a win, we get the option to log these metrics where as before it was tricky and you had to modify the logger itself and was a bit hacky to pass metrics all the way through to the logger. So if envs have custom metrics and we don't want to support them then we just don't add them to `extras["env_metrics"]` to begin with, otherwise we add them and process them if need be.

mava/configs/logger/logger.yaml
sash-a: Done for the arch, I gotta disagree with you on the comments though :joy:. Not too fussed though if you feel strongly about it I'll change it

mava/utils/logger.py
RuanJohn: Perfect, thanks for clarifying. Can you maybe just add a comment in the logger that briefly explains this?

mava/utils/logger.py
RuanJohn: Discussed offline.

OmaymaMahjoub (dismissed):
Thanks @sash-a 🔥 
I left two minor comments :)

OmaymaMahjoub (approved):
LGTM! 🔥

RuanJohn (approved):
🚀

PR #1167, opened 2025-03-28
fix: small rec_iql fixes

Louay-Ben-nessir:
## What?
update the default rec_iql default configs and init the obsteration step_count correctly


sash-a (approved):
:pray:

SimonDuToit (approved):

PR #1168, opened 2025-04-02
feat: add neptune group tags

RuanJohn:
## What?
Super small quality of life improvement that adds Neptune group tags. This makes it very easy to use the tags for a specific experiment that belongs to a group of experiments.


sash-a (approved):
:wood:

WiemKhlifi (approved):
🌟

PR #1171, opened 2025-04-07
feat: uv

### LINKED ISSUE #1092: [MAINTAIN] Use uv for managing dependencies.
Use [uv](https://github.com/astral-sh/uv) for dependency management and use it to produce a locked requirements file with which Mava is guaranteed to run. 

### Why? 
[uv](https://github.com/astral-sh/uv) can produce great compiled and locked requirements files. It is also a lot faster than using `pip` and is a drop in replacement for it. 

## How?
Make a locked requirements file and updated the README to suggest [uv](https://github.com/astral-sh/uv). The locked requirements file will still work with normal `pip`, but [uv](https://github.com/astral-sh/uv) will just be a lot faster.

sash-a:
## What?
Fully switch to uv.
## Why?
It's fast and easy to use :shrug: 
## How?
Remove requirements files
Moved all requirements to pyproject.toml
Changed instructions in readme

## Note
No functionality is lost, we can still use pip/conda if we want to (`pip install -e .`), this just makes uv work better and supports it as the package manager of choice

Closes: #1092


SimonDuToit:
Not super sure how uv works but should the uv.lock file be in the repo? Or can it just be generated locally?

README.md
```diff
@@ -41,11 +41,18 @@ Mava allows researchers to experiment with multi-agent reinforcement learning (M
...
+At the moment Mava is not meant to be installed as a library, but rather to be used as a research tool. We recommend cloning the Mava repo and installing dependencies using [uv](https://github.com/astral-sh/uv) as follows:
 
 ```bash
-git clone https://github.com/instadeepai/mava.git
-cd mava
+# Clone the repository
+git clone https://github.com/instadeepai/Mava.git
+cd Mava
+# Create a virtual environment and install all dependencies
+uv sync
```
RuanJohn, line 51: I think we need to and an extra step here where the user either pins their global `uv` python version or pins the python version used in the Mava directory. This is for incase someone's base python version is 3.13 in which case the installation won't run. 

Another detail is that we need to add `setuptools` to the requirements and `uv.lock` or explicitly install it after running `uv sync` otherwise the code won't run. 

Might be overkill but adding a line activating the venv here could be nice as well.

RuanJohn (changes requested):
Thanks for doing this @sash-a! I really like this and that it is so fast ⚡ 
Just a general comment around adding `setuptools` so that the code will run in the uv venv first try. 

Could you also please check the detailed install instructions and make sure everything is still good?

README.md
sash-a: Good call on the setuptools stuff will do.
So uv will read our pyproject and take the latest python version that is compatible, we should actually limit our supported version to 3.12 and then sync will always create an env with the correct version

sash-a:
> Not super sure how uv works but should the uv.lock file be in the repo? Or can it just be generated locally?

Good question, I *think* it should, it's basically a frozen version of requirements with everything pinned so that we know it works 100% of the time.

README.md
sash-a: @RuanJohn this should be done :smile:

RuanJohn: I had a weird situation where I only had python 3.13 locally so then `uv sync` wouldn't work. But I don't think we have to explicitly mention this in the readme. Perhaps one day an error like that could live something like a FAQ.

sash-a: What we could do for the default install instructions in make it `uv sync -p 3.12`?

sash-a:
Updated the detailed install, not sure if we should mention `uv run ...` in the main readme?

RuanJohn:
> Updated the detailed install, not sure if we should mention `uv run ...` in the main readme?

Thanks! I think we can leave `uv run` in the detailed install. The readme looks great to me now and everything works from my side.

RuanJohn (dismissed):
🔥

SimonDuToit (dismissed):
LGTM

OmaymaMahjoub (approved):
Thanks @sash-a 🔥

RuanJohn (approved):
🥇

PR #1174, opened 2025-04-14
fix: matrax env metrics when using global state

sash-a:
## What?
Fixes minor bug in the logging PR merged on Friday, which the integration tests found. Basically matrax didn't add `env_metrics` to extras if using a global state, it only added it in IL systems


RuanJohn (approved):
Thanks for the fix 🙌

WiemKhlifi (approved):
Nice catch 🙌

PR #1175, opened 2025-04-22
fix: remove path attribute in checkpoint config

WiemKhlifi:
## What?
The `Checkpointer ` class doesn't accept any argument with the name `path`.


sash-a (changes requested):
Instead of removing it can we change the `rel_dir` arg in the checkpointer to `path`

WiemKhlifi:
Yeah that's a good idea, and we set the default "`checkpoints`" in config 👌 
People just need to take into consideration that the new `rel_dir` isn't ignored in `.gitignore`

sash-a (approved):

OmaymaMahjoub (approved):
LGTM! thanks @WiemKhlifi 🙏

PR #1180, opened 2025-05-23
fix: neptune upload json files

sash-a:
## What?
Neptune json file uploading was broken. The json logger was making a folder `..../metrics.json/metrics.json` instead of just making the metrics file and so the neptune logger was pointing to the wrong place


WiemKhlifi (approved):
LGTM thanks for the fix 🌟

ArnolFokam (approved):
LGTM! 🔥

PR #1182, opened 2025-06-18
chore: revert to jumanji pip package

sash-a:
Jumanji is now released so we can use the pip package


SimonDuToit (approved):
whoop

Louay-Ben-nessir (approved):
🙆‍♂️

PR #1183, opened 2025-07-14
Add base env_metrics to extras for Sebulba systems

RuanJohn:
## What?
Add `env_metrics` as an empty dictionary to the timestep extras. 
## Why?
Otherwise when we call `metrics = timestep.extras["episode_metrics"] | timestep.extras["env_metrics"]` in the Sebulba system rollouts it will fail for environments that don't have custom env metrics. 
## Extra
Some minor updates to the contributing docs.


sash-a (approved):
:bug:

WiemKhlifi (approved):
🐛 ²²²²

PR #1191, opened 2025-09-29
Maintaining the quickstart notebook

### LINKED ISSUE #1185: [BUG] Quickstart notebook broken
### Describe the bug
The Quickstart notebook is currently broken when it tries to initialize the environment. It looks like this is because the environment initialization was changed in a recent PR to check for GNN-related stuff.

### To Reproduce
1. Run the Quickstart notebook, it will crash in the cell that creates the training and evaluation environments.

### Expected behavior
The environment should be initialized without crashing.

### Context (Environment)
 - Running the latest version of Mava
 - OS is WSL Linux
 - Running on CUDA Jax
 
### Additional context
No additional context.

### Possible Solution
Unsure, but this would theoretically break any custom neural networks that someone wants to do that don't use the existing Mava config framework, right?

OmaymaMahjoub:
## What?
- Fix the quickstart notebook to make it compatible with the latest changes in Mava.

## How?
- Updated types and configuration to match the recent changes.

## Extra
- close #1185


sash-a (approved):
Thanks @OmaymaMahjoub 🔥

WiemKhlifi (approved):
LGTM and works fine ✅

PR #1195, opened 2025-10-24
Update README.md

scott-vsi:
fix path in command


WiemKhlifi (approved):
Thanks @scott-vsi 😊

sash-a (approved):
Thanks!

PR #1209, opened 2026-08-22
feat: add batch dimension utility

### LINKED ISSUE #1153: [FEATURE] Create an `add_batch_dim` method
All over Mava we have statements like `tree.map(lambda x: x[jnp.newaxis], pytree)` I think it would make it a lot clearer if we had an `add_batch_dim` method in `utils.jax_utils`

quinnarnold:
## What?

Add an `add_batch_dim` pytree utility and use it in place of repeated `tree.map(lambda x: x[jnp.newaxis, ...], pytree)` expressions across Mava.

## Why?

The named helper makes the intent of these operations clearer and resolves #1153 without changing batching behavior.

## How?

- Add `add_batch_dim` to `mava/utils/jax_utils.py`.
- Replace 29 equivalent one-leading-axis tree maps across evaluator and system implementations.
- Leave operations that add multiple axes, insert a non-leading axis, repeat values, or slice before batching unchanged.

## Testing

- `uv run pytest -q test/integration_test.py -k "ppo_system or sable_system or q_learning_system or transformer_system" -p no:warnings` — 9 passed.
- Ruff, Ruff format, file-safety, license, and diff checks pass.
- `uv run mypy mava/utils/jax_utils.py` passes.

The repository-wide mypy hook currently reports an existing error at `mava/systems/gpo/anakin/rec_magpo.py:664`; this PR does not modify that line or the affected hidden-state types.

Closes #1153

### DIFF
--- a/mava/evaluator.py
+++ b/mava/evaluator.py
@@ -37,6 +37,7 @@ from mava.types import (
     RecActorApply,
     State,
 )
+from mava.utils.jax_utils import add_batch_dim
 from mava.wrappers.gym import GymToJumanji
 
 # Optional extras that are passed out of the actor and then into the actor in the next step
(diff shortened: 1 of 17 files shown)


sash-a (approved):
Much cleaner and clearer, thanks!

PR #1211, opened 2026-08-22
feat: use shared GAE in MAT and Sable

### LINKED ISSUE #1139: [FEATURE] Make MAT and Sable use the generic GAE function.
MAT and Sable still have their own GAE functions. They should be changed to use the generic GAE function shared by the PPO systems.

quinnarnold:
## What?
Use the generic GAE utility in the MAT and Sable Anakin systems.

Closes #1139.

## Why?
MAT and Sable still maintained local GAE implementations after the PPO systems were migrated to the shared utility. Using the common helper removes duplicated logic and keeps advantage estimation consistent across systems.

## How?
- Replace the local MAT and feed-forward/recurrent Sable GAE functions with `calculate_gae`.
- Store the previous timestep termination flag in MAT and feed-forward Sable transitions, as expected by the shared reverse scan.
- Pass the final timestep termination flag into the shared helper.

## Extra
Validation performed:
- `uv run pytest -q test/integration_test.py -k "transformer_system or sable_system" -p no:warnings` (3 passed)
- Targeted mypy for all three changed files (passed)
- Project-pinned Ruff and Ruff formatter hooks for all three changed files (passed)
- Numerical episode-boundary comparison between the previous local calculation and shared helper (passed)

The repository-wide mypy hook still reports the existing unrelated error in `mava/systems/gpo/anakin/rec_magpo.py:659`.

### DIFF
--- a/mava/systems/mat/anakin/mat.py
+++ b/mava/systems/mat/anakin/mat.py
@@ -44,6 +44,7 @@ from mava.utils.checkpointing import Checkpointer
 from mava.utils.config import check_total_timesteps
 from mava.utils.jax_utils import merge_leading_dims, unreplicate_batch_dim, unreplicate_n_dims
 from mava.utils.logger import LogEvent, MavaLogger
+from mava.utils.multistep import calculate_gae
 from mava.utils.network_utils import get_action_head
 from mava.utils.training import make_learning_rate
 from mava.wrappers.episode_metrics import get_final_step_metrics
@@ -95,9 +96,11 @@ def get_learner_fn(
             # Step environment
             env_state, timestep = jax.vmap(env.step, in_axes=(0, 0))(env_state, action)
 
-            done = timestep.last().repeat(env.num_agents).reshape(config.arch.num_envs, -1)
+            prev_done = (
+                last_timestep.last().repeat(env.num_agents).reshape(config.arch.num_envs, -1)
+            )
             transition = PPOTransition(
-                done, action, value, timestep.reward, log_prob, last_timestep.observation
+                prev_done, action, value, timestep.reward, log_prob, last_timestep.observation
             )
             learner_state = LearnerState(params, opt_state, key, env_state, timestep)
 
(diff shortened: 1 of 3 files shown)


sash-a:
This looks good, but I need to be sure this doesn't have an impact on performance. Can you please do 2 things:
1. Check that the output of the two functions are identical in a synthetic tests
2. Run a benchmark on rware small-4ag on develop and on this branch to make sure there is no regression and post a screenshot of the graph here

quinnarnold:
@sash-a I completed both requested checks:

- **Synthetic equivalence:** 100 randomized trajectories produced identical outputs for the previous MAT, feed-forward Sable, and recurrent Sable calculations versus the shared `calculate_gae` helper.
- **`rware` `small-4ag` benchmark:** develop (`8a1e0d89`) and this branch (`9eec495a`) used the same seed-42 feed-forward Sable configuration for 3,997,696 environment steps / 122 evaluations. The return arrays were exactly identical. Mean SPS after compilation was 15,270.81 on develop and 16,439.59 on this branch (+7.65%); wall time was 767.10s vs 693.83s. Final-10 mean return was 0.003125 and best return was 0.0625 on both.

mava-1211-rware-small-4ag-benchmark

RuanJohn (commented):
Hi @quinnarnold, thank you for the work on this! Along with @sash-a's request. For the benchmark, could you please do the benchmark on develop and your new branch using the same setup as we used in the [Sable paper](https://arxiv.org/pdf/2410.01706)? For this, can you please also use the same hyperparameters we used in the paper? You should be able to download get them from the Experiment Data tab [here](https://sites.google.com/view/sable-marl).

quinnarnold:
@RuanJohn Thanks — I reran the comparison on current `develop` (`c6893be5`) and this PR branch (`9eec495a`) using the Sable Experiment Data hyperparameters for `rware small-4ag` (64 environments, 4 minibatches, max grad norm 10, embedding size 128, 5 PPO epochs, clip epsilon 0.05, 4 heads, entropy coefficient 0.01, 1 block, actor LR 0.001), with seed 0 on both.

For a practical matched local CPU regression run, I retained the paper model, optimizer, environment, and 10-update evaluation cadence but scaled the duration from 1,220 updates to 120 updates: 12 evaluations and 1,966,080 environment steps. All 12 mean-return points were exactly identical, including the final return of 2.25. Mean throughput after the compilation point was 7,557.40 SPS on develop and 7,562.93 SPS on the PR branch (+0.073%), so I did not observe a behavioral or steady-state throughput regression.

develop-vs-pr-120-updates

sash-a (approved):
Had a look at the graphs in the paper and seems correct also check the output and it is the same

PR #1212, opened 2026-08-22
fix: report Anakin learner throughput separately

### LINKED ISSUE #1078: [BUG] steps_per_second under-reporting in SAC and IQL
### Describe the bug
The (originally combined) `steps_per_second` logging has been separated for now to enforce logging consistencies across Mava systems, enabling downstream aggregate plotting via `marl-eval`. However, the current `steps_per_second` logging in the IQL and SAC systems is not a true representation for the learner or the actor. It is not entirely fair to say that we do `steps_per_rollout / elapsed _time` steps per second in both cases because we are also taking `anakin_steps * cfg.system.epochs` gradient steps.

quinnarnold:
## What?
Report learner gradient-update throughput separately from Anakin environment throughput for recurrent IQL and the ISAC, MASAC, and HASAC systems.

Closes #1078.

## Why?
`steps_per_second` represents environment interactions. Learner updates are useful throughput information too, but combining the two counts gives the existing metric two different units.

## How?
Keep `steps_per_second` unchanged and add `learner_updates_per_second`, calculated as the learner updates performed between evaluations divided by elapsed update time.

## Extra
Validation:
- `python -m pytest -q test/integration_test.py::test_q_learning_system test/integration_test.py::test_sac_system` (5 passed)
- scoped Ruff lint and formatting checks
- scoped mypy checks for all four changed files
- `git diff --check`

The full-project mypy hook also reports an existing unrelated type error in `mava/systems/gpo/anakin/rec_magpo.py:659`.

### DIFF
--- a/mava/systems/q_learning/anakin/rec_iql.py
+++ b/mava/systems/q_learning/anakin/rec_iql.py
@@ -564,10 +564,10 @@ def run_experiment(cfg: DictConfig) -> float:
             cfg.system.eps_min, 1 - (t / cfg.system.eps_decay) * (1 - cfg.system.eps_min)
         )
         final_metrics, ep_completed = episode_metrics.get_final_step_metrics(metrics)
-        final_metrics["steps_per_second"] = (
-            steps_per_rollout + learn_steps_per_rollout
-        ) / elapsed_time
-        loss_metrics = losses
+        final_metrics["steps_per_second"] = steps_per_rollout / elapsed_time
+        loss_metrics = losses | {
+            "learner_updates_per_second": learn_steps_per_rollout / elapsed_time
+        }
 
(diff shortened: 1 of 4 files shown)


sash-a:
We can't combine these into 1 metric, it doesn't really make sense to do so. What we could add is `learner_updates_per_second` as a metric which would be `learn_steps_per_rollout / elapsed_time`. Could you add this instead?

quinnarnold:
@sash-a Updated as requested in 7ae87e95: `steps_per_second` remains environment-only, and the four affected systems now report `learner_updates_per_second = learn_steps_per_rollout / elapsed_time` separately. The targeted IQL/SAC integration tests pass (5 passed), along with scoped Ruff and mypy checks.

sash-a (approved):
Looks good! 🔥

PR #1213, opened 2026-08-24
fix: support JAX 0.6 manual axes in Sebulba

### LINKED ISSUE #1184: [BUG] `jax==0.6.0` causes type mismatches between `jax.lax.scan` outputs and initial carry values
### Describe the bug
Recent version of JAX (`jax==0.6.0` in this case) breaks the `jax.lax.scan` in the GAE and system loss calculations. The reason for this is that initial carry values are not explicitly put on the learner device leading to type mismatches. 

### To Reproduce
1. `uv sync --extra cuda12`
2. `python mava/systems/ppo/sebulba/ff_ippo.py env=lbf_gym`

### Expected behavior
The system will fail with an error message along these lines:
```
  File "Mava/mava/systems/ppo/sebulba/ff_ippo.py", line 400, in learner_thread
    learner_state, train_metrics = learn_fn(learner_state, traj_batch)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Mava/mava/systems/ppo/sebulba/ff_ippo.py", line 366, in learner_fn
    learner_state, loss_info = _update_step(learner_state, traj_batch)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Mava/mava/systems/ppo/sebulba/ff_ippo.py", line 196, in _update_step
    advantages, targets = calculate_gae(
                          ^^^^^^^^^^^^^^
  File "Mava/mava/utils/multistep.py", line 61, in calculate_gae
    _, advantages = jax.lax.scan(
                    ^^^^^^^^^^^^^
TypeError: scan body function carry input and carry output must have equal types, but they differ:

The input carry component carry[0] has type float32[32,2] but the corresponding output carry component has type float32[32,2]{learner_devices}, so the varying manual axes do not match.

This might be fixed by applying `jax.lax.pvary(..., ('learner_devices',))` to the initial carry value corresponding to the input carry component carry[0].

Revise the function so that all output types match the corresponding input types.
```

### Additional context
A potential solution is discussed [here](https://docs.jax.dev/en/latest/notebooks/shard_map.html#scan-vma) we just need to implement it. 

### Possible Solution
Implement the solution discussed above. In the mean time, we can just pin to `jax==0.5.2`.

quinnarnold:
## What?

Fixes #1184 by disabling `shard_map` replication checks for the affected Sebulba learner steps.

## Why?

JAX 0.6 tracks manual-axis variance through `shard_map`. Its default representation check rejects the valid scan carry in these learner steps because the input and output variance annotations differ.

## How?

Set `check_rep=False` on the seven affected Sebulba learner `shard_map` calls. This keeps the learner logic, GAE implementation, losses, CI configuration, and test suite unchanged.

## Extra

Validation:
- reproduced the issue's exact manual-axis scan-carry mismatch on the parent commit with JAX 0.6 and two learner devices
- the same two-device feed-forward IPPO/LBF smoke run completes on this branch
- scoped Ruff lint and formatting checks
- scoped mypy checks for all seven changed source files
- `git diff --check`


.github/workflows/ci.yaml
```diff
@@ -35,3 +35,12 @@ jobs:
 
       - name: Run linters 🖌️
         run: pre-commit run --all-files --verbose
+
+      - name: Run JAX 0.6 shard-map regression test
```
sash-a: This shouldn't go in our CI

test/multistep_test.py
sash-a, line 1: I don't think we need this test, could you please remove it

mava/utils/multistep.py
```diff
@@ -58,9 +60,14 @@ def _get_advantages(
         gae = delta + gamma * gae_lambda * (1 - next_done) * gae
         return (gae, value, done), gae
 
+    initial_gae = jnp.zeros_like(last_val)
+    pvary = getattr(jax.lax, "pvary", None)
+    if axis_name is not None and pvary is not None:
+        initial_gae = pvary(initial_gae, (axis_name,))
+
```
sash-a: A simpler version is 
```suggestion
    initial_gae = last_val * 0.0

```

sash-a (changes requested):
An easier fix to all of this is just to set `check_rep=False` in the shard_map call, that should fix it and not require any extra (would remove the inner `pmean`s and the gae axis_name argument.

Please do double check me that it fixes the issue

quinnarnold:
@sash-a Updated in 1b0f75be following all review notes: the CI and regression-test additions are removed, as are the GAE axis annotation and inner `pmean` changes. The effective fix is now only `check_rep=False` on the seven affected Sebulba learner `shard_map` calls. I double-checked the behavior by reproducing the exact manual-axis carry mismatch on the parent commit under JAX 0.6 with two learner devices; the same FF-IPPO/LBF smoke run completes on this branch. Scoped Ruff and mypy checks also pass.

sash-a (approved):
Thanks

PR #1216, opened 2026-08-24
feat: support Jumanji MultiCVRP

### LINKED ISSUE #1138: [FEATURE] Support Jumanji's MultiCVRP
### Feature
Adding a wrapper for jumanji's [MultiCVRP](https://github.com/instadeepai/jumanji/tree/main/jumanji/environments/routing/multi_cvrp). The motivation for this, is to support more environments that closely align with real-world applications.

### Proposal
A good starting point would be to update the existing non maintained [PR](https://github.com/instadeepai/Mava/pull/1043).

### Testing
Integrate the environment into the list of envs to test in the [integration_test.py](https://github.com/instadeepai/Mava/blob/develop/test/integration_test.py) and make sure it works fine.

### Benchmarking
Benchmark the performance of the existing algorithms on MultiCVRP.  
[Optional] Create a README file to document the benchmark results.

quinnarnold:
## What?
Add Mava support for Jumanji's MultiCVRP environment.

Closes #1138.

## Why?
MultiCVRP provides a cooperative routing task that is closely aligned with real-world multi-agent applications, but its observations and scalar rewards are not directly compatible with Mava's shared multi-agent interface.

## How?
- Add a `MultiCVRPWrapper` that exposes per-vehicle observations, action masks, rewards, discounts, and an optional centralized global state.
- Register the Jumanji generator and wrapper in Mava's environment factory.
- Add 2-vehicle/6-customer and 2-vehicle/20-customer configurations.
- Add MultiCVRP to the discrete-environment integration matrix.

## Extra
Validation:
- `uv run pytest test/integration_test.py -q` — 23 passed.
- `uv run pre-commit run --all-files` — all hooks pass except the existing MAGPO mypy mismatch at `mava/systems/gpo/anakin/rec_magpo.py:659` (`HiddenStates` versus `SableHiddenStates`), which this PR does not modify.

Development benchmark:
- Scenario: `multicvrp-2v-20c`
- Budget: 100,000 environment steps, seed 42
- Evaluation: final mean over 8 episodes
- FF-IPPO / FF-MAPPO: -213.29
- REC-IPPO / REC-MAPPO: -214.11
- MAT / FF-Sable / REC-Sable: -212.68
- REC-IQL: -1662.36
- REC-QMIX: -1148.44

All nine supported Anakin discrete systems completed successfully. These are single-seed development results intended to verify end-to-end training compatibility, not tuned multi-seed benchmark claims.

### DIFF
--- a/mava/configs/env/multicvrp.yaml
+++ b/mava/configs/env/multicvrp.yaml
@@ -1,0 +1,19 @@
+# ---Environment Configs---
+defaults:
+  - _self_
+  - scenario: multicvrp-2v-20c # [multicvrp-2v-6c, multicvrp-2v-20c]
+
+env_name: MultiCVRP # Used for logging purposes.
+
+# Defines the metric that will be used to evaluate the performance of the agent.
+# This metric is returned at the end of an experiment and can be used for hyperparameter tuning.
+eval_metric: episode_return
+
+# Whether the environment observations encode implicit agent IDs. If True, the AgentID wrapper is not used.
+# This should not be changed.
+implicit_agent_id: False
+# Whether or not to log the winrate of this environment. This should not be changed as not all
+# environments have a winrate metric.
+log_win_rate: False
+
+kwargs: {}
--- a/mava/configs/env/scenario/multicvrp-2v-20c.yaml
+++ b/mava/configs/env/scenario/multicvrp-2v-20c.yaml
@@ -1,0 +1,8 @@
+name: MultiCVRP-v0
+task_name: multicvrp-2v-20c
+
+task_config:
+  num_customers: 20
+  num_vehicles: 2
+
+env_kwargs: {}
--- a/mava/configs/env/scenario/multicvrp-2v-6c.yaml
+++ b/mava/configs/env/scenario/multicvrp-2v-6c.yaml
@@ -1,0 +1,8 @@
+name: MultiCVRP-v0
+task_name: multicvrp-2v-6c
+
+task_config:
+  num_customers: 6
+  num_vehicles: 2
+
+env_kwargs: {}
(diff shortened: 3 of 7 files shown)


sash-a:
Thanks for this, adding environments is quite tricky. We did try this a while ago and got pretty poor results for MultiCVRP. Can you please benchmark the performance (we'd want to see performance close to the results in the jumanji paper to accept this)

quinnarnold:
@sash-a I ran a fresh, short comparison on MultiCVRP-v0 (2 vehicles, 20 customers), using seed 0, 128 environments, and 3.84 million training steps per run. The matched final evaluations represent 3.456 million policy-training steps, with 1,024 episodes each.

- Mava MAT: **−183.41** final stochastic return.
- Native Jumanji A2C with the paper network/configuration: **−75.20**.

MAT substantially trails the reference in this short run. This is an early-learning diagnostic, not a reproduction of the paper’s full three-seed, 192-million-step result.

I also checked the wrapper directly with identical reset keys and actions across 128 paired rollouts / 2,232 transitions, covering both configured customer counts and global state enabled/disabled. States, rewards, discounts, action masks, termination, and cumulative returns matched exactly, and the raw observation values were preserved. These checks did not identify a wrapper correctness bug.

One substantial difference is the flattened customer information in each Mava vehicle observation versus Jumanji’s separate customer/vehicle attention network. The algorithms and update rules also differ, so the benchmark does not isolate the cause of the gap.

Do you have a working Mava configuration we should compare against, or would you prefer exploring a structured observation/network approach?

The figure below shows the matched stochastic evaluations; its dotted paper endpoint is explicitly at a different, much larger training budget.

comparison

sash-a:
It's almost certainly the network in jumanji and we can't do this generally in Mava without a massive refactor which I don't think is a good idea. So if you'd like to try and see if some better hyperparams could help that would be great. Otherwise I think we should potentially close this and the issue as not completable

quinnarnold:
@sash-a I ran a small learning-rate/value-loss-coefficient grid and repeated the best setting with another seed. Returns stayed around −182 to −185, well below the native Jumanji comparison. Given that and the scope of a network refactor, I agree we should close this PR and #1138 for now.

PR #1218, opened 2026-08-28
fix: align RWARE reward and discount shapes

### LINKED ISSUE #1110: [BUG]Jumanji RWARE reward shape mismatch
We found that in the 4ag RWARE task, env.reward_spec().generate_value() gives something of shape (), but the reward that environment outputs is actually of shape (num_agents).

This is curious because I believe the scenario should output a reward of shape ().

Maybe it's a wrapper thing?

quinnarnold:
## What?

Use singleton reward and discount dimensions in `RwareWrapper` and expose matching `(1,)` specs.

Closes #1110.

## Why?

RWARE repeated scalar rewards and discounts for every agent while retaining scalar underlying specs, so returned timesteps did not conform to the wrapper contract.

## How?

Expand each scalar with one dummy agent dimension and reshape the underlying reward and discount specs while preserving their dtype and bounds.

## Extra

Validation:
- Reset and step spec validation passed.
- Focused RWARE integration test passed.
- Ruff, formatting, mypy, and `git diff --check` passed.

### DIFF
--- a/mava/wrappers/jumanji.py
+++ b/mava/wrappers/jumanji.py
@@ -148,8 +148,8 @@ class RwareWrapper(JumanjiMarlWrapper):
             action_mask=timestep.observation.action_mask,
             step_count=jnp.repeat(timestep.observation.step_count, self.num_agents),
         )
-        reward = jnp.repeat(timestep.reward, self.num_agents)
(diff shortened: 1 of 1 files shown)


sash-a (approved):
⭐

PR #1219, opened 2026-08-30
feat: add evaluation video logging

### LINKED ISSUE #1178: [FEATURE] Add Video Logging during training
It would be really nice to log videos of agents' behavior during and after training to TensorBoard.

#### Proposed Solution:

1. Add logic to the existing Mava/Tensorboard logger class to handle the collection of RGB frames.
2. Call this new video logging function from the logger class during the evaluation phase of the algorithm.

quinnarnold:
## What?

Add opt-in TensorBoard video logging during evaluation. Closes #1178.

## Why?

Evaluation videos make it possible to inspect agent behavior alongside scalar metrics. This follows the Jumanji-style `env.animate(states)` interface requested in the issue discussion.

## How?

When TensorBoard logging and `record_video` are enabled, evaluation retains environment states, selects one completed episode, and passes it to the environment's animation method. The TensorBoard logger saves the animation as a GIF and embeds the GIF in the event file. The logging backend changes to `tensorboardX` so animated image summaries are supported without adding a video encoding dependency.

## Extra

Validated with:

- locked dependency synchronization
- Ruff formatting and lint checks
- mypy checks in the project environment
- enabled and disabled end-to-end video logging smoke checks

### DIFF
--- a/mava/evaluator.py
+++ b/mava/evaluator.py
@@ -15,7 +15,7 @@
 import math
 import time
 import warnings
-from typing import Any, Callable, Dict, Protocol, Tuple, Union
+from typing import Any, Callable, Dict, Protocol, Tuple, Union, cast
 
 import jax
 import jax.numpy as jnp
@@ -121,7 +121,9 @@ def get_eval_fn(
             stacklevel=2,
         )
 
-    def eval_fn(params: FrozenDict, key: PRNGKey, init_act_state: ActorState) -> Metrics:
+    def eval_fn(
+        params: FrozenDict, key: PRNGKey, init_act_state: ActorState
+    ) -> Union[Metrics, Tuple[Metrics, State]]:
         """Evaluates the given params on an environment and returns relevent metrics.
 
         Metrics are collected by the `RecordEpisodeMetrics` wrapper: episode return and length,
@@ -207,7 +209,7 @@ def get_eval_fn(
                 for index in range(episode_length + 1)
             ]
             state_sequence = [state.env_state for state in state_sequence]
-            metrics[_EVAL_VIDEO_KEY] = env.animate(state_sequence)
+            metrics[_EVAL_VIDEO_KEY] = cast(Any, env).animate(state_sequence)
 
         return metrics
 
--- a/mava/evaluator.py
+++ b/mava/evaluator.py
@@ -47,6 +47,8 @@ _EvalEnvStepState: TypeAlias = Tuple[State, TimeStep, PRNGKey, ActorState]
 # The function signature for the mava evaluation function (returned by `get_eval_fn`).
 EvalFn: TypeAlias = Callable[[FrozenDict, PRNGKey, ActorState], Metrics]
 
+_EVAL_VIDEO_KEY = "__eval_video__"
+
 
 class EvalActFn(Protocol):
     """The API for the acting function that is passed to the `EvalFn`.
@@ -103,6 +105,12 @@ def get_eval_fn(
     n_vmapped_envs = get_num_eval_envs(config, absolute_metric)
     n_parallel_envs = n_vmapped_envs * n_devices
     episode_loops = math.ceil(eval_episodes / n_parallel_envs)
+    record_video = (
+        not absolute_metric
+        and config.logger.loggers.tensorboard.enabled
+        and config.logger.loggers.tensorboard.record_video
+        and hasattr(env, "animate")
+    )
 
     # Warnings if num eval episodes is not divisible by num parallel envs.
     if eval_episodes % n_parallel_envs != 0:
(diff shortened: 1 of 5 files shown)


mava/evaluator.py
```diff
@@ -148,25 +163,54 @@ def _episode(key: PRNGKey, _: Any) -> Tuple[PRNGKey, Metrics]:
...
+
+        if record_video:
+            episode_length = int(metrics["episode_length"][0, 0])
+            episode_states = tree.map(lambda x: x[0, 0, : episode_length + 1, 0], states)
+            state_sequence = [
+                tree.map(lambda x, index=index: x[index], episode_states)
+                for index in range(episode_length + 1)
+            ]
+            state_sequence = [state.env_state for state in state_sequence]
+            metrics[_EVAL_VIDEO_KEY] = cast(Any, env).animate(state_sequence)
```
sash-a, line 212: This is the crux of this PR. I am fairly certain that not all envs support animation or have the same API. We'd need to come up with a way to support this generally, fairly sure many envs just render live and don't have an option to save.

SMAX is an example of one that doesn't support the same API

quinnarnold: Thanks, you’re right. Rereading #1178, your suggestion was to unify the rendering APIs around `env.animate(states)`. This PR implements the evaluation-state collection and logging path but assumes that unification already exists. The `hasattr` check doesn’t establish compatible inputs or a saveable result, and the logger specifically expects a Matplotlib animation.

My proposed direction would be for Mava’s environment wrappers to handle the environment-specific rendering and state conversion, exposing the common `animate(states)` interface you suggested. The open questions are what it returns to the logger and how environments without an exportable rendering path report that limitation.

Would you prefer a proposal for that interface first, with Jumanji-style environments and SMAX as concrete cases, before I revise the TensorBoard integration? That would help assess whether the same contract works across different rendering APIs without assuming every environment can save a video. I’ll hold further edits pending your feedback on the interface and initial coverage.

sash-a: Yes would prefer a proposal first, but mainly testing around which environments this would even be possible in. If there are a significant number of envs we can't do this for then we should potentially close the issue

quinnarnold: I tested the rendering paths to see whether a common interface is practical. There seems to be enough support to keep #1178 open, but I’d revise this PR around wrapper-level rendering before adding the evaluation/logging integration.

I successfully exported and inspected short GIFs for Jumanji RobotWarehouse, LevelBasedForaging, Connector and Cleaner, plus MPE, SMAX, Gigastep and MaBrax. VectorConnector shares Connector’s renderer. These were small export tests, so they establish that the rendering paths work, not that the full evaluation integration is correct across every scenario.

Matrax’s renderer raises `NotImplementedError`. RWARE, LBF and SMACLite expose RGB rendering APIs, but missing local dependencies prevented me from validating them. Sebulba would also need a different collection path: those environments are mutable, so we would need to capture frames during evaluation rather than render saved JAX states afterward. I’d defer that work from the first version.

The main interface question is SMAX. Jumanji can animate saved states directly, while SMAX’s existing visualizer needs `(step_key, pre_step_state, actions)` to reconstruct the internal simulation ticks. MPE and SMAX also save animations to a file rather than returning the Matplotlib object the current logger expects.

I propose:

1. **Wrappers own rendering and state conversion.** Each supported wrapper exposes `animate(states, *, actions=None) -> bytes`, returning an encoded GIF with its frame timing. Jumanji wrappers would use native animations; MPE/SMAX would use their visualizers; Gigastep/MaBrax would encode rendered RGB frames. The logger would only write the result, without knowing which renderer produced it.

2. **Record actions alongside states where needed.** For SMAX, retain the actions already computed during evaluation. The original environment-step key can be recovered from each saved `JaxMarlState.key` using the same split as `step`, so recording would not require additional RNG draws. Before integrating this, I’d check that offline replay matches the evaluated trajectory and stops correctly at termination.

3. **Keep recording opt-in and bounded.** Check whether the selected environment supports recording before collecting trajectories. An explicitly requested but unsupported recording should give a clear error. With recording disabled, evaluation should behave exactly as before; with it enabled, retain only the selected episode’s data.

4. **Start with the demonstrated Anakin rendering paths.** Mark Matrax unsupported and defer Gym/Sebulba until their capture/export paths are validated. Each included environment would still need an end-to-end evaluation/export check before claiming support.

I also checked the proposed GIF return type against the PR’s existing TensorBoard image-summary path: the exported SMAX animation survived event-file serialization byte-for-byte, including all 17 frames. That confirms the logger can store it; frontend playback still needs checking.

Would this action-aware wrapper interface be acceptable? If you’d prefer to preserve a strict `animate(states)` contract, I’d leave SMAX out initially rather than make the interface appear compatible with its existing visualizer.

